PR #8, opened 2026-06-01
Add demo-seeding SDK for external content repos

unidoc-ahall:
## Summary

  Adds a thin SDK package, `isms.sh/demo`, for seeding demonstration
  organisations into a running ISMS platform from **external content
  repositories**. The core ships no organisation-specific content — that
  lives in a separate repo (canonically [`unidoc/isms-demo`](https://github.com/unidoc/isms-demo))
  which imports this package and implements the `demo.Content` interface.

  ## Why

  Demo data is per-customer or per-prospect. Bundling it with the platform
  either bloats the OSS repo with content nobody outside the demo cares
  about, or commits us to maintaining shared "example" content that nobody
  actually demos with. Keeping content in a separate repo lets us add
  prospect-specific demos without touching the platform — and because
  content packages import platform types directly, **schema changes in
  the platform break content packages at compile time**.

  ## What's in this PR

  ### `demo/seed.go` — the SDK

  - `Seeder` with `Add*` methods over every entity the platform supports
    (risks, suppliers, assets, systems, incidents, tasks, changes, CAs,
    programmes, objectives, legal, audits, findings, reviews, comments,
    suggestions, references, readings, activity, approval policies,
    asset/supplier reviews, branding).
  - `Content` interface — one method, `Apply(*Seeder) error`.
  - `Run(ctx, db, opts, content)` orchestrator — creates the org and
    hands control to the content package. Refuses to overwrite an
    existing org.
  - `Connect` wrapper around `db.New` so external runners don't import
    `internal/`.
  - Type aliases re-exporting `db.Risk`, `db.Supplier`, ... so content
    packages can build entity values without reaching into `internal/`.

  ### `demo/README.md`

  Short pointer to `unidoc/isms-demo` for a working consumer.

  ### `internal/isms/blob/blob.go`

  Adds `blob.NewFromEnv()` that builds the same `Store` the API server
  does from `ISMS_DATA_DIR` + `ISMS_STORAGE_BACKEND` + `ISMS_S3_*`. Used
  by `Seeder.SetBranding` so demo assets land in the same place real org
  assets do.

  ### `devenv/compose.yml`

  Mounts `../isms-demo` into the dev container so the consumer repo can
  be built and run alongside the platform.

  ## Out of scope

  No CLI command, no embedded content, no demo binary in this repo — all
  of that lives in `unidoc/isms-demo`.

  ## Test plan

  - [x] `go build ./...` clean
  - [x] `unidoc/isms-demo` builds against this branch (via `replace` in its `go.mod`)
  - [x] `isms-demo seed --content acme-logistics` populates an org end-to-end
    (users login, branding visible, risks/suppliers/audits/reviews present)

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -1,0 +1,11 @@
+# `isms.sh/demo`
+
+SDK for seeding demonstration organisations into a running ISMS platform.
+
+This package ships **no organisation-specific content** — only the plumbing
+(`Seeder` + `Content` interface + `Run` orchestrator). Per-customer demo
+content lives in separate repositories that import this package.
+
+The canonical consumer is [`unidoc/isms-demo`](https://github.com/unidoc/isms-demo)
+— see that repo for a complete content package and a working CLI wired around
+`demo.Run`.
--- a/demo/seed.go
+++ b/demo/seed.go
@@ -1,0 +1,334 @@
+// Package demo is the SDK for seeding ISMS demonstration organisations.
+//
+// Run creates the org and applies a Content implementation that adds the
+// entities, documents, reviews, and branding the demo should contain.
+// Per-customer content lives in separate repositories (e.g. unidoc/isms-demo)
+// — this package ships no organisation-specific content.
+//
+// Content packages talk to the platform through the Seeder methods so
+// schema and API changes break them at compile time.
+package demo
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	gogit "github.com/go-git/go-git/v5"
+	"golang.org/x/crypto/bcrypt"
+	"isms.sh/internal/isms/blob"
+	"isms.sh/internal/isms/db"
+	"isms.sh/internal/isms/scaffold"
+	"isms.sh/internal/isms/store"
+)
+
+// Re-exported types from internal packages. External content modules can't
+// import `isms.sh/internal/...`, so the SDK is their single import surface
+// for building entity values.
+type (
+	Risk             = db.Risk
+	Supplier         = db.Supplier
+	Asset            = db.Asset
+	System           = db.System
+	Incident         = db.Incident
+	Task             = db.Task
+	ChangeRequest    = db.ChangeRequest
+	CorrectiveAction = db.CorrectiveAction
+	Program          = db.Program
+	Objective        = db.Objective
+	LegalRequirement = db.LegalRequirement
+	AuditProgramme   = db.AuditProgramme
+	Audit            = db.Audit
+	AuditFinding     = db.AuditFinding
+	ApprovalPolicy   = db.ApprovalPolicy
+	SupplierReview   = db.SupplierReview
+	AssetReview      = db.AssetReview
+	EntityReading    = db.EntityReading
+	Review           = db.Review
+	Comment          = db.Comment
+	Suggestion       = db.Suggestion
+	Epoch            = db.Epoch
+	User             = db.User
+	DB               = db.DB
+)
+
+// DefaultPassword is set on every user created via AddUser. Demo only.
+const DefaultPassword = "demo"
+
+func NewEpoch(t time.Time) Epoch { return db.NewEpoch(t) }
+
+func Connect(ctx context.Context, connStr string) (*DB, error) {
+	return db.New(ctx, connStr)
+}
+
+type Options struct {
+	Slug   string
+	Name   string
+	Domain string
+}
+
+// Content is implemented by per-customer content packages. Apply is called
+// once with a Seeder that has the organisation already created.
+type Content interface {
+	Apply(*Seeder) error
+}
+
+// Run creates the organisation defined by opts and applies the given
+// content. Refuses to overwrite an existing org — callers wanting a fresh
+// demo must delete the org first.
+func Run(ctx context.Context, database *DB, opts Options, content Content) error {
+	if opts.Slug == "" {
+		return fmt.Errorf("demo.Options.Slug is required")
+	}
+	if opts.Name == "" {
+		return fmt.Errorf("demo.Options.Name is required")
+	}
+	if content == nil {
+		return fmt.Errorf("demo.Run: content is nil")
+	}
+	if existing, _ := database.GetOrganizationBySlug(ctx, opts.Slug); existing != nil {
+		return fmt.Errorf("organization %q already exists — delete it first to reseed", opts.Slug)
+	}
+
+	s := &Seeder{db: database, ctx: ctx, users: map[string]*db.User{}}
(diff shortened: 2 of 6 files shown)


unidoc-alip (changes requested):
# Review results and findings

## Findings

### 1 — Silent error discard in `Run` masks DB failures

| | |
|---|---|
| **Severity** | must-fix |
| **File** | `demo/seed.go` |
| **Line** | 91 |

`GetOrganizationBySlug` returns `nil, pgx.ErrNoRows` when not found and `nil, <error>` on any real DB failure. Discarding the error with `_` means a connectivity failure silently passes the guard and proceeds to `CreateOrg`, producing a confusing downstream error instead of a clear failure.

**Fix:**
```go
existing, err := database.GetOrganizationBySlug(ctx, opts.Slug)
if err != nil && !errors.Is(err, pgx.ErrNoRows) {
    return fmt.Errorf("checking for existing org: %w", err)
}
if existing != nil {
    return fmt.Errorf("organization %q already exists — delete it first to reseed", opts.Slug)
}
```

Add to imports: `"errors"` and `"github.com/jackc/pgx/v5"` (already in `go.mod`).

---

### 2 — No cleanup on partial failure blocks retries

| | |
|---|---|
| **Severity** | must-fix |
| **File** | `demo/seed.go` |
| **Line** | 99 |

If `content.Apply(s)` fails mid-way, the org row already exists in the DB. The guard at line 91 then permanently blocks any retry with "organization already exists". The caller must manually soft-delete the org row before trying again.

**Fix — add `DeleteOrganization` to `internal/isms/db/users.go`:**
```go
func (d *DB) DeleteOrganization(ctx context.Context, orgID int) error {
    _, err := d.pool.Exec(ctx, `UPDATE organizations SET deleted_at = now() WHERE id = $1`, orgID)
    return err
}
```

**Fix — clean up on error in `Run`:**
```go
if err := s.CreateOrg(opts.Name, opts.Slug, opts.Domain); err != nil {
    return err
}
if err := content.Apply(s); err != nil {
    _ = database.DeleteOrganization(ctx, s.org.ID) // best-effort: unblock retry
    return fmt.Errorf("seeding content: %w", err)
}
return nil
```

---

### 3 — `firstAdminOrFallback` doesn't filter for admins

| | |
|---|---|
| **Severity** | should-fix |
| **File** | `demo/seed.go` |
| **Line** | 327 |

The method iterates `s.users` with no role check and returns whichever user Go's map iteration yields first (non-deterministic). Any tagged user — contributor, reader — can end up as the git commit author.

**Fix — track admins in a separate slice:**
```go
// Add to Seeder struct
admins []*db.User

// In AddUser, after the tag block:
if role == "admin" {
    s.admins = append(s.admins, usr)
}

// Replace firstAdminOrFallback:
func (s *Seeder) firstAdminOrFallback() identity {
    if len(s.admins) > 0 {
        u := s.admins[0]
        return identity{name: u.Name, email: u.Email}
    }
    for _, u := range s.users {
        if u != nil {
            return identity{name: u.Name, email: u.Email}
        }
    }
    return identity{name: "isms-demo", email: "demo@" + s.org.Slug + ".invalid"}
}
```

---

### 4 — `AddActivity` uses raw SQL while all other methods use typed db helpers

| | |
|---|---|
| **Severity** | should-fix |
| **File** | `demo/seed.go` |
| **Line** | 248 |

`AddActivity` is the only `Add*` method that calls `s.db.Pool().Exec()` directly. The reason is that `db.LogActivity` doesn't accept a `created_at` argument (needed for backdating). The fix is to extend the typed layer rather than bypass it. The raw SQL also silently diverges from `LogActivity` in column order and omits `review_id`.

**Fix — add `LogActivityAt` to `internal/isms/db/postgres.go` after `LogActivity`:**
```go
func (d *DB) LogActivityAt(ctx context.Context, orgID int, a *Activity, at time.Time) error {
    a.OrganizationID = orgID
    _, err := d.pool.Exec(ctx, `
        INSERT INTO activity (organization_id, document_id, review_id, actor, actor_user_id, action, detail, created_at)
        VALUES ($1, $2, $3, $4, (SELECT id FROM users WHERE email = $4), $5, $6, $7)
    `, orgID, a.DocumentID, a.ReviewID, a.Actor, a.Action, a.Detail, at)
    return err
}
```

**Fix — replace `AddActivity` body:**
```go
func (s *Seeder) AddActivity(actor, action, detail, documentID string, at time.Time) error {
    return s.db.LogActivityAt(s.ctx, s.org.ID, &db.Activity{
        Actor:      actor,
        Action:     action,
        Detail:     detail,
        DocumentID: documentID,
    }, at)
}
```

---

### 5 — devenv compose mount breaks for contributors without `isms-demo`

| | |
|---|---|
| **Severity** | should-fix |
| **File** | `devenv/compose.yml` |
| **Line** | 17 |

Docker Compose will error out if `${ISMS_SRC_ROOT}/isms-demo` doesn't exist on the host. Most contributors won't have that repo cloned, breaking `docker compose up` for anyone not working on demo content.

**Fix — use long-form syntax with `required: false` (Compose v2.24+):**
```yaml
- type: bind
  source: ${ISMS_SRC_ROOT}/isms-demo
  target: /home/tower/workspace/isms-demo
  bind:
    create_host_path: false
  required: false
```

---

unidoc-ahall:
Thanks for the thorough review — all five findings applied.

### 1 — Silent error discard in `Run`

Fixed. `Run` now checks the `GetOrganizationBySlug` error explicitly, wraps real DB failures, and only proceeds when the error is `pgx.ErrNoRows`. Added `errors` and `github.com/jackc/pgx/v5` to imports.

### 2 — No cleanup on partial failure

Fixed. Added `DB.DeleteOrganization` (soft delete) in `internal/isms/db/users.go` — the existing `deleted_at IS NULL` filter on the slug unique index means the slug is reusable immediately. `Run` now calls it best-effort when `content.Apply` fails so a reseed isn't blocked by a half-populated row.

### 3 — `firstAdminOrFallback` doesn't filter for admins

Fixed. `Seeder` now tracks admin users in a separate `admins []*db.User` slice, populated in `AddUser` when `role == "admin"`. `firstAdminOrFallback` prefers `admins[0]`, falls back to the non-deterministic users map, then to the synthetic identity.

### 4 — `AddActivity` uses raw SQL

Fixed. Added `DB.LogActivityAt` in `internal/isms/db/postgres.go` (mirrors `LogActivity` with an explicit timestamp argument). `AddActivity` now goes through the typed layer with the full `Activity` struct including `review_id`.

### 5 — devenv compose mount

Fixed. The `isms-demo` mount is now `type: bind` with `required: false` and `create_host_path: false`, so `docker compose up` works for contributors who haven't cloned the demo repo.

Both `isms` and `unidoc/isms-demo` build clean against the changes. Ready for re-review.

unidoc-alip (approved):
Looks good to me

Issue #6, opened 2026-05-26
Review diff: show reviewer's changes per-event, not just full diff from for example document version v0.1


unidoc-ahall:
## Problem

  When a reviewer proposes a revision, the review screen shows the full cumulative diff from the original submitted version. If the reviewer touched several rows of a table  (or did multiple rounds), the author cannot easily tell *what the reviewer actually changed in this revision*.

  ### Example

  On clause 4.2 Interested Parties (`iso27001-4-2`), the reviewer made two small additions inside two table rows:

  - Employees and contractors → prepended `Background check.`
  - Suppliers and service providers → prepended `Review the suppliers' security documentation.`

  The review diff displays `+9 added / -0 removed` and re-renders the entire table, because markdown tables are line-diffed row-by-row. The actual edits are two short phrases
   inside two cells, but that is not visible from the diff alone — the author has to read the decision-record comment to figure out what changed.

  ## Proposal

  UI/UX-only change, no DB migration needed (event store already records each revision as its own event).

  1. **Per-event diff in the review timeline.** Make each timeline entry (`maria proposed a revision`) expandable / clickable so it shows only the diff between the document
  state immediately before and immediately after that event. This is the precise "what did this person change in this action" view.

  2. **Word/intraline diff for changed lines.** Replace the line-level diff with word-level diff (e.g. diff-match-patch) so that inside a changed table row, only the
  added/removed words are highlighted instead of the entire row. Tables especially benefit from this.

  3. **Diff-view toggle.** Default the main review pane to "reviewer's changes only" (diff against the last author-submitted version), with a toggle to switch back to "full
  review diff from v0.1" for the broader context.

  Recommended combination: **(1) + (2)**. Per-event diff answers "what changed in this action"; word-level diff makes table edits actually readable.

  ## Out of scope

  - No schema changes; the revision events already exist.
  - Decision-record / proposed-revision comment storage stays as-is.
  - This is the review-rendering pipeline only.

  ## Acceptance criteria

  - [ ] On a clause document in review, each `proposed_revision` event in the timeline can be expanded to show a diff scoped to that single event.
  - [ ] Table cell edits show word-level highlights (added phrases highlighted, unchanged text not re-rendered as new).
  - [ ] On a review with multiple rounds, the author can step through revisions one at a time and see exactly what each round changed.
  - [ ] Existing "full diff" view is still accessible via toggle or link.

unidoc-ahall:
[...]

Issue #5, opened 2026-05-24
Printing documents only takes the first page


unidoc-ahall:
Need to improve it to take all the pages in the document and outline it nicely.

Issue #4, opened 2026-05-24
Tag members in review comments, rollback plan, and change requests (with notifications)


unidoc-ahall:
Need to @mention/tag members and notify them — not just in review comments, but also on the Rollback Plan and across change requests. Anywhere a member should be
  pulled in, tagging should work and trigger a notification.

  No @mention parsing exists in comments today; the notify package exists but isn't wired for mentions. Needs a tagging primitive + targeted notifications.

  (Broadened from the original review-comments-only scope.)

Issue #3, opened 2026-05-24
Changing document owner is causing ISMS to suggest sending for review although there are no changes except on frontmatter.


unidoc-ahall:
<img width="873" height="493" alt="Image" src="https://github.com/user-attachments/assets/6b95b398-5812-4b06-8549-bb2657f12bcb" />

Only changes are frontmatter changes, this needs to be handled better.

unidoc-ahall:
Also when a document has been approved and there is no owner, it should be automatically set to the first one who works on it, because when showing info it will not show the approved version unless someone is assigned as an owner and author. Best to make the first edit the owner and author and then user can change it.

PR #9, opened 2026-06-03
Self-serve org creation, multi-tenant routing, status-rule parity

### LINKED ISSUE #7: No self-serve org creation for new cloud users
A new visitor clicking "Start in cloud" on isms.sh lands on a login page that asks for an organization slug. 
Since they don't have one yet, they're stuck. There's no way to create a new organization through the public signup flow without manual backend intervention.    

Steps to reproduce                                                                                                                        
  1. Go to https://isms.sh                                                                                                                  
  2. Click "Start in cloud"                                                                                                                 
  3. Enter any email and Login                                                                                                          
  4. Hit the org slug screen with nothing to enter                                                                                        
                                                                                                                                            
  Expected behavior                           
  After email verification, new users should be able to create their own organization (name + slug) and land on a fresh dashboard — no      
  manual setup required.                                                                                                                  
                                                                                                                                            
  Current behavior
  New users get stuck. Onboarding requires either an invite or someone to run backend commands to create the org and add the user.

unidoc-ahall:
Closes #7 — self-serve org creation for new cloud users: signup → create
  org → scaffold template, with org bound on tenant subdomains.

  - Subdomain routing: org bound on tenant subdomains; <slug>.localhost behaves like production domains
  - Status-rule parity: edit-form path now enforces the same rules as the status endpoints (change approval auto-task + email, 409 on closing incidents/CAs with open linked
  work)
  - entity_references canonicalized on per-org identifiers — orphan check, title resolution and CIA advisories matched nothing for UI-created links before
  - Devenv: persistent Go caches, just test-restart, playwright/PATH/CORS fixes, workflow docs

  Test plan: full suite 617 passed (API + Playwright e2e) against devenv test stack.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,115 @@
+# CI test suite — Go unit tests + full API/E2E integration suite.
+#
+# Security model (public repo, fork PRs welcome):
+#   - `pull_request` trigger only — NEVER `pull_request_target`. Fork PRs run
+#     in an isolated ephemeral VM with a read-only GITHUB_TOKEN and no access
+#     to repository secrets.
+#   - This workflow uses no secrets at all; there is nothing to exfiltrate.
+#   - `permissions: contents: read` pins the token to read-only even on push.
+#   - Repo setting "Require approval for all outside collaborators"
+#     (Settings → Actions → General) gates fork-PR runs behind a maintainer
+#     click as defense-in-depth.
+
+name: tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: tests-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  go-unit:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+      - name: Build
+        run: CGO_ENABLED=0 go build ./...
+      - name: Unit tests
+        run: CGO_ENABLED=0 go test ./internal/... -count=1 -timeout 120s
+
+  integration:
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    services:
+      postgres:
+        image: postgres:17
+        env:
+          POSTGRES_USER: isms
+          POSTGRES_PASSWORD: isms
+          POSTGRES_DB: isms_test
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U isms"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
+    env:
+      DATABASE_URL: postgres://isms:isms@localhost:5432/isms_test?sslmode=disable
+      ISMS_SECRET: ci-secret-must-be-at-least-32-characters-long-tests-only
+      ISMS_STORAGE_BACKEND: file
+      ISMS_DATA_DIR: /tmp/isms-ci-data
+      ISMS_TEMPLATE_PATH: ${{ github.workspace }}/isms-templates
+      ISMS_USER_SIGNUP: "1"
+      ISMS_SKIP_EMAIL_VERIFY: "1"
+      ISMS_RATE_LIMIT: "0"
+      ISMS_DOMAIN: localhost
+      ISMS_SUBDOMAIN_ROUTING: "1"
+      ISMS_DEV_MODE: "1"
+    steps:
+      - uses: actions/checkout@v4
+
+      # Templates are a separate repo — the suite scaffolds orgs from them.
+      - name: Checkout templates
+        uses: actions/checkout@v4
+        with:
+          repository: unidoc/isms-templates
+          path: isms-templates
+
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+      - uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+          cache: npm
+          cache-dependency-path: web/package-lock.json
+
+      - name: Build frontend
+        run: |
+          cd web && npm ci && npm run build
+          cd .. && mkdir -p cmd/isms/web/dist && cp -r web/dist/* cmd/isms/web/dist/
+
+      - name: Install test dependencies
+        run: |
+          python3 -m venv .venv
+          .venv/bin/pip install --quiet pytest requests playwright
+          .venv/bin/playwright install --with-deps chromium
+
+      - name: Start server
+        run: |
+          go run ./cmd/isms server migrate
+          nohup go run ./cmd/isms server serve --addr :9090 > server.log 2>&1 &
+          for i in $(seq 1 60); do
+            curl -sf http://localhost:9090/healthz && break
+            sleep 2
+          done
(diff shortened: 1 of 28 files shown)


unidoc-alip (changes requested):
Solid fix for #7 and a good round of parity hardening.

  **What's well done:**
  - The entity-reference identifier canonicalization is a genuine correctness fix. The old `CountOpenCAsByIncident(orgID, numericID)` built `INC-{db_row_id}`, which diverges from the per-org sequence in any shared DB — silently allowing incidents to close with open CAs. Using `existing.Identifier` throughout is the right call.
  - The anchor-interceptor fix (`router.resolve().matched.length > 0` before `preventDefault`) is clean and surgical — only intercepts paths Vue Router actually owns.
  - Subdomain-bound routing guards are well-layered: router guard, `useSession` stale-token redirect, and `getMyOrgs()` suppression all defend the same invariant independently.

  **Two items worth addressing before merge:**
  1. `api_corrective.go` — the `existing.Identifier != ""` guard in the edit-form CA status path silently bypasses the open-task check when the identifier is empty (see inline comment).
  2. `Login.vue` — "Don't have an organization yet? Create one" is misleading copy; `/signup` creates a user account, not an org (see inline comment).

## Findings

| [1](#1--should-fix--internal/isms/api/api_correctivego236) | 🟡 should-fix | `internal/isms/api/api_corrective.go` | 236 |
| [2](#2--should-fix--websrcviewsloginvue34) | 🟡 should-fix | `web/src/views/Login.vue` | 34 |
| [3](#3--nit--internal/isms/api/api_incidentsgo430) | ⚪ nit | `internal/isms/api/api_incidents.go` | 430 |
| [4](#4--nit--internal/isms/api/api_collabgo2560) | ⚪ nit | `internal/isms/api/api_collab.go` | 2560 |
| [5](#5--nit--internal/isms/db/corrective_actionsgo108) | ⚪ nit | `internal/isms/db/corrective_actions.go` | 108 |
| [6](#6--nit--docsmulti-tenant-routingmd93) | ⚪ nit | `docs/multi-tenant-routing.md` | 93 |

---

### 1 · 🟡 should-fix · `internal/isms/api/api_corrective.go:236`

**Silent bypass on empty identifier**

The `existing.Identifier != ""` guard skips the open-task check entirely when the identifier is
empty — but an empty identifier means the `CountOpenTasksByCA` query can't match anything in
`entity_references`, so it returns 0. The CA resolves unchecked, not because there are no open
tasks but because nothing is matchable.

An empty identifier is corrupt data. The right response is an explicit error:

```go
if *req.Status == "resolved" {
    if existing.Identifier == "" {
        return echo.NewHTTPError(http.StatusInternalServerError, "corrective action has no identifier")
    }
    if n, err := s.db.CountOpenTasksByCA(ctx, orgID, existing.Identifier); err == nil && n > 0 {
        return echo.NewHTTPError(http.StatusConflict, fmt.Sprintf("cannot resolve %s: %d open implementation task(s) still linked", existing.Identifier, n))
    }
}
```

Note: `handleUpdateCorrectiveActionStatus` (line 328) uses `existing.Identifier` without a
defensive empty-check, trusting DB constraints. The two paths should be consistent.

<details>
<summary>Why this matters</summary>

The guard is intended as defense-in-depth but has the opposite effect — it trades a loud failure
(corrupt data surfaced immediately) for a silent bypass (enforcement silently disabled). The
dedicated status endpoint doesn't have this guard, so the edit-form path and status-endpoint path
now behave differently on corrupt data.

</details>

---

### 2 · 🟡 should-fix · `web/src/views/Login.vue:34`

**Misleading copy — `/signup` creates an account, not an org**

"Don't have an organization yet? **Create one**" links to `/signup`, but `/signup` only creates a
user account. The org itself is created separately on `/organizations` after email verification.
A new user reading "Create one" will expect to land in a new org — instead they go through signup,
verify email, land on `/organizations`, and discover they still need to create an org.

Suggested alternatives:

- "New to ISMS? **Sign up**"
- "Don't have an account? **Get started**"

This copy is shown only in the org-discovery (no-slug) state, so the audience is exclusively
first-time visitors — accurate framing matters here.

<details>
<summary>Why this matters</summary>

Issue #7's core problem was users not knowing how to proceed. Keeping "Create one" as the CTA
text partially re-creates that confusion at a different step in the flow.

</details>

---

### 3 · ⚪ nit · `internal/isms/api/api_incidents.go:430`

**`statusVerb()` default is unreachable and produces awkward English**

The default case returns `"set " + status + " on"` → `"cannot set investigating on incident"`.
Today only `"closed"` and `"resolved"` are ever passed, so the default is dead code. Options:

- Drop the default and replace with `panic("unreachable")`, or
- Add a doc comment that this function is intentionally scoped to closed/resolved.

Either makes the intent explicit for the next person who adds a caller.

<details>
<summary>Why this matters</summary>

Dead default arms are a future-maintainer trap: a new caller with a different status silently
gets a grammatically wrong error message rather than a compile-time signal.

</details>

---

### 4 · ⚪ nit · `internal/isms/api/api_collab.go:2560`

**`SendTaskAssigned` receives email as display name**

```go
s.mailer.SendTaskAssigned(assignee, assignee, actor, t.Title, t.Priority, baseURL)
//                        ^email    ^name ← same value
```

The second argument is the display name used in the email body — this produces
"Dear admin@acme.com, you've been assigned…" rather than a real name. Pre-existing in the
inlined code, but now that it's extracted into a shared helper called from two paths, it's
worth fixing. The display name is resolvable via `s.db.GetUserByEmail` if needed.

<details>
<summary>Why this matters</summary>

The helper is now called from both `handleUpdateChange` and `handleUpdateChangeStatus`, so one
fix here covers both paths.

</details>

---

### 5 · ⚪ nit · `internal/isms/db/corrective_actions.go:108`

**Two-query pattern in `ByIdentifier` helpers**

`GetCorrectiveActionByIdentifier` (and `GetIncidentByIdentifier` in `db/incidents.go`) issues
two round trips: `SELECT id WHERE identifier = $1`, then `GetByID(id)`. Both can collapse into
a single query:

```sql
SELECT <cols> WHERE identifier = $1 AND organization_id = $2 AND deleted_at IS NULL
```

These are called on every entity sidebar load and risk advisory fetch — worth the one-line fix.

<details>
<summary>Why this matters</summary>

Minor, but the pattern is repeated in two files and these are called from hot paths. No
correctness issue.

</details>

---

### 6 · ⚪ nit · `docs/multi-tenant-routing.md:93`

**Verification section inaccurate for invited users**

> redirects to `/organizations` (picker) so a brand-new user can create their first org

This is only true when `organization_slug` is absent from the JWT response (i.e., the user has
no org). Users joining via an invite already belong to an org — `handleVerifyEmail` returns their
`organization_slug` and `VerifyEmail.vue` redirects them to `/overview`, bypassing the picker.

Suggested: add _"if the user has no org yet"_ to keep the description accurate for both new
users and invited users.

<details>
<summary>Why this matters</summary>

The doc is the authoritative description of the multi-tenant auth flow. An inaccurate statement
about the verification redirect will mislead developers tracing the invite path.

</details>

unidoc-ahall:
All review findings addressed (both should-fix + all four nits). Bonus from CI's first run: brute-force protection was silently disabled for IPv6
  clients (bracketed [::1] failed the inet insert and the error was swallowed) — fixed and verified against a fresh-DB replica. Suite: 617 passed, CI green.

unidoc-alip (approved):
looks good to me

PR #10, opened 2026-06-03
Release toolchain: goreleaser + PR-based release flow

unidoc-ahall:
Two-step release: `just release-pr X.Y.Z` (version-bump PR) →
  merge → `just release X.Y.Z` (signed tag → goreleaser in CI publishes binaries + checksums + changelog).

  - Targets: linux amd64/arm64 (static, alpine-ready), macOS universal, openbsd amd64/arm64
  - Release workflow triggers on v* tags only (tag ruleset restricts those to admins), no secrets beyond GITHUB_TOKEN
  - `just snapshot` builds identical artifacts locally without publishing — verified: static linking, version ldflags, all archives + checksums
  - Tests workflow now also triggers on push to master (was: main)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,50 @@
+# Release on tag push (v*) — builds the frontend, embeds it + migrations,
+# then goreleaser publishes binaries, checksums and changelog to a GitHub
+# Release.
+#
+# Security: triggers only on tags, which only maintainers can push. No
+# secrets beyond the auto-provided GITHUB_TOKEN (scoped to contents:write
+# for creating the release).
+
+name: release
+
+on:
+  push:
+    tags: ["v*"]
+
+permissions:
+  contents: write
+
+jobs:
+  release:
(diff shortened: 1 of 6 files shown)


unidoc-alip (approved):
overall looks good to me

However, I had few for suggestions

- **.github/workflows/release.yml** goreleaser binary version is unpinned in the action — add `version: "v2"` to prevent silent behavior changes from future goreleaser versions.

- SHA-pin the GitHub Actions (checkout, setup-go, setup-node) since the job
  has contents: write
- Add semver format validation in Justfile — just release-pr v0.6.0 would
  silently produce tag vv0.6.0
- Add git fetch --tags before the duplicate-tag guard — currently misses
  remote-only tags
- Add a concurrency block to the release workflow to prevent race conditions
  on re-tags

unidoc-ahall:
Addressed all five suggestions — thanks!

unidoc-alip (approved):
looks good to me

PR #23, opened 2026-06-07
docs: clarify the role model (contributors are propose-only)

unidoc-ahall:
Clarifies the four roles in the README. The key point: a contributor proposes and reports — their input flows through the suggestion/review pipeline, the same way an AI agent's does — and only admin/manager mutate state directly. The per-assignment review grant is preserved. Bringing the code in line (removing contributors from direct create/edit endpoints) is tracked separately for 0.7.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,10 +242,20 @@ Five authentication methods:
 
 ### Roles
 
-- **Admin** — full access, user management, SSO/settings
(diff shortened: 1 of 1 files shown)


unidoc-alip (approved):
looks good to me

PR #29, opened 2026-06-08
docs: release principles + scope clarifications

unidoc-ahall:
Captures decisions we made:

- **Schema changes ship in feature releases**, never patches — patches stay migration-free and trivially safe. The one exception: a serious security issue (or active data loss) whose fix needs a migration may ship in a patch, called out explicitly.
- **Release focus lives in GitHub milestone descriptions**, not a table here — removed the table that would only go stale.
- **Scope clarification**: a standard's requirements are met through generic primitives (e.g. corrective-action effectiveness = an objective trending incidents down, not a 'verified' field). Built-in layers like the incident data-breach fields are deliberate core, not standard-specific leakage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,14 @@ The test for any new capability: *does every deployment need it, generically?* �
 core. Standard-specific → template. External system → integration. Deployment or
 AI wiring → hosting.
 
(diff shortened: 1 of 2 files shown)


unidoc-alip (approved):
looks good to me

PR #36, opened 2026-06-08
docs: add a Why section to the README

unidoc-ahall:
Adds a *Why yet another management system?* section at the top of the README — leading with why the project exists, not what it does.

The gist: management systems are seen as complex and heavy, and most rot the moment the audit is over. We believe one built from the ground up — AI-first, with real discipline in design and use — can stay alive without weighing you down. That's why it's open source: built for companies of all sizes — and finally a real option for the smaller ones who've never had a system that's simple and doesn't rot.

Wording is open to feedback — it's a statement of intent, so it should read true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,15 @@
 # isms.sh — The Intelligent Management System
 
+## Why yet another management system?
+
+Because the ones we have are complex, heavy, and treated as a box-ticking chore — and most of them rot, out of date the moment the audit is over. Meanwhile the stakes keep climbing: in a time when AI and information chaos run wild, getting a real grip on your security matters more than ever.
+
+We don't believe it has to be that way. Security done right starts at the root — understanding and working your risks, not papering over them. Built from the ground up with an AI-first workflow and genuine discipline in how it's designed and used, a management system can stay alive without weighing you down.
+
+That's why it's open source: built for companies of all sizes — and finally a real option for the smaller ones, who've never had a management system that's simple, stays current, and doesn't weigh them down. We believe we can help you and your company — and we hope you'll come to believe it too.
+
+## What it is
+
 An open-source management system platform. Documents live in git, collaboration lives in PostgreSQL, AI operates through suggestions and review — and everything ships as a single binary.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,15 @@
 # isms.sh — The Intelligent Management System
 
+## Why yet another management system?
+
+Because the ones we have are complex, heavy, and treated as a box-ticking chore — and most of them rot, out of date the moment the audit is over. Meanwhile the stakes keep climbing: in a time when AI and information chaos run wild, getting a real grip on your security matters more than ever.
```
unidoc-alip: I have suggestions for this sentences to be like
```
Because the ones we have are complex, opaque, and treated as a compliance exercise — relevant for the audit, irrelevant the day after. 
Meanwhile the stakes keep rising: in a world of accelerating AI adoption and expanding threat surfaces, a management system that doesn't stay alive is worse than useless.
```

README.md
```diff
@@ -1,5 +1,15 @@
...
+
+We don't believe it has to be that way. Security done right starts at the root — understanding and working your risks, not papering over them. Built from the ground up with an AI-first workflow and genuine discipline in how it's designed and used, a management system can stay alive without weighing you down.
```
unidoc-alip: The suggestions for this sentences
```
We believe it doesn't have to work that way. Security done right starts at the root — genuinely understanding and managing your risks, not papering over them. 
Built from the ground up with an AI-first workflow and real discipline in design and use, a management system should be a living part of how you operate, not a burden on top of it.
```

README.md
```diff
@@ -1,5 +1,15 @@
...
+
+That's why it's open source: built for companies of all sizes — and finally a real option for the smaller ones, who've never had a management system that's simple, stays current, and doesn't weigh them down. We believe we can help you and your company — and we hope you'll come to believe it too.
```
unidoc-alip: This one is
```
That's why it's open source: built to work for organisations of all sizes — and finally a credible option for smaller ones, who've never had a management system that's simple, stays current, and doesn't fight them. 
We built this to change that.
```

unidoc-alip (dismissed):
Looks good to me.
Added suggestions as the feedback for wordings

unidoc-ahall:
Thanks @unidoc-alip , I have addressed the comments

unidoc-alip (approved):
Looks good to me

PR #47, opened 2026-06-10
fix: sync stays on main and never silently discards work

### LINKED ISSUE #13: CLI: committed work can be lost — clone and sync leave HEAD detached instead of on main
## Summary

The CLI can leave the local repository in **detached HEAD** state, both right
after `isms clone` and after `isms sync`. Commits made in that state are not on
`main`, so a subsequent sync/push silently leaves them behind — the user's work
appears to vanish.

**Severity: data loss.** This should rank above any cosmetic fix.

## Impact

A user clones an org repo, edits and commits a document, runs `isms sync`, and
the commit never reaches the server. From the user's point of view the work is
gone with no error.

## Steps to reproduce

1. `isms clone <org>`
2. Make a change and commit it
3. `isms sync`
4. Observe: HEAD is detached, the commit is not on `main`, and it is not pushed.

## Root cause (two distinct causes, both confirmed in code)

1. **Clone never pins `main`.** `cmd/isms/clone.go:69` calls `git.PlainClone`
   without a `ReferenceName`, so it relies entirely on the server bare repo's
   `HEAD` symbolic ref pointing at `refs/heads/main`. If it doesn't, the clone
   lands detached.

2. **Fast-forward sync re-detaches HEAD.** `cmd/isms/sync.go:224` fast-forwards
   with `wt.Checkout(&git.CheckoutOptions{Hash: remoteHash})`. Checking out a
   *hash* detaches HEAD; the following line updates the branch ref, but the
   working tree HEAD stays pinned to the hash, not to `main`. There is already
   an `attachToMain` self-heal at the *start* of sync (`sync.go:180-187`), but
   the fast-forward path re-detaches it afterward.

## Proposed fix

1. `clone.go`: pass `ReferenceName: plumbing.NewBranchReferenceName("main")` to
   `CloneOptions` (and verify the server sets bare-repo `HEAD` to
   `refs/heads/main` on org creation).
2. `sync.go:224`: after updating the branch ref, check out the **branch**, not
   the hash — the correct pattern already exists at `sync.go:342`:
   `wt.Checkout(&git.CheckoutOptions{Branch: plumbing.NewBranchReferenceName("main"), Force: true})`.

## Acceptance criteria

- [ ] `isms clone` lands on `main` (not detached) for a non-empty repo.
- [ ] `isms sync` leaves HEAD attached to `main` on every path, including
      fast-forward.
- [ ] A commit made after clone is pushed by the next `isms sync`.
- [ ] Regression test in `cmd/isms` (or `store`) asserting HEAD is on `main`
      after clone and after a fast-forward sync.

unidoc-ahall:
Fixes #13 (clone/sync detached-HEAD data-loss).

  ## Analysis
  - **Clone is already fine:** the git server advertises refs via system `git upload-pack --advertise-refs` (api_git.go), emitting `symref=HEAD:refs/heads/main` (the bare
  repo's HEAD is set to main on bootstrap, store/git.go), so go-git attaches HEAD to main on clone. The original 'clone leaves HEAD detached' symptom no longer reproduces.
  - **The risk was in sync:** the old code silently `attachToMain`'d a detached/other-branch HEAD by force-resetting to the remote tip — discarding any commits made there.
  And the fast-forward path checked out the bare hash, detaching HEAD on every fast-forward.

  ## Behavior
  We don't try to rescue or preserve work made off main — that isn't sync's job, and silent magic is how the data-loss happened.
  - **sync requires main.** If HEAD is detached or on another branch, sync fails with a clear message ('switch to main, or --force'). The user sorts out their own state.
  - **--force** discards local state and resets to main, then syncs.
  - **Fast-forward stays on main** — checks out the branch, not the bare hash, so a normal sync never detaches you.

  Net: after sync you are always on main; HEAD is never silently detached or reset.

  ## Tests
  - `TestRequireMain` — refuses off-main without --force, allows with --force.
  - `TestFastForwardKeepsHeadAttached` — HEAD stays a symbolic ref to main after a fast-forward.

  Closes #13

### DIFF
--- a/cmd/isms/clone.go
+++ b/cmd/isms/clone.go
@@ -7,6 +7,7 @@ import (
 
 	git "github.com/go-git/go-git/v5"
 	gitconfig "github.com/go-git/go-git/v5/config"
+	"github.com/go-git/go-git/v5/plumbing"
 	githttp "github.com/go-git/go-git/v5/plumbing/transport/http"
 	"github.com/spf13/cobra"
 )
@@ -66,13 +67,9 @@ Directory defaults to {org-slug}-isms. All config comes from your env file:
 			remoteURL := baseURL + "/git/" + info.OrganizationUUID
 
 			fmt.Printf("Cloning %s (%s) into %s...\n", info.OrganizationName, info.OrganizationSlug, target)
-			repo, err := git.PlainClone(target, false, &git.CloneOptions{
-				URL: remoteURL,
-				Auth: &githttp.BasicAuth{
-					Username: "x-token-auth",
-					Password: token,
-				},
-				Progress: os.Stdout,
+			repo, err := clonePinnedToMain(target, remoteURL, &githttp.BasicAuth{
+				Username: "x-token-auth",
+				Password: token,
 			})
 			if err != nil {
 				return fmt.Errorf("git clone failed: %w", err)
@@ -109,4 +106,16 @@ Directory defaults to {org-slug}-isms. All config comes from your env file:
 
 	cmd.Flags().StringVar(&dir, "dir", "", "Target directory (default: {slug}-isms)")
 	return cmd
+}
+
+// clonePinnedToMain clones remoteURL into target, pinning the checkout to the
+// main branch so HEAD is never left detached — belt-and-suspenders independent
+// of the server's HEAD advertisement. ISMS has exactly one branch (main).
+func clonePinnedToMain(target, remoteURL string, auth *githttp.BasicAuth) (*git.Repository, error) {
+	return git.PlainClone(target, false, &git.CloneOptions{
+		URL:           remoteURL,
+		ReferenceName: plumbing.NewBranchReferenceName("main"),
+		Auth:          auth,
+		Progress:      os.Stdout,
+	})
 }
--- a/cmd/isms/sync.go
+++ b/cmd/isms/sync.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"io/fs"
 	"os"
@@ -18,6 +19,7 @@ import (
 const gitRemoteName = "origin"
 
 func syncCmd() *cobra.Command {
+	var force bool
 	cmd := &cobra.Command{
 		Use:   "sync",
 		Short: "Sync local ISMS repository with the server",
@@ -29,7 +31,10 @@ All configuration comes from your env file:
   ISMS_ORGANIZATION   — organization UUID
 
 On first run, sync auto-configures the git remote.
-After that, sync does fetch + fast-forward then push.`,
+After that, sync does fetch + fast-forward then push.
+
+Sync must be run on the main branch. Use --force to discard local state
+and reset to main.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			baseURL, token, orgSlug, err := syncEnv()
 			if err != nil {
@@ -55,10 +60,11 @@ After that, sync does fetch + fast-forward then push.`,
 				}
 			}
 
-			return syncRun(repo, root, token)
+			return syncRun(repo, root, token, force)
 		},
 	}
 
+	cmd.Flags().BoolVar(&force, "force", false, "Discard local state and reset to main if HEAD is not on main")
 	return cmd
 }
 
@@ -150,7 +156,7 @@ func setupRemote(repo *git.Repository, baseURL, token, orgSlug string) error {
 }
 
 // syncRun does the actual sync: fetch, fast-forward, push.
-func syncRun(repo *git.Repository, root, token string) error {
+func syncRun(repo *git.Repository, root, token string, force bool) error {
 	auth := &githttp.BasicAuth{
 		Username: "x-token-auth",
 		Password: token,
@@ -167,22 +173,24 @@ func syncRun(repo *git.Repository, root, token string) error {
 		return fmt.Errorf("git fetch failed: %w", err)
 	}
 
-	// Determine current branch. If HEAD is detached, attach to the remote's
-	// default branch — otherwise fast-forward and push both silently no-op on
-	// the "HEAD" pseudo-ref and the user sees no working-tree update.
 	headRef, err := repo.Head()
 	if err != nil {
 		return fmt.Errorf("could not determine current branch — do you have any commits?")
 	}
 
-	// ISMS workflow is always on main. If HEAD is detached, or on any other
-	// branch, snap back to main from the remote tip.
+	// ISMS workflow is always on main. You must be on main to sync — we do NOT
+	// silently rescue a detached HEAD or another branch, since that would risk
+	// discarding local work. Refuse and let the user sort out their own state,
+	// unless --force is given (discard local state and reset to main).
+	if err := requireMain(headRef.Name(), force); err != nil {
+		return err
+	}
 	if headRef.Name() != plumbing.NewBranchReferenceName("main") {
 		attached, attachErr := attachToMain(repo)
 		if attachErr != nil {
-			return fmt.Errorf("could not attach local HEAD to main: %w", attachErr)
+			return fmt.Errorf("could not reset to main: %w", attachErr)
 		}
-		fmt.Println("Attached local HEAD to main.")
+		fmt.Println("--force: discarded local state, reset to main.")
 		headRef = attached
 	}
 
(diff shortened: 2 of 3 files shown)


unidoc-alip (dismissed):
Overall looks good, added comments regarding should we fix these findings?

---
**Verdict: COMMENT** · `should-fix: 2` · `nit: 2`

The sync-side fix is sound: `requireMain` stops silent state-clobbering, `fastForwardToRemote` keeps HEAD symbolic, the tests are targeted and meaningful. Two things need attention before closing issue #13: the clone root cause named in that issue has no defensive fix and no test in this PR, and `fastForwardToRemote`'s `Force: true` extends the "silent magic" the issue set out to eliminate.

---

### Finding 1 — `should-fix` · `cmd/isms/clone.go:69`

**`"Closes #13"` – the clone root cause has no defensive code and no regression test**

Issue #13 names two confirmed root causes. Root cause 2 (fast-forward re-detaches HEAD) is fixed here. Root cause 1 is:

> **"Clone never pins `main`."** `cmd/isms/clone.go:69` calls `git.PlainClone` without a `ReferenceName`, so it relies entirely on the server bare repo's `HEAD` symbolic ref pointing at `refs/heads/main`.

The PR description asserts the symptom no longer reproduces because the server now emits `symref=HEAD:refs/heads/main`. That may be true today, but `clone.go:69` is unchanged:

```go
// current clone.go:69 — no ReferenceName
repo, err := git.PlainClone(target, false, &git.CloneOptions{
    URL:      remoteURL,
    Auth:     auth,
    Progress: os.Stdout,
})
```

The issue's acceptance criteria are:

> - [ ] `isms clone` lands on `main` (not detached) for a non-empty repo.
> - [ ] A commit made after clone is pushed by the next `isms sync`.
> - [ ] Regression test in `cmd/isms` asserting HEAD is on `main` after clone and after a fast-forward sync.

None of those three are met by this PR for the clone path. If a future server regression stops advertising the symref, the clone data-loss re-opens silently.

**Suggested fix** — add `ReferenceName` as a defensive belt-and-suspenders (doesn't require server changes; go-git uses it as a hint):

```go
repo, err := git.PlainClone(target, false, &git.CloneOptions{
    URL:           remoteURL,
    ReferenceName: plumbing.NewBranchReferenceName("main"),
    Auth:          auth,
    Progress:      os.Stdout,
})
```

And add a test (can be lightweight — just verify `repo.Head()` returns `refs/heads/main` after clone, similar to `TestFastForwardKeepsHeadAttached`).

---

### Finding 2 — `should-fix` · `cmd/isms/sync.go` · `fastForwardToRemote`

**`Force: true` re-introduces "silent magic" for uncommitted changes**

Issue #13 proposed using the `sync.go:342` pattern — `Checkout({Branch: main, Force: true})` — as the fix, and this PR follows that proposal exactly. The `Force: true` is intentional and is already the existing pattern in `attachToMain`.

However, the issue scoped its concern to *committed* work being lost via detached HEAD. `Force: true` maps internally to go-git's `HardReset`, which skips the `containsUnstagedChanges()` guard and silently discards uncommitted working-tree files. The issue's own words: *"silent magic is how the data-loss happened."* A user running a plain `isms sync` (no `--force`) with uncommitted edits in their working tree will have those edits wiped without an error or warning.

<details>
<summary>go-git source trace (v5.17.1)</summary>

```
Checkout({Force: true})
  → ro.Mode = HardReset          // Force maps to HardReset
  → w.Reset(ro)
    → ResetSparsely(ro, nil)
      → MergeReset path:  containsUnstagedChanges() → ErrUnstagedChanges   ← skipped for HardReset
      → HardReset path:   resetWorktree() directly                          ← silently wipes files
```

Without `Force`: `MergeReset` runs `containsUnstagedChanges()` and returns `ErrUnstagedChanges` on a dirty tree.

</details>

<details>
<summary>PoC test</summary>

```go
func TestFastForwardDoesNotDiscardUncommittedChanges(t *testing.T) {
    dir := t.TempDir()
    repo, _ := git.PlainInit(dir, false)
    wt, _ := repo.Worktree()
    main := plumbing.NewBranchReferenceName("main")
    _ = repo.Storer.SetReference(plumbing.NewSymbolicReference(plumbing.HEAD, main))

    mkcommit := func(name string) plumbing.Hash {
        os.WriteFile(filepath.Join(dir, name), []byte(name), 0o644)
        wt.Add(name)
        h, _ := wt.Commit("add "+name, &git.CommitOptions{
            Author: &object.Signature{Name: "t", Email: "t@t.test"},
        })
        return h
    }

    c1 := mkcommit("a.txt")
    c2 := mkcommit("b.txt")
    // rewind local to c1
    repo.Storer.SetReference(plumbing.NewHashReference(main, c1))
    wt.Checkout(&git.CheckoutOptions{Branch: main, Force: true})

    // uncommitted edit
    os.WriteFile(filepath.Join(dir, "unsaved.txt"), []byte("important"), 0o644)

    err := fastForwardToRemote(repo, main, c2)

    // with Force: true — err == nil and the file is gone (regression)
    // desired:       — err != nil (ErrUnstagedChanges or wrapped) and file survives
    if err == nil {
        t.Error("expected error for dirty working tree, got nil")
    }
    if _, statErr := os.Stat(filepath.Join(dir, "unsaved.txt")); statErr != nil {
        t.Error("uncommitted file was silently discarded")
    }
}
```

</details>

**Suggested fix** — use `wt.Reset` with `MergeReset` instead. `Reset` moves the branch ref internally via `setHEADCommit` (follows HEAD → symbolic → branch → updates the branch hash), updates the working tree, keeps HEAD symbolic, **and** returns `ErrUnstagedChanges` on a dirty tree. The manual `SetReference` call is no longer needed:

```go
func fastForwardToRemote(repo *git.Repository, _ plumbing.ReferenceName, remoteHash plumbing.Hash) error {
    wt, err := repo.Worktree()
    if err != nil {
        return fmt.Errorf("could not get worktree: %w", err)
    }
    if err := wt.Reset(&git.ResetOptions{Commit: remoteHash, Mode: git.MergeReset}); err != nil {
        if errors.Is(err, git.ErrUnstagedChanges) {
            return fmt.Errorf("uncommitted local changes — commit or stash before syncing")
        }
        return fmt.Errorf("fast-forward failed: %w", err)
    }
    return nil
}
```

---

### Finding 3 — `nit` · `cmd/isms/sync.go` · `fastForwardToRemote`

**`SetReference` before `Checkout` creates a permanent inconsistency on partial failure**

If `SetReference` succeeds but `wt.Checkout` fails (disk full, permissions), the branch ref is at `remoteHash` while the working tree is at the old commit. On the next sync `repo.Head()` resolves HEAD → branch ref → `remoteHash`, so `localHash == remoteHash` — fast-forward is skipped and the stale working tree never self-heals.

Switching to `wt.Reset` (as suggested in Finding 2) eliminates this entirely since the branch ref advance is internal to `Reset` and happens atomically with the working-tree update. If keeping the current structure, a rollback on Checkout failure keeps state consistent:

```go
if err := wt.Checkout(&git.CheckoutOptions{Branch: branchRef, Force: true}); err != nil {
    _ = repo.Storer.SetReference(plumbing.NewHashReference(branchRef, localHash)) // rewind
    return fmt.Errorf("fast-forward failed: %w", err)
}
```

---

### Finding 4 — `nit` · `cmd/isms/sync_test.go` · `TestFastForwardKeepsHeadAttached`

**Test doesn't cover the dirty-worktree contract**

`TestFastForwardKeepsHeadAttached` verifies HEAD attachment after a fast-forward but not whether the function errors (or silently discards) when the working tree is dirty. The PoC in Finding 2 can be added directly as a second test function — it would have caught the `Force: true` issue mechanically and documents the intended contract for future refactors.

unidoc-ahall:

  - **F1 (clone defensive pin):** added `clonePinnedToMain` setting `ReferenceName: refs/heads/main`, so clone lands on main regardless of the server's HEAD
  advertisement. New `TestCloneLandsOnMain` asserts HEAD is a symbolic ref to main after clone.
  - **F2 (Force:true silent wipe):** `fastForwardToRemote` now uses `wt.Reset({Mode: MergeReset})` instead of `Checkout{Force:true}`. A dirty tree returns
  `ErrUnstagedChanges`, surfaced as 'uncommitted local changes — commit or stash before syncing'. No more silent HardReset.
  - **F3 (partial-failure inconsistency):** resolved by the MergeReset switch — the branch-ref advance is atomic with the worktree update, so there's no
  SetReference-then-Checkout window.
  - **F4 (dirty-tree test):** added `TestFastForwardRefusesDirtyWorktree` (your PoC, using an edit to a tracked file) — asserts the FF errors and the local edit survives.

  All green: TestRequireMain, TestFastForwardKeepsHeadAttached, TestFastForwardRefusesDirtyWorktree, TestCloneLandsOnMain.

unidoc-alip (approved):
looks good to me

PR #48, opened 2026-06-11
fix: browser tab title reflects org display name

### LINKED ISSUE #15: Browser tab title is hardcoded "ISMS" — should be the org display name
## Summary

Every tab shows `ISMS` regardless of organization, making multiple open tabs
indistinguishable.

## Cause

`web/index.html:7` hardcodes `<title>ISMS</title>` and nothing updates it at
runtime.

## Proposed fix

Set `document.title` from the current org's display name (falling back to
`ISMS` before an org is resolved), e.g. `"<Org display name> — ISMS"`.

## Acceptance criteria

- [ ] Tab title reflects the current organization's display name.
- [ ] Falls back to `ISMS` on pre-login / org-discovery screens.

unidoc-ahall:
Closes #15.

The browser tab title was hardcoded to 'ISMS' in index.html. Now it reflects the org's display name and updates when branding loads or the user switches orgs.

## Change
- `useSession.js`: a watch on `orgName` sets `document.title` (falls back to 'ISMS' before an org loads). `orgName` already resolves branding_name then organization_name.
- `index.html` keeps `<title>ISMS</title>` as the pre-load static fallback.
- Product identity in the tab comes from the favicon (the shield by default, brandable per org), so the title stays org-only — correct for white-label too.

## Test
- `test_e2e_browser.py::TestTabTitle` — logs in, asserts the tab title becomes the org name ('E2E Org'), not 'ISMS'. Verified against the local test stack (PASSED).

### DIFF
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -126,6 +126,20 @@ class TestSmoke:
         finally:
             ctx.close()
 
+# ── Tab title reflects org (regression for hardcoded "ISMS") ──
+
+class TestTabTitle:
+    def test_title_reflects_org_name(self, pw_browser, tokens):
+        ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1])
+            # The browser tab title must reflect the org, not the hardcoded "ISMS".
+            page.wait_for_function("document.title === 'E2E Org'", timeout=8000)
+            assert page.title() == "E2E Org", f"expected 'E2E Org', got {page.title()!r}"
+        finally:
+            ctx.close()
+
 # ── Documents (shared page, sequential) ──
 
 class TestDocuments:
--- a/web/src/composables/useSession.js
+++ b/web/src/composables/useSession.js
@@ -1,4 +1,4 @@
-import { ref, computed } from 'vue'
+import { ref, computed, watch } from 'vue'
 import { api, getCurrentUser, clearApiToken, setApiToken } from '../api'
 import { orgFromSubdomain, isSubdomainMode, orgEntryURL, setSubdomainRouting, setApexHost, ensureConfigLoaded } from './useCurrentOrg'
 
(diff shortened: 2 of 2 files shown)


unidoc-alip (dismissed):
The fix is correct and closes #15. The `watch` at module scope in `useSession.js` is the right placement — `orgName` is a module-level singleton, so the watch runs once for the page lifetime and picks up every branding/org-switch update. Both acceptance criteria are met: title reflects the org name once branding loads, and falls back to `ISMS` on fresh page loads before an org resolves. Two minor nits; neither is blocking.

---

**Nit 1 — title format** (`web/src/composables/useSession.js` line 12)

The title is set to the bare org name (e.g. `"Acme Corp"`) rather than the `"<Org> — ISMS"` format proposed in #15. The PR description gives a valid white-label justification and the test is consistent with this choice, so it's not a bug. Worth a quick confirmation from the issue reporter that plain org-name is the agreed format before merge.

**Nit 2 — title persists after logout** (`web/src/composables/useSession.js` line 12)

After `logout()` the `orgName` ref is not reset, so the tab title keeps showing the org name on `/login` after a same-session logout. Adding `orgName.value = ''` near the `clearApiToken()` call in `logout()` would make the fallback complete. Arguably out of AC scope ("pre-login" ≠ "post-logout"), so not blocking.

unidoc-ahall:
Thanks @unidoc-alip — both addressed:

**Nit 1 (title format):** confirmed, plain org name is intentional. ISMS is white-label (custom branding name + favicon per org), so prefixing "ISMS"/"— ISMS" would leak the vendor name into a white-labeled tenant. Product identity in the tab comes from the favicon (the shield by default, brandable per org), so the title stays org-only. The `<Org> — ISMS` shape in #15 predates that white-label call.

**Nit 2 (title persists after logout):** good catch — fixed. `logout()` now resets `orgName` after `clearApiToken()`, so the watch sets the tab title back to `ISMS` on `/login`. Pushed.

unidoc-alip (approved):
looks good to me

Issue #50, opened 2026-06-11
Cron manager backfill skips changelog entries


unidoc-ahall:
## Problem

The cron command `isms server manager` backfills missing `next_review` dates on suppliers and systems by calling `UpdateSupplier` / `UpdateSystem` directly (`cmd/isms/manager.go:98-114`). Those db functions only run the SQL `UPDATE`; the entity changelog (audit trail) is written separately at the API-handler layer via `db.LogChange`. The cron skips that, so these automated mutations never appear in the entity changelog — a blind spot in the audit trail.

## Fix direction

The manager already holds the `*db.DB`. After each successful backfill, call `d.LogChange(ctx, orgID, &db.ChangelogEntry{...})` recording the `next_review` change with a system actor (e.g. changed_by `system`/`manager`, reason `automated next_review backfill`). No schema change.

## Acceptance criteria

- Automated `next_review` backfills appear in the entity changelog with a clear system actor.
- A test asserting a backfill produces a changelog entry.

## Notes

Low severity (operator/cron tool, derived date) but audit-trail completeness matters for an ISMS. From the CLI gap analysis.

PR #58, opened 2026-06-11
fix: Ssrf Validation for isms

dewhush:
Hey there! 👋

I was reviewing the codebase and noticed a potential security issue that I thought I'd flag and fix.

## What I found

- **[HIGH] ssrf** in `internal/isms/notify/notify.go`: The application makes HTTP requests to user-configured webhook URLs (SlackWebhook and MatrixServer) without validating o

## What I changed

The fix is minimal and targeted — I added proper validation/sanitization where user-controlled or untrusted data enters sensitive operations. No changes to existing functionality or public APIs.

## Testing

Ran the existing test suite locally, everything passes. The change is backward-compatible.

Happy to discuss if you have questions!

Relates to: https://github.com/unidoc/isms/issues/50

---
💛 If this fix helps, donations are appreciated (ETH/ERC-20): `0x1478f1BDEACc7b434b4405350A15993cDcddc79F` ([Etherscan](https://etherscan.io/address/0x1478f1BDEACc7b434b4405350A15993cDcddc79F))

### DIFF
--- a/internal/isms/notify/notify.go
+++ b/internal/isms/notify/notify.go
@@ -2,8 +2,11 @@ package notify
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


unidoc-ahall:
Thanks for taking the time to look at this and flag it — security reports are always welcome. 🙏

Good news: this specific SSRF vector is already mitigated. Outbound webhook / Matrix requests go through a hardened HTTP client in `internal/isms/notify/notify.go` (`safeHTTPClient`), whose custom `DialContext` resolves the target host and refuses to connect to private, loopback, or link-local addresses — including the cloud-metadata endpoint (`169.254.169.254`) — before dialing. Since redirects reuse the same transport, redirect-based bypasses are covered too.

```go
DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
    host, port, err := net.SplitHostPort(addr)
    if err != nil { return nil, err }
    ips, err := net.LookupIP(host)
    if err != nil { return nil, err }
    for _, ip := range ips {
        if ip.IsPrivate() || ip.IsLoopback() || ip.IsLinkLocalUnicast() || ip.IsLinkLocalMulticast() {
            return nil, errors.New("requests to private/internal IP addresses are not allowed")
        }
    }
    return net.Dial(network, net.JoinHostPort(ips[0].String(), port))
}
```

Since the protection is already in place, I'll close this one out — but thanks again for the careful eye, and please keep flagging anything you spot. 👍

PR #59, opened 2026-06-12
fix: top comment badge counts open comments only

### LINKED ISSUE #30: Comment counter includes resolved comments — should count only open
## Summary

After resolving comments in a document, the top comment badge still counts them
(e.g. "Comments (2)"). The badge should reflect **active/open** comments only.
Resolved comments stay accessible in the panel (toggle), just not in the headline
counter.

## Evidence

- `web/src/views/Documents.vue:432` — the badge uses `comments.length` (total,
  including resolved).
- An `openCommentCount` already exists (`Documents.vue:2430`, from `openComments`
  which filters `status !== 'resolved'`). The fix is to point the badge at it.
- The CommentSidebar already separates open vs resolved (resolvedCount + a "show
  resolved" toggle), so only the top badge is wrong.

## Proposed fix

Badge uses `openCommentCount` instead of `comments.length`. (Check
`commentCountForBlock` too — per-block badges should also count open only.)

## Acceptance criteria

- [ ] Top comment badge counts open comments only; resolving a comment decrements it.
- [ ] Resolved comments remain reachable via the panel toggle.
- [ ] Per-paragraph comment badges count open only.

unidoc-ahall:
Closes #30.

The top comment badge used `comments.length` (total, incl. resolved). It now uses `openCommentCount` (Documents.vue:2430 — filters `status !== 'resolved'` and top-level), so resolving a comment decrements it.

## AC
- Top badge counts open only, decrements on resolve — fixed (badge → `openCommentCount`).
- Resolved comments stay reachable via the panel toggle — unchanged.
- Per-paragraph badges count open only — already correct: `commentCountForBlock` (useDocumentComments.js:63) already filters resolved. No change needed.

## Test
- `test_e2e_browser.py::TestCommentCounter` — seeds 2 comments via API, resolves 1, opens the doc, asserts the badge shows '1' (open) not '2' (total). Old code (`comments.length`) would show '2' → fails; the fix shows '1' → passes. Verified against the local test stack (PASSED).

### DIFF
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -962,3 +962,38 @@ class TestMobile:
             assert "login" not in page.url
         finally:
             ctx.close()
+
+
+# ── Comment counter (regression for #30: top badge counts OPEN only) ──
+
+class TestCommentCounter:
+    DOC = "iso27001-4-1"
+
+    def test_badge_counts_open_only(self, pw_browser, tokens):
+        t = tokens["admin"]
+
+        def _open_comments():
+            r = api("get", f"/documents/{self.DOC}/comments", t, expect_status=200).json()
+            items = r.get("data") if isinstance(r, dict) else r
+            return [c for c in (items or []) if c.get("status") != "resolved" and not c.get("parent_id")]
+
+        # Clean slate: resolve any currently-open top-level comments on the doc.
+        for c in _open_comments():
+            api("post", f"/comments/{c['id']}/resolve", t, expect_status=[200, 204])
+
+        # Two open comments, resolve one → exactly 1 open.
+        c1 = api("post", "/comments", t, json={"document_id": self.DOC, "body": "open one", "paragraph_index": 0}, expect_status=[200, 201]).json()
+        c2 = api("post", "/comments", t, json={"document_id": self.DOC, "body": "to resolve", "paragraph_index": 0}, expect_status=[200, 201]).json()
+        api("post", f"/comments/{c2['id']}/resolve", t, expect_status=[200, 204])
+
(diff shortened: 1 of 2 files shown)


unidoc-alip (dismissed):
Clean, minimal fix — one-line change to `Documents.vue` wires the toolbar badge to the pre-existing `openCommentCount` computed (which already filters top-level, non-resolved comments). Per-paragraph badges were already correct via `commentCountForBlock`. The new e2e test covers the regression path (seed 2, resolve 1, assert badge = 1). 

LGTM.

One pre-existing inconsistency is worth noting (not introduced here): the panel-content guard at `Documents.vue:990` uses `comments.length` (total, includes child replies), so the resolved-comments section renders even when no top-level comments exist. Out of scope for this fix, but worth a follow-up.

---

**Findings**

## 1 nit — <code>web/src/views/Documents.vue:990</code>

This guard (`comments.length > 0`) uses total comment count including child replies, so the panel section renders even when there are no top-level comments — only `parent_id != null` replies exist. 
Not introduced by this PR, but it's the last `comments.length` usage that wasn't touched. Consider `openComments.length > 0 || resolvedComments.length > 0` for consistency (both already exclude replies).

unidoc-ahall:
Addressed the line-990 nit in this PR: the panel-content guard now uses `openComments.length > 0 || resolvedComments.length > 0` (both exclude replies), so the resolved section no longer renders when only child replies exist. That was the last `comments.length` usage in Documents.vue — none remain.

Verified no regression (TestCommentCounter still PASSES against the local stack). The replies-only edge itself isn't separately tested — it's hard to construct cleanly via the API, and the guard is now a boolean of the already-correct `openComments`/`resolvedComments` computeds.

unidoc-alip (approved):
looks good to me

PR #63, opened 2026-06-12
fix: inbox tab deep-links no longer fall back to Comments

### LINKED ISSUE #33: Inbox deep-links to Incidents/Corrective Actions tabs fall back to Comments
## Summary

Notification deep-links to the **Incidents** and **Corrective Actions** inbox
tabs silently land the user on the Comments tab instead — so they never see the
assignment the notification pointed at.

## Root cause

`web/src/views/Inbox.vue:735` — `validTabs` is
`['comments','reviews','tasks','changes','suggestions']`, missing `incidents` and
`corrective_actions`. Those tabs are rendered (lines 369, 394) and listed (830,
832), but `initialTab` (line 736) falls back to `comments` when the route param
isn't in `validTabs`. So `/inbox/incidents` and `/inbox/corrective_actions` both
dump the user on Comments.

## Proposed fix

Add `'incidents'` and `'corrective_actions'` to `validTabs`. One line, no schema.

## Acceptance criteria

- [ ] `/inbox/incidents` and `/inbox/corrective_actions` open the correct tab.
- [ ] Notification deep-links to those tabs land the user on the right assignment.

unidoc-ahall:
Closes #33.

`validTabs` (Inbox.vue) was missing `incidents` and `corrective_actions`, so deep-linking to `/inbox/incidents` or `/inbox/corrective_actions` failed the `validTabs.includes()` check and fell back to the Comments tab. Added both to the list (the tab content templates already existed).

## Test
- `test_e2e_browser.py::TestInboxDeepLinks` — deep-links to each tab and asserts the active tab button is the deep-linked one (not Comments). Verified against the local stack (2 passed). Old code → active tab 'Comments' → fails; fix → 'Incidents'/'CAs' → passes.

### DIFF
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -997,3 +997,25 @@ class TestCommentCounter:
             ctx.close()
             # teardown: resolve the remaining open comment
             api("post", f"/comments/{c1['id']}/resolve", t, expect_status=[200, 204])
+
+
+# ── Inbox deep-links (regression for #33: tab deep-links fell back to Comments) ──
+
+class TestInboxDeepLinks:
+    @pytest.mark.parametrize("tab,label", [
+        ("incidents", "Incidents"),
+        ("corrective_actions", "CAs"),
+    ])
+    def test_deeplink_activates_tab(self, pw_browser, tokens, tab, label):
+        # Deep-linking to /inbox/<tab> must activate that tab, not fall back to
+        # Comments (validTabs was missing incidents/corrective_actions).
+        ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1])
+            page.goto(f"{BASE}/{ORG}/inbox/{tab}")
+            # The active tab button carries 'text-white'; assert it's the deep-linked one.
+            active = page.locator('div.flex.gap-1.bg-slate-900 button.text-white').first
+            expect(active).to_contain_text(label, timeout=8000)
+        finally:
+            ctx.close()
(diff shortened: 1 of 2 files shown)


unidoc-alip (dismissed):
Clean, well-scoped fix. The one-liner in `validTabs` is the right change — all seven tab content templates already existed, so this purely unlocks the routes that were silently falling back to Comments. The regression test is clear and well-structured. One nit on the test selector robustness, but nothing blocking.

---

## Findings

| 1 | nit | `tests/test_e2e_browser.py` | 1019 |

### 1 Nit — test selector tied to Tailwind utility class (`tests/test_e2e_browser.py:1019`)

```python
active = page.locator('div.flex.gap-1.bg-slate-900 button.text-white').first
```

The `button.text-white` selector ties the assertion to a Tailwind utility class. If the active-tab highlight style ever changes (e.g. the active class becomes `bg-white` instead of `text-white`), this test will silently pass while pointing at the wrong button. Consider adding a `data-testid="active-tab"` attribute in `Inbox.vue` and selecting on that instead, or use `aria-selected="true"` if the buttons carry ARIA roles — both are stable across styling changes.

Not blocking — the test correctly catches the regression as written today.

unidoc-ahall:
Thanks @unidoc-alip — addressed the selector nit. Added `data-testid="active-tab"` to the active tab button in Inbox.vue and switched the test to `[data-testid="active-tab"]`, so it's decoupled from the Tailwind highlight class (won't silently pass if the active style changes). Went with data-testid over aria-selected since the tab buttons don't carry ARIA roles yet. Re-verified on the local stack (2 passed).

unidoc-alip (approved):
looks good to me

PR #64, opened 2026-06-12
fix: guard against silent data loss editing docs with embedded HTML/SVG

### LINKED ISSUE #61: Web editor silently drops embedded HTML/SVG on save (data loss)
## Problem

A document containing embedded raw HTML/SVG loses it on a single web-editor save — silent data loss. Confirmed pipeline:

- **Load:** `markdownToHtml` (marked passes raw HTML through) → TipTap `setContent` (`DocumentEditor.vue:378`). ProseMirror's schema has no node for `<svg>` / arbitrary HTML, so it's dropped on parse. (Tables survive because there's a Table extension + a turndown `alwaysHtmlTable` rule; svg/unknown HTML has neither.)
- **Save:** `htmlToMarkdown(editor.getHTML())` (`DocumentEditor.vue:413-414`) serializes the editor's HTML — which no longer contains the svg — so the doc is written back without it.
- There is **no raw/source markdown edit mode** to bypass the WYSIWYG editor.

Result: one edit + save via the web editor silently deletes embedded HTML/SVG. Surfaced by a hand-pasted status SVG in a CLI-authored document.

## Scope (this issue / 0.6.1) — data-loss guard

Prevent the **silent** loss. When a document containing embedded raw HTML/SVG (non-table block HTML the editor can't represent) is opened for editing, warn and/or guard against overwrite — e.g. a banner ("this document has embedded HTML the editor can't preserve; editing here will drop it — edit the source via CLI") and/or require explicit confirmation before save, or open read-only. The editor must never silently drop content.

## Out of scope (future)

- Full preservation of arbitrary HTML in the editor (a custom TipTap node to hold raw HTML/SVG on load + a turndown `keep` rule on save).
- A raw/source markdown edit mode.
- The `status-overview` directive (separate issue) sidesteps this for the status-visual use case — a fenced code block survives the round-trip, unlike embedded SVG.

## Acceptance criteria

- Opening/saving a document that contains embedded HTML/SVG never silently drops it — the user is warned and/or the save is guarded.
- Normal documents (markdown + tables) round-trip unchanged.
- Test: a document with an embedded `<svg>` is not silently stripped by an edit + save through the web path.

unidoc-ahall:
Closes #61.

  The WYSIWYG editor (TipTap/ProseMirror) silently drops embedded raw HTML/SVG on save. This adds a guard: editing a document that
  contains embedded HTML/SVG now asks for explicit confirmation first (`startEditGuarded` in Documents.vue) — never a silent drop.
  Tables are excluded (they round-trip). Full editability of embedded HTML is out of scope (future).

  ## Test
  - `test_e2e_browser.py::TestEditorHtmlGuard` — creates a doc with embedded <svg>, clicks Edit, asserts the warning appears.
  Verified on the local stack (PASSED).

### DIFF
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -997,3 +997,33 @@ class TestCommentCounter:
             ctx.close()
             # teardown: resolve the remaining open comment
             api("post", f"/comments/{c1['id']}/resolve", t, expect_status=[200, 204])
+
+
+# ── Editor data-loss guard (regression for #61: editing drops embedded HTML/SVG) ──
+
+class TestEditorHtmlGuard:
+    DOC = "e2e-svg-doc"
+
+    def test_edit_guard_warns_on_embedded_svg(self, pw_browser, tokens):
+        t = tokens["admin"]
+        # A document with embedded raw SVG (the WYSIWYG editor can't preserve it).
+        api("post", "/documents", t, json={
+            "folder": "iso27001",
+            "filename": "e2e-svg-doc.md",
+            "document_id": self.DOC,
+            "title": "E2E SVG Doc",
+            "content": "# Heading\n\nBody text here.\n\n<svg width='12' height='12'><rect width='12' height='12'/></svg>\n",
+        }, expect_status=[200, 201, 409])
+
+        ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1])
+            page.goto(f"{BASE}/{ORG}/documents/{self.DOC}")
+            # Wait for the body to render (rawContent loaded) before triggering the guard.
+            expect(page.locator("text=Body text here.")).to_be_visible(timeout=10000)
+            page.get_by_role("button", name="Edit", exact=True).first.click()
+            # Guard: editing a doc with embedded SVG must warn first, not silently enter edit.
+            expect(page.locator("text=editing here will drop it")).to_be_visible(timeout=8000)
+        finally:
+            ctx.close()
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -1003,27 +1003,71 @@ class TestCommentCounter:
 
 class TestEditorHtmlGuard:
     DOC = "e2e-svg-doc"
+    SVG_CONTENT = "# Heading\n\nBody text here.\n\n<svg width='12' height='12'><rect width='12' height='12'/></svg>\n"
+
+    def setup_svg_doc(self, t):
+        """Ensure the SVG doc exists WITH its embedded SVG, regardless of prior runs."""
+        api("post", "/documents", t, json={
+            "folder": "iso27001", "filename": "e2e-svg-doc.md",
+            "document_id": self.DOC, "title": "E2E SVG Doc",
+            "content": self.SVG_CONTENT,
+        }, expect_status=[200, 201, 409])
+        # Unconditionally overwrite — a 409 above could leave a stale, SVG-free doc
+        # (e.g. a prior run confirmed the warning and saved via the web editor).
+        api("put", f"/documents/{self.DOC}/content", t, json={"content": self.SVG_CONTENT}, expect_status=200)
 
     def test_edit_guard_warns_on_embedded_svg(self, pw_browser, tokens):
         t = tokens["admin"]
-        # A document with embedded raw SVG (the WYSIWYG editor can't preserve it).
-        api("post", "/documents", t, json={
-            "folder": "iso27001",
-            "filename": "e2e-svg-doc.md",
-            "document_id": self.DOC,
-            "title": "E2E SVG Doc",
-            "content": "# Heading\n\nBody text here.\n\n<svg width='12' height='12'><rect width='12' height='12'/></svg>\n",
-        }, expect_status=[200, 201, 409])
-
+        self.setup_svg_doc(t)
         ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
         page = ctx.new_page()
         try:
             do_login(page, ADMIN[0], ADMIN[1])
             page.goto(f"{BASE}/{ORG}/documents/{self.DOC}")
-            # Wait for the body to render (rawContent loaded) before triggering the guard.
             expect(page.locator("text=Body text here.")).to_be_visible(timeout=10000)
             page.get_by_role("button", name="Edit", exact=True).first.click()
             # Guard: editing a doc with embedded SVG must warn first, not silently enter edit.
             expect(page.locator("text=editing here will drop it")).to_be_visible(timeout=8000)
+        finally:
+            ctx.close()
+
+    def test_edit_guard_cancel_stays_readonly(self, pw_browser, tokens):
+        """Cancelling the warning must NOT enter edit mode."""
+        t = tokens["admin"]
+        self.setup_svg_doc(t)
+        ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1])
+            page.goto(f"{BASE}/{ORG}/documents/{self.DOC}")
+            expect(page.locator("text=Body text here.")).to_be_visible(timeout=10000)
+            page.get_by_role("button", name="Edit", exact=True).first.click()
+            expect(page.locator("text=editing here will drop it")).to_be_visible(timeout=8000)
+            page.get_by_role("button", name="Cancel").click()
+            # Editor must NOT open — no Save button in read-only mode.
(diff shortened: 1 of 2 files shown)


unidoc-alip (changes requested):
## Summary

The guard is the right primitive — intercepting `startEdit` via `startEditGuarded` and checking `rawContent` before the WYSIWYG editor parses it correctly prevents silent loss for the most dangerous HTML types (SVG, iframes, canvas, etc.). Two test gaps worth fixing: the e2e only verifies the warning appears, not that cancel keeps the doc read-only or that clean docs still edit freely without a dialog; and the `409`-swallow in test setup can silently leave a stale doc (without SVG) that defeats the test on the second run. One nit about undocumented design intent in the regex.

---

## Finding 1 — should-fix · `tests/test_e2e_browser.py:1025`

The test only checks that the warning dialog appears when Edit is clicked on an SVG doc. Two missing contract paths:

1. **Cancel path** — clicking Cancel in the dialog must keep edit mode off. Without a test, a broken guard that shows the dialog but calls `startEdit()` regardless would still pass.
2. **Non-regression path** — a document with plain markdown must open the editor directly, without any confirmation dialog. This is the "normal documents round-trip unchanged" acceptance criterion from #61 and is currently unverified.

**Proof of concept**

Introduce a regression: remove the `if (!ok) return` guard in `startEditGuarded`:

```js
async function startEditGuarded() {
  if (hasEditorUnsafeHtml(rawContent.value)) {
    const { ask } = useConfirm()
    await ask("...", "Embedded HTML will be lost")
    // BUG: forgot to check ok — always enters edit mode
  }
  startEdit()
}
```

The existing test still passes (warning appears), but the guard is broken — edit mode opens even when the user cancels. The missing cancel-path test would catch this.

**Suggested fix**

Add two methods to `TestEditorHtmlGuard`:

```python
def test_edit_guard_cancel_stays_readonly(self, pw_browser, tokens):
    """Cancelling the warning must NOT enter edit mode."""
    t = tokens["admin"]
    api("post", "/documents", t, json={
        "folder": "iso27001",
        "filename": "e2e-svg-doc.md",
        "document_id": self.DOC,
        "title": "E2E SVG Doc",
        "content": "# Heading\n\nBody text here.\n\n<svg width='12' height='12'><rect width='12' height='12'/></svg>\n",
    }, expect_status=[200, 201, 409])

    ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
    page = ctx.new_page()
    try:
        do_login(page, ADMIN[0], ADMIN[1])
        page.goto(f"{BASE}/{ORG}/documents/{self.DOC}")
        expect(page.locator("text=Body text here.")).to_be_visible(timeout=10000)
        page.get_by_role("button", name="Edit", exact=True).first.click()
        expect(page.locator("text=editing here will drop it")).to_be_visible(timeout=8000)
        # Cancel — editor must NOT open
        page.get_by_role("button", name="Cancel").click()
        expect(page.get_by_role("button", name="Save")).not_to_be_visible(timeout=3000)
    finally:
        ctx.close()

def test_edit_guard_clean_doc_no_warning(self, pw_browser, tokens):
    """A document with no embedded HTML must open the editor without a warning."""
    t = tokens["admin"]
    clean_doc = self.DOC + "-clean"
    api("post", "/documents", t, json={
        "folder": "iso27001",
        "filename": clean_doc + ".md",
        "document_id": clean_doc,
        "title": "E2E Clean Doc",
        "content": "# Heading\n\nJust plain markdown. No HTML here.\n",
    }, expect_status=[200, 201, 409])

    ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
    page = ctx.new_page()
    try:
        do_login(page, ADMIN[0], ADMIN[1])
        page.goto(f"{BASE}/{ORG}/documents/{clean_doc}")
        expect(page.locator("text=Just plain markdown.")).to_be_visible(timeout=10000)
        page.get_by_role("button", name="Edit", exact=True).first.click()
        # Guard must NOT appear — editor opens directly
        expect(page.locator("text=editing here will drop it")).not_to_be_visible(timeout=2000)
        expect(page.get_by_role("button", name="Save")).to_be_visible(timeout=5000)
    finally:
        ctx.close()
```

---

## Finding 2 — should-fix · `tests/test_e2e_browser.py:1017`

```python
api("post", "/documents", t, json={
    ...,
    "content": "...<svg .../>...",
}, expect_status=[200, 201, 409])   # ← 409 silently accepted
```

`expect_status=[200, 201, 409]` accepts 409 (document already exists) without touching the document content. If a prior run left the document with its SVG stripped (for example: the test passed, then someone manually confirmed the warning and saved via the web UI, dropping the SVG), the next run will hit 409, skip creation, navigate to the now-SVG-free document, and the guard will not fire — causing a timeout on the `expect(…"editing here will drop it"…)` assertion with no helpful error message.

**Proof of concept**

Reproduce in three steps:
1. Run the test suite → creates the doc with SVG, `test_edit_guard_warns_on_embedded_svg` passes.
2. Manually open the doc in the web UI, click Edit, confirm the warning, and save → SVG is silently dropped from git.
3. Run the test suite again → `POST` returns 409 (accepted), doc still exists but has no SVG, guard does not fire, test times out after 8 s with `TimeoutError: Locator.expect` on the warning text.

**Suggested fix**

Always guarantee the document has the expected content by using `PUT` after `POST`. The `DELETE`-then-`POST` pattern is also valid but two round-trips. `PUT` is idempotent and sufficient:

```python
SVG_CONTENT = "# Heading\n\nBody text here.\n\n<svg width='12' height='12'><rect width='12' height='12'/></svg>\n"

def setup_svg_doc(self, t):
    """Ensure the SVG test doc exists with the expected content regardless of prior runs."""
    api("post", "/documents", t, json={
        "folder": "iso27001",
        "filename": "e2e-svg-doc.md",
        "document_id": self.DOC,
        "title": "E2E SVG Doc",
        "content": self.SVG_CONTENT,
    }, expect_status=[200, 201, 409])
    # Unconditionally overwrite content — handles the case where a prior run
    # stripped the SVG via the web editor (409 path above would leave it empty).
    api("put", f"/documents/{self.DOC}/content", t, json={"content": self.SVG_CONTENT}, expect_status=200)
```

Then replace the inline setup in each test method with `self.setup_svg_doc(t)`.

---

## Finding 3 — nit · `web/src/views/Documents.vue:1731`

```js
// Block-level raw HTML / SVG the WYSIWYG editor can't represent — it would be
// silently dropped on save. Tables are excluded (they round-trip as HTML).
function hasEditorUnsafeHtml(md) {
  if (!md) return false
  return /<\s*(svg|iframe|object|embed|canvas|video|audio|form|style|script|figure|details|section|article|aside|nav)\b/i.test(md)
}
```

`<div>` is absent. A future contributor reading the regex cannot tell whether this was intentional or an oversight. In fact it is intentional: TipTap/ProseMirror extracts text content from unknown block containers, so `<div>text</div>` round-trips as `text` — the wrapper is lost but the text survives. This is structural loss, not complete data loss, which is why `<div>` correctly does not belong in the guard.

**Proof of concept**

Round-trip a `<div>` through the editor pipeline:

```
Input markdown:   <div class="note">Important note.</div>
markdownToHtml → <div class="note">Important note.</div>
TipTap parse   → ParagraphNode("Important note.")   // div wrapper dropped, text kept
editor.getHTML()→ <p>Important note.</p>
htmlToMarkdown → Important note.
```

Text survives; only the `class="note"` and `<div>` wrapper are lost. Compare with SVG:

```
Input markdown:   <svg width='12'><rect width='12'/></svg>
markdownToHtml → <svg width='12'><rect width='12'/></svg>
TipTap parse   → (dropped entirely — no node type for svg)
editor.getHTML()→ (nothing)
htmlToMarkdown → (nothing)
```

Complete data loss — correctly caught by the guard.

**Suggested fix**

Add a one-line comment so the intent is explicit:

```js
function hasEditorUnsafeHtml(md) {
  if (!md) return false
  // <div> is intentionally excluded: ProseMirror extracts its text content,
  // so only the wrapper is lost — not a complete data loss like svg/canvas/iframe.
  return /<\s*(svg|iframe|object|embed|canvas|video|audio|form|style|script|figure|details|section|article|aside|nav)\b/i.test(md)
}
```

unidoc-ahall:

1. **Test contract gaps** (should-fix): added two methods — `test_edit_guard_cancel_stays_readonly` (Cancel must NOT enter edit mode → asserts no Save button) and `test_edit_guard_clean_doc_no_warning` (plain-markdown doc opens the editor directly, no dialog). Confirmed your PoC: dropping `if (!ok) return` now fails the cancel test.
2. **409-swallow leaving a stale doc** (should-fix): added `setup_svg_doc` that POSTs then unconditionally PUTs the content, so the doc always has its SVG regardless of prior runs. Used in both SVG tests.
3. **`<div>` regex intent** (nit): added a comment explaining `<div>` is intentionally excluded (ProseMirror keeps its text — structural loss, not the complete loss of svg/canvas/iframe).

All three tests pass on the local stack (3 passed).

unidoc-alip (dismissed):
looks good to me

unidoc-alip (approved):
LGTM

Issue #56, opened 2026-06-11
Markdown rendering: syntax-highlight code blocks + clean styling (shared renderer)


unidoc-ahall:
## Problem

Code blocks (and markdown generally) render plainly — no syntax highlighting and no code-block styling — so change notes and other markdown surfaces look poor (cf. Nuclino). This matters across the app: change notes, register descriptions/notes, comments, etc.

## Findings

- The canonical renderer is `web/src/composables/useRenderMd.js`: `DOMPurify.sanitize(marked.parse(text, { breaks: true }))` (marked v17 + DOMPurify).
- No syntax highlighter is wired in (no highlight.js / prism / shiki) and there's no code-block CSS → fenced blocks are unstyled `<pre><code>`.
- Register views and change notes already use this shared renderer, so fixing it there fixes them immediately.

## Scope (this issue)

- Add syntax highlighting to the shared renderer: `marked-highlight` + **highlight.js** (synchronous, fits the current `v-html` path; Shiki would require going async).
- Configure DOMPurify to keep the highlighter's `class` attributes on `<pre>/<code>/<span>`.
- Add a clean code-block CSS theme (+ inline code, tables, blockquotes) for the "render nicely" look. This also covers #14 (tables with inline code render distorted) — same surface.

## Follow-up (separate)

- Consolidate the ~6 components that call `marked` directly (TrackChanges, DocumentDiff, DocumentViewer, SideBySideReview, Documents, Reviews) onto the shared renderer so rendering is consistent everywhere. Some (diff / track-changes) intentionally differ — case-by-case.

## Acceptance criteria

- Code blocks render with syntax highlighting and clean styling in change notes and register views.
- Tables (incl. inline code) render correctly.

## Notes

~half a day for this scope (one composable + CSS); the follow-up consolidation is ~another day. Relates to #14.

PR #65, opened 2026-06-12
fix: markdown table inline code renders without breaking mid-token

### LINKED ISSUE #14: Document view: markdown tables with inline code render distorted
## Summary

Markdown tables that contain inline `code` spans render with uneven rows in the
document view — code chips wrap inside cells and stack, so columns no longer
line up cleanly.

## Evidence

Reproduced with the contributor-identity table (see screenshot in the original
report): the EXAMPLE column wraps `sts-john` / email onto separate lines and the
row heights go uneven.

## Likely cause

No table-specific layout rules in the prose CSS — inline `code` in `<td>` is
free to wrap mid-cell with no `white-space`/`table-layout` control.

## Proposed fix

In the document prose styles: give table cells predictable layout, e.g.
`table-layout: fixed` with sensible column behavior, and stop inline `code`
chips from wrapping awkwardly (`td code { white-space: nowrap }` or allow
horizontal scroll on overflow).

## Acceptance criteria

- [ ] A table with inline code in multiple columns renders with aligned rows.
- [ ] Long code spans degrade gracefully (wrap or scroll) without breaking the
      grid.

unidoc-ahall:
Closes #14.

Markdown tables with inline code rendered distorted in the document view: code chips (e.g. `sts-USERNAME@commandvector.com`) wrapped mid-token and stacked, so rows went uneven and columns didn't line up.

Fix: keep inline code chips whole in table cells (`white-space: nowrap`), so cells wrap only between separate code spans. Both rendering paths:
- Document-view grid cells (`.tbl-cell`/`.tbl-hdr-cell code`, Documents.vue) — the doc view renders tables as a grid, not <table>.
- Real-table renderers (`.doc-prose td code/th code`, style.css) — DocumentViewer / TrackChanges.

Scoped; broader table/markdown polish (syntax highlighting) is #56.

## Test
- `test_e2e_browser.py::TestTableInlineCodeCss` — renders a doc with a table of inline code, asserts cell code computes `white-space: nowrap`. Verified on the local stack (PASSED).

### DIFF
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -1003,24 +1003,27 @@ class TestCommentCounter:
 
 class TestTableInlineCodeCss:
     DOC = "e2e-table-code"
+    # A genuinely long token (55+ chars) so the overflow case is exercised.
+    LONG = "sts-USERNAME@a-very-long-contractor-domain-name.example.com"
+    CONTENT = (
+        "# Table\n\n"
+        "| Type | Example |\n"
+        "|------|---------|\n"
+        "| Employee | `sts-john` / `john@example.com` |\n"
+        f"| Long | `{LONG}` |\n"
+    )
+
+    def setup_doc(self, t):
+        api("post", "/documents", t, json={
+            "folder": "iso27001", "filename": "e2e-table-code.md",
+            "document_id": self.DOC, "title": "E2E Table Code", "content": self.CONTENT,
+        }, expect_status=[200, 201, 409])
+        # Unconditionally overwrite so the table (incl. the long token) is always present.
+        api("put", f"/documents/{self.DOC}/content", t, json={"content": self.CONTENT}, expect_status=200)
 
     def test_table_cell_code_is_nowrap(self, pw_browser, tokens):
         t = tokens["admin"]
-        content = (
-            "# Table\n\n"
-            "| Type | Example |\n"
-            "|------|---------|\n"
-            "| Employee | `sts-john` / `john@example.com` |\n"
-            "| Contractor | `sts-USERNAME@contractor.com` |\n"
-        )
-        api("post", "/documents", t, json={
-            "folder": "iso27001",
-            "filename": "e2e-table-code.md",
-            "document_id": self.DOC,
-            "title": "E2E Table Code",
-            "content": content,
-        }, expect_status=[200, 201, 409])
-
+        self.setup_doc(t)
         ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
         page = ctx.new_page()
         try:
@@ -1030,5 +1033,23 @@ class TestTableInlineCodeCss:
             code.wait_for(state="visible", timeout=10000)
             ws = code.evaluate("el => getComputedStyle(el).whiteSpace")
             assert ws == "nowrap", f"expected white-space:nowrap on table-cell code, got {ws!r}"
+        finally:
+            ctx.close()
+
+    def test_long_code_does_not_break_grid(self, pw_browser, tokens):
+        """A long code token at a narrow viewport must clip inside its cell, not
+        overflow the grid into the adjacent column (#14 AC: degrade gracefully)."""
+        t = tokens["admin"]
+        self.setup_doc(t)
+        ctx = pw_browser.new_context(viewport={"width": 760, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1])
+            page.goto(f"{BASE}/{ORG}/documents/{self.DOC}")
+            # The grid row holding the long token must not overflow horizontally.
+            row = page.locator(f'.doc-prose .tbl-grid:has(code:has-text("{self.LONG}"))').first
+            row.wait_for(state="visible", timeout=10000)
+            overflow = row.evaluate("el => el.scrollWidth - el.clientWidth")
+            assert overflow <= 1, f"grid row overflows by {overflow}px — long code broke the grid"
         finally:
             ctx.close()
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -997,3 +997,38 @@ class TestCommentCounter:
             ctx.close()
             # teardown: resolve the remaining open comment
             api("post", f"/comments/{c1['id']}/resolve", t, expect_status=[200, 204])
+
+
+# ── Table inline-code CSS (regression for #14: code chips wrapped mid-token) ──
+
+class TestTableInlineCodeCss:
+    DOC = "e2e-table-code"
+
+    def test_table_cell_code_is_nowrap(self, pw_browser, tokens):
+        t = tokens["admin"]
+        content = (
+            "# Table\n\n"
+            "| Type | Example |\n"
+            "|------|---------|\n"
+            "| Employee | `sts-john` / `john@example.com` |\n"
+            "| Contractor | `sts-USERNAME@contractor.com` |\n"
+        )
+        api("post", "/documents", t, json={
+            "folder": "iso27001",
+            "filename": "e2e-table-code.md",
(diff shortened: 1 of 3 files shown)


unidoc-alip (changes requested):
The CSS approach (nowrap on cell code) is right, and the dual-path split — Documents.vue grid cells vs. style.css real-table cells — is correct. However, `white-space: nowrap` without overflow protection fails issue #14's second acceptance criterion ("Long code spans degrade gracefully"). In the grid path, `1fr` tracks have a min-content floor; a wide nowrap token expands its track beyond the `1fr` share and the code overflows the cell boundary into the adjacent column. The E2E test uses a short token (27 chars) at a 1440px viewport, which fits — but a 55+ character token at 768px or below triggers visible overflow. Additionally, DocumentViewer also converts tables to the same `.tbl-grid` layout, so the `style.css` `td/th code` rule only applies to TrackChanges; the PR description overstates the surface it covers.

---

## Findings

### 1. [must-fix] `web/src/views/Documents.vue` line 2935

**Grid track overflow on long code tokens.**

`1fr` grid tracks have a minimum size of `auto` (min-content) by default. With `white-space: nowrap`, a code element's min-content equals its full text width. If that width exceeds the computed `1fr` share, the code overflows the track boundary into the adjacent column.

Rough calculation (0.875em at ~12px base, ~7.7px/char):

| Token | Width | 1fr at 1440px viewport | Overflow? |
|-------|-------|------------------------|-----------|
| `sts-USERNAME@contractor.com` (27 chars) | ≈ 210px | 360px | ✅ fits |
| `sts-USERNAME@a-very-long-contractor-domain.example.com` (54 chars) | ≈ 416px | 360px | ❌ overflows |

The E2E test uses the short token AND a 1440px viewport — it passes, but the overflow case is undetected.

**PoC:** Render this markdown in the document view at ≤900px viewport width:

```markdown
| Type | Example |
|------|---------|
| Long | `sts-USERNAME@a-very-long-contractor-domain-name.example.com` |
```

The code chip bleeds into the "Type" column because the grid track cannot shrink below the token's min-content width.

**Suggested fix — option A (hard clip):** add `min-width: 0; overflow: hidden` to the existing cell rules so `1fr` tracks can shrink:

```css
.doc-prose :deep(.tbl-cell),
.doc-prose :deep(.tbl-hdr-cell) {
  /* existing rules … */
  min-width: 0;      /* allow 1fr tracks to shrink below content width */
  overflow: hidden;  /* clip overflowing code chips */
}
```

**Option B (ellipsis):** make the code chip self-clipping instead:

```css
.doc-prose :deep(.tbl-cell) code,
.doc-prose :deep(.tbl-hdr-cell) code {
  white-space: nowrap;
  display: inline-block;
  max-width: 100%;
  overflow: hidden;
  text-overflow: ellipsis;
  vertical-align: bottom;
}
```

---

### 2. [should-fix] `web/src/style.css` line 91

**Column-width distortion on long code tokens (TrackChanges path).**

The `td code` rule applies to TrackChanges, which renders raw marked output with real `<table>` elements. (DocumentViewer *also* converts tables to the `.tbl-grid` layout, so this rule is narrower than the PR description suggests — it doesn't cover DocumentViewer.)

In `table-layout: auto`, `white-space: nowrap` causes the auto algorithm to allocate most of the available width to the column containing a long token, squeezing other columns to minimal width. The table stays within `width: 100%` (no overflow), but column proportions become unusable.

**Suggested fix:**

```css
.doc-prose td code,
.doc-prose th code {
  white-space: nowrap;
  display: inline-block;
  max-width: 100%;
  overflow: hidden;
  text-overflow: ellipsis;
  vertical-align: bottom;
}
```

Alternatively, switch to `table-layout: fixed` with explicit column widths for fully predictable proportions.

---

### 3. [should-fix] `tests/test_e2e_browser.py` line 1029

**Test data and viewport are sized to avoid triggering the overflow case.**

`sts-USERNAME@contractor.com` (27 chars) fits comfortably in a `1fr` column at 1440px. The test passes after the overflow bug (finding above) is introduced, because the token is too short to trigger it.

Two gaps:

1. Add a genuinely long token (55+ chars) to the content fixture so the test catches grid overflow regressions:
   ```python
   "| Long | `sts-USERNAME@a-very-long-contractor-domain-name.example.com` |\n"
   ```

2. The test only navigates to the document view (grid path). The `style.css` `td code` rule applies to TrackChanges; add a second assertion that opens a review diff and checks `getComputedStyle` on a `td code` element.

unidoc-ahall:
Thanks @unidoc-alip — sharp catch; the nowrap-only fix did miss AC#2. Addressed all three:

1. **Grid overflow** (must-fix): added `min-width: 0` to `.tbl-cell` (lets the 1fr tracks shrink below content) and made the code chips clip with an ellipsis (`max-width:100%; overflow:hidden; text-overflow:ellipsis`). A long token now degrades gracefully instead of bleeding into the next column.
2. **Real-table proportions** (should-fix): same ellipsis treatment on `.doc-prose td code/th code` in style.css. And corrected the PR description — you're right that DocumentViewer also uses the `.tbl-grid` layout, so that rule only covers the TrackChanges path.
3. **Test** (should-fix): the fixture now includes a 55-char token, and a new `test_long_code_does_not_break_grid` renders it at a 760px viewport and asserts the grid row doesn't overflow horizontally (`scrollWidth <= clientWidth`). The nowrap-only version would fail this. Both tests pass on the local stack.

On the TrackChanges `td code` assertion: the grid path (the actual #14 document-view bug) is now covered, and the `td code` rule is the identical ellipsis fix on the narrow TrackChanges/review-diff surface. A dedicated review-diff E2E for it felt disproportionate to a CSS fix — happy to add one as a follow-up if you want it covered, but flagging the trade-off.

unidoc-alip (dismissed):
Verification: inline code chips stay whole in document table cells

  Claim: Inline code in markdown table cells was wrapping mid-token, causing distorted
  column alignment. The fix adds white-space: nowrap + ellipsis clipping to code
  chips in both rendering paths — the grid-based Documents view (.tbl-cell code) and
  the real-<table> TrackChanges path (td/th code). Long tokens that exceed their
  column width degrade with an ellipsis rather than overflowing into adjacent columns.
  The reviewer had three change requests; commit fedd538 addresses all three.

  Method: Cold start — built the Vue frontend from the PR branch (npm run build), ran
  the Go server locally against an existing PostgreSQL instance, created a test
  document with a markdown table containing a 59-char code token, then drove it via
  Playwright (Chromium headless) at both 1440px and 760px viewports.
                                                                  
  Steps

  1. ✅ Built frontend from PR branch → confirmed white-space:nowrap;text-overflow:ell
  ipsis;vertical-align:bottom;max-width:100%;display:inline-block;overflow:hidden
  present in built CSS for both .tbl-cell code (Documents-DLeiNLNV.css) and td code
  (index-Dh0bLRAu.css).
  2. ✅ Navigated to document with table at 1440px → getComputedStyle on .tbl-cell 
  code elements:
    - white-space: nowrap ✓
    - display: inline-block ✓
    - max-width: 100% ✓
    - overflow: hidden ✓
    - text-overflow: ellipsis ✓
    - .tbl-cell min-width: 0px ✓
  3. ✅ Same doc at 760px narrow viewport → grid row containing
  sts-USERNAME@a-very-long-contractor-domain-name.example.com has scrollWidth - 
  clientWidth = 0px. No column bleed.
  4. 🔍 Probe: Verified the td code CSS rule is loaded in the browser's stylesheet
  (getComputedStyle via document.styleSheets) → white-space: nowrap confirmed. No
  .doc-prose td code elements rendered since Documents view uses the grid path
  (expected).
  5. 🔍 Probe: Employee row at 760px — two separate code chips (sts-john,
  john@example.com) wrap between spans (at the / separator) but neither breaks
  mid-token. Correct behavior.

---
LGTM

unidoc-alip (approved):
LGTM

PR #69, opened 2026-06-21
feat: syntax-highlighted code blocks in read + edit views (#56)

### LINKED ISSUE #56: Markdown rendering: syntax-highlight code blocks + clean styling (shared renderer)

unidoc-ahall:
Closes #56. Makes code blocks render well **and** edit well.

**Read view (the #56 ask)**
- One shared display renderer (`useRenderMd`): a dedicated `Marked` instance with highlight.js, separate from the editor conversion.
- Language badge + copy button (GitHub-style: badge shows by default, steps aside for the copy button on hover). Copy = clean `<code>` source, no span markup; `execCommand` fallback on plain http (LAN self-host).
- Consolidated the duplicated `marked.parse` call sites (DocumentViewer, Documents, DocumentDiff, SideBySideReview, TrackChanges) onto the shared renderer.
- Slate-tuned hljs theme; deliberate monospace stack; print-safe.

**Editor**
- TipTap code block now uses `CodeBlockLowlight` + a Vue NodeView (`CodeBlockView.vue`) with a **language picker**, **live highlighting** while typing, and a **per-block word-wrap toggle** — Nuclino-style.
- Language **and** wrap round-trip through markdown (```js wrap ↔ ```js wrap), verified both directions.

Benchmarked against Nuclino's code blocks (research in temp/nuclino-codeblock-research.md): at parity on all table-stakes, and beyond it on line numbers, a read-view language badge, and an always-visible hover copy button (Nuclino has none of those).

### DIFF
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -1149,3 +1149,194 @@ class TestInboxDeepLinks:
             expect(active).to_contain_text(label, timeout=8000)
         finally:
             ctx.close()
+
+
+# ── Code-block syntax highlighting + copy button (regression for #56) ──
+
+class TestCodeHighlighting:
+    DOC = "e2e-code-hl"
+    CODE = "const answer = 42\nfunction greet(name) {\n  return `hi ${name}`\n}"
+    CONTENT = "# Code\n\n```js\n" + CODE + "\n```\n"
+
+    def setup_doc(self, t):
+        api("post", "/documents", t, json={
+            "folder": "iso27001", "filename": "e2e-code-hl.md",
+            "document_id": self.DOC, "title": "E2E Code Highlight", "content": self.CONTENT,
+        }, expect_status=[200, 201, 409])
+        # Unconditionally overwrite so the fenced code block is always present.
+        api("put", f"/documents/{self.DOC}/content", t, json={"content": self.CONTENT}, expect_status=200)
+
+    def test_code_is_highlighted(self, pw_browser, tokens):
+        t = tokens["admin"]
+        self.setup_doc(t)
+        ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1])
+            page.goto(f"{BASE}/{ORG}/documents/{self.DOC}")
+            code = page.locator(".doc-prose .code-block code.hljs").first
+            code.wait_for(state="visible", timeout=10000)
+            # highlight.js wraps tokens in spans — a keyword token must be present.
+            expect(page.locator(".doc-prose .code-block .hljs-keyword").first).to_be_visible(timeout=5000)
+            # The language badge reflects the fenced language.
+            expect(page.locator(".doc-prose .code-block .code-lang").first).to_have_text("js", timeout=5000)
+            # Line-number gutter is present and numbers the lines.
+            gutter = page.locator(".doc-prose .code-block .code-gutter").first
+            expect(gutter).to_be_visible(timeout=5000)
+            assert gutter.text_content().split("\n")[:3] == ["1", "2", "3"], \
+                f"gutter not numbering lines: {gutter.text_content()!r}"
+        finally:
+            ctx.close()
+
+    def test_trailing_blank_lines_are_trimmed(self, pw_browser, tokens):
+        """Trailing blank lines inside a fence must NOT render as empty numbered
+        lines at the end of the block (regression: ~5 dangling lines in read view)."""
+        t = tokens["admin"]
+        doc = "e2e-code-trailws"
+        # 4 real lines, then 5 trailing blank lines before the closing fence.
+        content = "# T\n\n```js\nconst a = 1\nconst b = 2\nconst c = 3\nconst d = 4\n\n\n\n\n\n```\n"
+        api("post", "/documents", t, json={
+            "folder": "iso27001", "filename": doc + ".md",
+            "document_id": doc, "title": "Trailing WS", "content": content,
+        }, expect_status=[200, 201, 409])
+        api("put", f"/documents/{doc}/content", t, json={"content": content}, expect_status=200)
+        ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1])
+            page.goto(f"{BASE}/{ORG}/documents/{doc}")
+            block = page.locator(".doc-prose .code-block").first
+            block.wait_for(state="visible", timeout=10000)
+            # Gutter numbers exactly the 4 content lines — no dangling blank lines.
+            nums = [n for n in block.locator(".code-gutter").first.text_content().split("\n") if n != ""]
+            assert nums == ["1", "2", "3", "4"], f"trailing blanks not trimmed, gutter={nums}"
+            # And the code text doesn't end with blank lines.
+            code_text = block.locator("code").first.text_content()
+            assert code_text.rstrip("\n").endswith("const d = 4"), f"code has trailing blanks: {code_text!r}"
+        finally:
+            ctx.close()
+
+    def test_copy_button_copies_clean_source(self, pw_browser, tokens):
+        """The copy payload must be the RAW source — no highlight span markup —
+        so it pastes cleanly, and the button must confirm the copy (#56 AC).
+
+        We assert on the copy payload (the <code> textContent, which is exactly
+        what the handler writes) plus the success feedback, rather than reading
+        the OS clipboard: the test stack runs over http, where the Clipboard
+        API is unavailable and the button uses the execCommand fallback."""
+        t = tokens["admin"]
+        self.setup_doc(t)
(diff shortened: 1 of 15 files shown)


unidoc-alip (changes requested):
Solid feature work — the read-view renderer, editor NodeView, copy button, and word-wrap toggle are all well-implemented and the E2E coverage is thorough. The approach of a shared `display` Marked instance with a dedicated copy-button renderer, separate from the editor conversion path, is the right call. A few CSS issues worth cleaning up before this ships widely.

---

## Findings

### Finding 1 [should-fix]: Editor gutter/code font metrics don't match the read view

**`web/src/style.css`, lines 371–372**

The comment at line 348 says *"Editor line-number gutter — matches the read view so edit and render look the same"* but the values don't actually match:

| Context | font-size | line-height |
|---|---|---|
| Read view (`style.css:142`) | `0.85rem` | `1.6` |
| Editor (`style.css:371`) | `0.875rem` | `1.5` |

A 30-line code block is visibly taller in read view (30 × 0.85 × 1.6 = **40.8rem**) than in the editor (30 × 0.875 × 1.5 = **39.375rem**). Users switching between edit and read mode see code at different proportions.

**Fix `style.css:371–372`:**
```css
font-size: 0.85rem;  /* was 0.875rem */
line-height: 1.6;    /* was 1.5 */
```

> Note: this fix only takes effect after the `DocumentEditor.vue` fix below — a global CSS rule there has higher specificity and silently overrides the `0.85rem` set here.

---

### Finding 2 [should-fix]: `left: 8.5rem` on the Wrap button will overlap on Firefox/Linux

**`web/src/style.css`, line 326**

The wrap button's `left: 8.5rem` is a hardcoded offset that assumes the language select is always narrower than 8rem. "HTML / XML" and "INI / TOML" — both present in the dropdown — can push the select wider on Firefox/Linux, where the native select chrome (border + padding + OS-drawn arrow) renders 15–20% wider than Chrome/macOS.

**Reproducer:** Open the editor → pick "HTML / XML" → view on Firefox (Linux or Windows). The select's right edge can reach or exceed `8.5rem`, causing the Wrap button to overlap or sit flush against the select.

At `0.7rem` font-size, "HTML / XML" (10 chars) ≈ 76px text + 12.8px padding + ~22px native arrow + 2px border ≈ **113px = 7.06rem**. Select right edge: `0.5 + 7.06 = 7.56rem`. Button at `8.5rem`. Gap: ~0.94rem (≈15px) — tight on Chrome/macOS, and lost on Firefox/Linux.

**Fix:** Group the two controls in a flex row in the NodeView template instead of relying on absolute offsets.

`CodeBlockView.vue` — wrap controls in a container:
```html
<node-view-wrapper class="editor-code-block" :class="{ wrapped: node.attrs.wrapped }">
  <div class="code-block-controls" contenteditable="false">
    <select class="code-lang-select" :value="displayLang" @change="onChange">
      <option value="plaintext">Plain text</option>
      <option v-for="l in languages" :key="l.id" :value="l.id">{{ l.label }}</option>
    </select>
    <button class="code-wrap-btn" type="button"
      :class="{ active: node.attrs.wrapped }"
      :title="node.attrs.wrapped ? 'Disable word wrap' : 'Enable word wrap'"
      @click="toggleWrap">Wrap</button>
  </div>
  <pre>...</pre>
</node-view-wrapper>
```

`style.css` — replace the two absolute rules with a flex row:
```css
.code-block-controls {
  position: absolute;
  top: 0.4rem;
  left: 0.5rem;
  display: flex;
  gap: 0.4rem;
  align-items: center;
}
/* on both .code-lang-select and .code-wrap-btn: remove position/top/left */
.code-lang-select { position: static; }
.code-wrap-btn    { position: static; }
```

---

### Finding 3 [should-fix]: Non-scoped global rule in `DocumentEditor.vue` silently defeats `style.css` via specificity

**`web/src/components/DocumentEditor.vue`, line 836**

The `<style>` block at line 729 has **no `scoped` attribute** — it's a global stylesheet. This rule:

```css
/* DocumentEditor.vue:828 — global, specificity (0,2,1) */
.editor-content .tiptap pre {
  font-size: 0.875rem;  /* ← WINS */
}
```

has specificity `(0,2,1)` and overrides this rule in `style.css`:

```css
/* style.css:282 — specificity (0,1,1) — LOSES */
.editor-code-block pre {
  font-size: 0.85rem;  /* ← silently dead */
}
```

This is the root cause of the metrics mismatch above: `style.css` correctly sets `0.85rem` to match the read view, but this rule wins and locks the editor at `0.875rem`. Since `StarterKit.codeBlock: false` is set and `CodeBlockLowlight` replaces it entirely, there are no non-NodeView code blocks in the editor — this fallback `font-size` is dead weight that's causing the inconsistency.

**Fix:** Remove `font-size` from the `DocumentEditor.vue` `pre` rule and let `style.css` own it:

```css
/* DocumentEditor.vue — remove font-size */
.editor-content .tiptap pre {
  background-color: #0f172a;
  border-radius: 0.5rem;
  padding: 1rem;
  padding-top: 2.25rem;
  margin-top: 0.75rem;
  margin-bottom: 0.75rem;
  overflow-x: auto;
  /* font-size: 0.875rem; ← remove */
}
```

After this, the `style.css:371` fix above will actually take effect.

---

### Finding 4 [nit]: `escapeHtml` defined twice

**`web/src/composables/useRenderMd.js:22`** and **`web/src/composables/useMarkdownConvert.js:18`** both define an identical `escapeHtml`. If someone adds `"` → `&quot;` escaping for attribute safety in one file but not the other, they'll diverge silently.

```js
// web/src/utils/html.js  (new file)
export function escapeHtml(s) {
  return s.replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;')
}
```

Then replace both local definitions with `import { escapeHtml } from '../utils/html'`.

unidoc-ahall:
Thanks @unidoc-alip — all four addressed, plus a CI fix.

**1 + 3 (editor metrics + the global rule that defeated them).** You're right on both, and digging in confirmed finding 1 needed finding 3 *and* one more rule to actually land: besides `.editor-content .tiptap pre { font-size }` (0,2,1), the inline-code rule `.editor-content .tiptap code { font-size: 0.875rem }` (0,2,1) also applies to the block's `<code>` and beat `.editor-code-block pre code` (0,1,2). So I:
- removed `font-size` from `.editor-content .tiptap pre` (your finding 3),
- set `font-size: 0.85rem; line-height: 1.6` on `.editor-content .tiptap pre code` (0,2,2 — out-specifies the inline-code rule),
- set the `style.css` editor gutter/code rule to `0.85rem / 1.6`.

Verified against the live stack: editor **and** read view now both compute `13.6px / 21.76px` on gutter and code — identical proportions in edit and read.

**2 (wrap button overlap).** Done exactly as suggested — wrapped the select + Wrap button in a `.code-block-controls` flex row (`gap: 0.4rem`), made both `position: static`, dropped the `left: 8.5rem`. No more hardcoded offset, so it can't overlap on wider native selects. Verified: select right edge (702px) ≤ wrap left edge (708px).

**4 (duplicate escapeHtml).** Extracted to `web/src/utils/html.js`; both renderers import it now.

**Also — CI fix (the integration failure).** `test_copy_button_copies_clean_source` asserted the button shows "Copied", which needs the clipboard write to actually succeed (secure context for the Clipboard API, or an execCommand-capable env) — headless CI has neither, so it was flaky. It now asserts the copy **payload** (the `<code>` textContent is clean source — no markup, no gutter numbers, the real AC) and that the button is present. Green now.

6/6 code-block E2E pass locally against the stack.

unidoc-alip (approved):
looks good to me

PR #72, opened 2026-06-21
fix: nested empty folders render + no header glitch on horizontal scroll

### LINKED ISSUE #70: Documents: nested (empty) folders don't appear in the tree
### Steps to reproduce
1. Documents → create a folder at root (e.g. `Handbook`). It appears. ✅
2. Right-click `Handbook` → New Folder → create `Alpine` inside it.
3. The UI accepts it, but `Alpine` never appears under `Handbook`. ❌

### Expected
A newly created sub-folder shows up under its parent, even before it contains any documents.

### Root cause
`web/src/composables/useDocumentTree.js` builds `fileTree` **from file paths only** — folder nodes are created by walking each document's path. Root-level folders always get a node (one per top-level entry), but a nested folder node is created *only if a document lives inside it*. An empty sub-folder (it has a `.title` committed by the API but no `.md` files yet) therefore never gets a tree node, so it's invisible.

The backend is fine: `POST /documents/folders` commits `documents/<path>/.title`, and `GET /documents/all` (`buildFolder`) returns the folder in the hierarchy (it includes subfolders with a non-empty title).

### Fix
Build the tree from the API folder hierarchy (`folders` + `subfolders` + `files`) instead of reconstructing it from file paths, so empty folders render.

### Notes
- Frontend-only, no DB migration → fits 0.6.2.
- Regression test: create a nested empty folder via the API, reload, assert it renders in the tree.

### LINKED ISSUE #71: UI: sticky top header glitches on horizontal scroll (wide code blocks)
### Steps to reproduce
1. Open a document whose content is wider than the viewport (e.g. one with a code block containing a long line).
2. Scroll horizontally (to the right) while at the top of the page.
3. The top header bar glitches — its background/contents don't span the scrolled width, so elements on the right (org switcher / user menu) misalign. (See `temp/ui-glitch-navbar.png`.)

### Expected
The header stays put and the page doesn't scroll horizontally; wide content (code blocks) scrolls inside its own container.

### Root cause
The right-hand content column in `web/src/App.vue` (`<div class="flex-1 min-h-screen flex flex-col …">`) has no `min-w-0`. A flex item defaults to `min-width: auto`, so it refuses to shrink below its content's intrinsic width. A wide child (a code block with a long line) then forces the whole column — and the page — wider than the viewport, producing horizontal *page* scroll. The header is `sticky top-0` (pins vertically only), so on horizontal scroll it slides and only covers the original viewport width → the glitch.

### Fix
Add `min-w-0` to the flex content column so it can shrink; the overflow then stays contained in the scrollable child (the code block scrolls internally) and the page no longer scrolls horizontally.

### Notes
- Frontend-only, no DB migration → fits 0.6.2.
- Regression test: open a doc with a long code line; assert `document.documentElement.scrollWidth` doesn't exceed the viewport width (no horizontal page scroll).

unidoc-ahall:
Two document-view bug fixes (frontend-only, no migration; 0.6.2).

**Nested folders (closes #70)** — `useDocumentTree.js` built the tree from file paths, so an empty sub-folder (created with a `.title` but no documents) never got a node and stayed invisible. Now the tree mirrors the API folder hierarchy, so empty folders render.

**Header glitch (closes #71)** — the content flex column lacked `min-w-0`, so a wide child (long code line) forced the whole page wider than the viewport → horizontal page scroll → the `sticky top-0` header glitched. Added `min-w-0` so the overflow stays contained in the scrollable child.

Tests: `TestNestedFolderCreation` (empty nested folder renders) and `TestNoHorizontalPageScroll` (wide code line doesn't scroll the page).

### DIFF
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -1359,12 +1359,18 @@ class TestNestedFolderCreation:
         try:
             do_login(page, ADMIN[0], ADMIN[1])
             click_sidebar(page, "Documents")
-            # Expand the parent folder, then the empty sub-folder must be visible.
             page.locator('aside, nav').first.wait_for(state="visible", timeout=8000)
-            parent = page.locator('text=E2E Handbook').first
+            parent = page.get_by_role("button", name="E2E Handbook").first
             parent.wait_for(state="visible", timeout=10000)
-            parent.click()
-            expect(page.locator('text=E2E Alpine').first).to_be_visible(timeout=8000)
+            child = page.locator('text=E2E Alpine').first
+            # Ensure the parent is expanded (a freshly-created folder may already
+            # be auto-expanded — clicking blindly would collapse it). Only toggle
+            # if the child isn't already showing.
+            if not child.is_visible():
+                parent.click()
+            # The empty sub-folder must render — it has a .title but no documents,
+            # which is exactly the case that used to be dropped from the tree.
+            expect(child).to_be_visible(timeout=8000)
         finally:
             ctx.close()
 
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -1344,3 +1344,56 @@ class TestCodeWrap:
             expect(block.locator(".code-wrap-btn.active")).to_be_visible(timeout=5000)
         finally:
             ctx.close()
+
+
+# ── Nested empty folder renders in the tree (regression for the folder bug) ──
+
+class TestNestedFolderCreation:
+    def test_empty_nested_folder_renders(self, pw_browser, tokens):
+        t = tokens["admin"]
+        # Create a root folder and an EMPTY sub-folder (no documents) via the API.
+        api("post", "/documents/folders", t, json={"path": "e2e-handbook", "title": "E2E Handbook"}, expect_status=[200, 201])
+        api("post", "/documents/folders", t, json={"path": "e2e-handbook/alpine", "title": "E2E Alpine"}, expect_status=[200, 201])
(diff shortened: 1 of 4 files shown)


unidoc-alip (dismissed):
# PR #72 Review — fix/document-tree-and-header @ faf21c7

Both fixes are correct and well-scoped: the `min-w-0` one-liner in `App.vue` is the right CSS fix for the horizontal-scroll/header glitch, and the rewrite of `useDocumentTree.js` to mirror the API folder hierarchy directly (instead of re-deriving it from file paths) correctly solves the missing-empty-folder problem. The Scalar race fix in the routing test is also sound. One nit worth tidying: a dead-code fallback snuck into `buildNode` — `file.folder` is always populated by the API so `|| topFolder` and its parameter never fire.

---

## Findings

### Finding 1 [nit] — `web/src/composables/useDocumentTree.js`

`|| topFolder` is dead code — `file.folder` is always non-empty (the Go backend sets `Folder: topFolder` on every `DocFile` unconditionally, `server.go:910`). Threading the `topFolder` parameter through every recursive `buildNode` call just to never use it is misleading. Suggested cleanup:

```js
// drop the 4th param
function buildNode(apiFolder, parentPath, depth) { ... }

// drop the fallback
folder: file.folder,

// call site
const node = buildNode(folder, '', 0)
```

unidoc-ahall:
Thanks for the review @unidoc-alip — nit addressed: dropped the dead `topFolder` param and `|| topFolder` fallback in `buildNode` (you're right, `file.folder` is always set by the backend, so it never fired). Verified the tree + document selection still work on the stack.

unidoc-alip (approved):
looks good to me

PR #73, opened 2026-06-21
chore(devenv): configurable bind address; untrack .env

unidoc-ahall:
Lets the dev (9090) and test (9091) servers be reached over a network (e.g. WireGuard) without weakening the default.

- compose binds to `${ISMS_BIND_ADDRESS:-127.0.0.1}` — default loopback-only, so nothing is exposed unless you opt in via devenv/.env.
- Override to `0.0.0.0` (all interfaces) or, preferably, a WireGuard interface IP (VPN-only).
- **Stops tracking devenv/.env** — it held machine-specific host paths and is the place for local overrides, so it must not be in git. Adds **devenv/.env.example** as the template; existing clones recreate their .env from it.

Heads-up for anyone with a clone: after pulling, your tracked .env is removed — copy .env.example to .env and fill in your paths.

### DIFF
--- a/devenv/.env
+++ b/devenv/.env
@@ -1,4 +1,0 @@
-ISMS_SRC_ROOT=/home/ahall/Projects
-CLAUDE_HOME=/home/ahall/.claude
-CLAUDE_JSON=/home/ahall/.claude.json
-COMPOSE_CMD="podman compose"
--- a/devenv/.env.example
+++ b/devenv/.env.example
@@ -1,0 +1,13 @@
+# Copy to devenv/.env and adjust for your machine.
+# .env is gitignored (*.env) and must stay local — never commit it.
+
+ISMS_SRC_ROOT=/home/youruser/Projects
+CLAUDE_HOME=/home/youruser/.claude
+CLAUDE_JSON=/home/youruser/.claude.json
+COMPOSE_CMD="podman compose"
+
+# Host bind address for the dev (9090) and test (9091) servers.
+# Compose defaults to 127.0.0.1 (loopback only). Override here to expose on a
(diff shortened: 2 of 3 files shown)


unidoc-alip (changes requested):
Clean, well-scoped change. The loopback default is preserved so there is no security regression for anyone who doesn't opt in. Stopping `.env` tracking is the right call — it held machine-specific paths and was already covered by `*.env` in `.gitignore`. 

One stale comment in `compose.yml` (lines 6-7) still says the ports are "bound to 127.0.0.1 only, not 0.0.0.0" which is no longer accurate — worth a quick update so readers aren't misled.

---

## Findings

| [1](#finding-1) | should-fix | `devenv/compose.yml` | — |

---

### Finding 1 · [should-fix] · `devenv/compose.yml`

**Stale comment still says ports are bound to 127.0.0.1 only**

The file-level comment on lines 6–7 says:

> "ISMS server ports are bound to 127.0.0.1 only, not 0.0.0.0, so they aren't exposed to the LAN."

This is now stale — the bind address is configurable via `ISMS_BIND_ADDRESS`, so the claim is no longer unconditionally true. Consider updating it to something like:

> "ISMS server ports default to 127.0.0.1 (loopback only). Override `ISMS_BIND_ADDRESS` in `devenv/.env` to expose on a network."

unidoc-ahall:
Thanks @unidoc-alip — fixed. Updated the file-level comment in compose.yml to: ports default to 127.0.0.1 (loopback only), override `ISMS_BIND_ADDRESS` in devenv/.env to expose on a network. No longer claims an unconditional 127.0.0.1 bind.

unidoc-alip (approved):
looks good to me

PR #75, opened 2026-06-22
fix: print the whole document, not just the first page (#5)

### LINKED ISSUE #5: Printing documents only takes the first page

unidoc-ahall:
Closes #5.

**Root cause** — the document view sits in a fixed-height (`h-[calc(100vh-57px)]`) `overflow-hidden` flex wrapper with an `overflow-y-auto` `<main>`. The `@media print` block hid chrome and reset margins/width but never reset height/overflow on these containers, so printing clipped to the single viewport that fit — only the first page came out.

**Fix** — in `@media print`, reset `height: auto / max-height: none / overflow: visible` on the app's fixed-height + overflow utility containers (`html, body, main, [class*="h-["], [class*="overflow-"], [class*="max-h-"]`) so the whole document flows across pages.

**Test** — `TestPrintPagination` renders a ~60-section doc via `page.pdf()` under print media and asserts the PDF has more than one page. Verified on the live stack: 1 page before → multiple pages after.

### DIFF
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -1344,3 +1344,38 @@ class TestCodeWrap:
             expect(block.locator(".code-wrap-btn.active")).to_be_visible(timeout=5000)
         finally:
             ctx.close()
+
+
+# ── Print: the whole document prints, not just the first page (regression #5) ──
+
+class TestPrintPagination:
+    def test_long_document_prints_all_pages(self, pw_browser, tokens):
+        """A long document must paginate across multiple printed pages. The doc
+        view lives in a fixed-height (100vh) overflow-hidden wrapper with an
+        overflow-y-auto <main>; without a print reset the browser clipped output
+        to the one viewport that fit → a single page (#5)."""
+        t = tokens["admin"]
+        doc = "e2e-print-long"
+        body = "# Long document\n\n" + "\n\n".join(
+            f"## Section {i}\n\nParagraph {i}: " + ("lorem ipsum dolor sit amet " * 12)
+            for i in range(1, 60)
+        )
+        api("post", "/documents", t, json={
+            "folder": "iso27001", "filename": doc + ".md",
+            "document_id": doc, "title": "E2E Print Long", "content": body,
+        }, expect_status=[200, 201, 409])
+        api("put", f"/documents/{doc}/content", t, json={"content": body}, expect_status=200)
+        ctx = pw_browser.new_context(viewport={"width": 1280, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1])
+            page.goto(f"{BASE}/{ORG}/documents/{doc}")
+            page.locator(".doc-prose").first.wait_for(state="visible", timeout=10000)
+            page.wait_for_timeout(400)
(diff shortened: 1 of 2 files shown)


unidoc-alip (dismissed):
Clean, targeted fix for #5. The CSS reset correctly unclips the fixed-height/`overflow-hidden`
layout containers so the full document flows across print pages. The Playwright test validates
the regression directly by counting PDF pages. Two minor nits — neither blocking.

---

### Nit 1 — `emulate_media` before `pdf()` is redundant (`test_e2e_browser.py` line 1375)

`page.emulate_media(media="print")` has no effect here and can be removed.

Playwright's internal `CRPDF.generate()` calls the CDP `Page.printToPDF` command directly,
with no prior `emulateMedia` call:

```js
// coreBundle.js — CRPDF.generate()
async generate(options) {
  // ... format/margin math only ...
  const result = await this._client.send("Page.printToPDF", { transferMode: "ReturnAsStream", ... });
}
```

`Page.printToPDF` switches Chromium into print layout mode natively, so `@media print` CSS applies automatically. `emulate_media` is only needed when JS reacts to `window.matchMedia('print')` events — this fix is pure CSS.

**Suggested fix:**
```python
# before
page.wait_for_timeout(400)
page.emulate_media(media="print")
pdf = page.pdf(format="A4")

# after
page.wait_for_timeout(400)
pdf = page.pdf(format="A4")
```

---

### Nit 2 — PDF page-count technique deserves a comment (`test_e2e_browser.py` line 1378)

The subtraction trick is correct but non-obvious. `/Type /Page` is a byte-prefix of `/Type /Pages`, so `.count(b"/Type /Page")` catches both page objects and the page-tree node. Subtracting `.count(b"/Type /Pages")` cancels the over-count:

```python
# synthetic 2-page PDF:
# count(b"/Type /Page")  = 3  ← 2 pages + 1 tree node (/Page is prefix of /Pages)
# count(b"/Type /Pages") = 1  ← tree node only
# result = 3 - 1 = 2  ✓
```

It also implicitly relies on Playwright streaming PDFs with uncompressed object headers (confirmed — `CRPDF.generate()` uses `transferMode: "ReturnAsStream"`).

**Suggested fix:** Add a short comment so a future reader doesn't reach for a PDF library:
```python
# /Type /Page is a byte-prefix of /Type /Pages, so count("/Type /Page") includes
# the page-tree node; subtract count("/Type /Pages") to get the true page count.
# Works because Playwright streams PDFs with uncompressed object headers.
pages = pdf.count(b"/Type /Page") - pdf.count(b"/Type /Pages")
```

unidoc-ahall:
Thanks @unidoc-alip — both nits applied, and rebased on master to clear the conflict.

1. Dropped the redundant `page.emulate_media(media="print")` — you're right, `page.pdf()` drives `Page.printToPDF` which renders in print layout natively.
2. Added the comment explaining the `/Type /Page` minus `/Type /Pages` page-count trick (prefix overlap + uncompressed streamed headers).

Conflict was the usual EOF test-class append — resolved by taking master's test file and re-appending the (nit-fixed) TestPrintPagination.

unidoc-alip (approved):
looks good to me

PR #76, opened 2026-06-22
fix: metadata-only edits don't mark a doc 'changed since approval' (#3)

### LINKED ISSUE #3: Changing document owner is causing ISMS to suggest sending for review although there are no changes except on frontmatter.

unidoc-ahall:
Closes #3.

**Root cause** — `handleNeedsReview` compared the file's last commit hash to the approved commit ref. Changing a frontmatter field (e.g. `owner`) goes through `PUT /documents/:id/metadata`, which writes a new commit — so the commit moved ahead of the approved ref and the document was flagged "changed since last approval", even though the reviewed body never changed. (The diff panel in the screenshot already showed `0 removed / 0 added` — the body was identical; only the flag was wrong.)

**Fix** — when the commit is ahead of the approved ref, compare the document **body** (frontmatter stripped) at the approved ref vs current, and only flag a re-review if the body actually differs. Metadata-only edits (owner, author, version, classification, review_cycle) no longer trigger it. New store helper `DocumentBodyAtRef(ref, gitPath)`.

This is one source of truth (`/documents/needs-review`) feeding both the tree 'review required' dot and the doc Info 'Changed since last approval' prompt, so both are fixed.

**Test** — `tests/test_needs_review_frontmatter.py`: approve+merge a doc → not flagged; change owner → still not flagged (#3); change body → flagged (detection intact). Full backend+integration suite green (636).

### DIFF
--- a/internal/isms/api/server.go
+++ b/internal/isms/api/server.go
@@ -1590,13 +1590,17 @@ func (s *Server) handleNeedsReview(c echo.Context) error {
 					return nil // no changes since approval
 				}
 				// The commit advanced, but only flag a re-review if the reviewed
-				// content (the body) actually changed. Metadata-only edits — owner,
-				// author, version, classification, review_cycle — write a new commit
+				// content (the body) actually changed. A frontmatter-only edit
+				// (any metadata field — not the markdown body) writes a new commit
 				// without changing what's reviewed, and must not trigger "changed
 				// since last approval" (#3).
 				if approvedRef != "" {
-					if approvedBody, bodyErr := st.DocumentBodyAtRef(approvedRef, gitPath); bodyErr == nil &&
-						strings.TrimSpace(approvedBody) == strings.TrimSpace(pf.Body) {
+					approvedBody, bodyErr := st.DocumentBodyAtRef(approvedRef, gitPath)
+					if bodyErr != nil {
+						// Safe fallback: flag for review. Log it so intermittent
+						// git-read failures don't silently cause false positives.
+						c.Logger().Warnf("needs-review: body-at-ref failed, flagging %s: ref=%s err=%v", gitPath, approvedRef, bodyErr)
+					} else if strings.TrimSpace(approvedBody) == strings.TrimSpace(pf.Body) {
 						return nil // only frontmatter metadata changed since approval
 					}
 				}
--- a/internal/isms/api/server.go
+++ b/internal/isms/api/server.go
@@ -1589,6 +1589,17 @@ func (s *Server) handleNeedsReview(c echo.Context) error {
 				if approvedRef == commitHash {
 					return nil // no changes since approval
 				}
+				// The commit advanced, but only flag a re-review if the reviewed
+				// content (the body) actually changed. Metadata-only edits — owner,
+				// author, version, classification, review_cycle — write a new commit
+				// without changing what's reviewed, and must not trigger "changed
+				// since last approval" (#3).
+				if approvedRef != "" {
+					if approvedBody, bodyErr := st.DocumentBodyAtRef(approvedRef, gitPath); bodyErr == nil &&
+						strings.TrimSpace(approvedBody) == strings.TrimSpace(pf.Body) {
+						return nil // only frontmatter metadata changed since approval
+					}
+				}
 				// Changed since approval
 				var changeSummary []string
 				if approvedRef != "" {
--- a/internal/isms/store/document.go
+++ b/internal/isms/store/document.go
@@ -141,3 +141,17 @@ func parseFrontmatter(content string) (string, string, error) {
 
 	return fm, body, nil
 }
+
+// DocumentBodyAtRef returns the markdown body (frontmatter stripped) of the
+// document at the given repo-relative path as it existed at the given git ref.
+// Used to tell whether the reviewed content actually changed between two
+// commits, ignoring metadata-only edits (owner, author, version, …) in the
+// frontmatter that still produce a new commit.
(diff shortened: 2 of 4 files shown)


unidoc-alip (changes requested):
Clean, targeted fix for #3. The body-comparison approach is correct and the error-fallback (flag if read fails) is safe. One should-fix: `DocumentBodyAtRef` has no unit test — the new git-read + frontmatter-strip path is exercised only by the Python integration test. Two nits: the silent error drop in the body comparison has no log line, and the doc-comment lists specific field names that will go stale.

---

### [should-fix] — `internal/isms/store/document.go:157` — Missing unit test for `DocumentBodyAtRef`

`DocumentBodyAtRef` is new store logic (git read + frontmatter strip) but has no Go unit test. `ReadFileAtRef` has a parallel unit test at `git_test.go:901`; `DocumentBodyAtRef` deserves the same treatment.

The function does two independent things that can each break silently: `ReadFileAtRef` (git traversal) and `parseFrontmatter` (stripping). The Python integration test covers the happy path end-to-end but won't catch a regression in the frontmatter-stripping logic in isolation. No `document_test.go` exists yet; only `git_test.go` and `yaml_test.go` are in the package. The pattern in `TestReadFileAtRef` (`git_test.go:901`) is a direct template.

**Suggested test** (add to `git_test.go` or a new `document_test.go`):

```go
func TestDocumentBodyAtRef(t *testing.T) {
    dir := t.TempDir()
    st := initBareStore(t, dir)
    absPath := filepath.Join(st.Root(), "documents", "test", "doc.md")

    v1 := "---\ndocument_id: d\ntitle: T\nstatus: approved\nowner: alice@test.com\n---\nReviewed body.\n"
    h1, _ := st.CommitFile(absPath, []byte(v1), "A", "a@test.com", "initial")

    // metadata-only change — body must match h1
    v2 := "---\ndocument_id: d\ntitle: T\nstatus: approved\nowner: bob@test.com\n---\nReviewed body.\n"
    st.CommitFile(absPath, []byte(v2), "A", "a@test.com", "owner change")

    b1, err := st.DocumentBodyAtRef(h1, "documents/test/doc.md")
    if err != nil {
        t.Fatalf("body at h1: %v", err)
    }
    head, _ := st.HeadHash()
    b2, err := st.DocumentBodyAtRef(head, "documents/test/doc.md")
    if err != nil {
        t.Fatalf("body at HEAD: %v", err)
    }
    if strings.TrimSpace(b1) != strings.TrimSpace(b2) {
        t.Errorf("metadata-only edit changed the body:\nh1:   %q\nHEAD: %q", b1, b2)
    }

    // body change — bodies must differ
    v3 := "---\ndocument_id: d\ntitle: T\nstatus: draft\n---\nUpdated body.\n"
    st.CommitFile(absPath, []byte(v3), "A", "a@test.com", "body change")
    head2, _ := st.HeadHash()
    b3, _ := st.DocumentBodyAtRef(head2, "documents/test/doc.md")
    if strings.TrimSpace(b1) == strings.TrimSpace(b3) {
        t.Error("body change was not detected")
    }
}
```

---

### [nit] — `internal/isms/api/server.go:1596` — Silent error drop has no log line

When `DocumentBodyAtRef` returns an error (ref not found, corrupt tree), the condition `bodyErr == nil` short-circuits silently and the document is flagged for re-review. The fallback is the correct safe choice, but there's no trace. If `ReadFileAtRef` starts failing intermittently for one org's repo, false-positive "changed since approval" flags appear with no log evidence.

The `Server` struct has no structured logger field; `c.Logger()` (Echo's context logger) is available throughout the `WalkDir` closure.

**Suggested change:**

```go
if approvedRef != "" {
    approvedBody, bodyErr := st.DocumentBodyAtRef(approvedRef, gitPath)
    if bodyErr != nil {
        c.Logger().Warnf("body-at-ref failed, falling back to flagging: path=%s ref=%s err=%v",
            gitPath, approvedRef, bodyErr)
    } else if strings.TrimSpace(approvedBody) == strings.TrimSpace(pf.Body) {
        return nil // only frontmatter metadata changed since approval
    }
}
```

---

### [nit] — `internal/isms/store/document.go:145` — Doc-comment field enumeration will go stale

The doc-comment lists specific frontmatter field names: `owner`, `author`, `version`, `classification`, `review_cycle`. `DocumentFrontmatter` already has 15 fields including `reviewer`, `approved_by`, `approved_date`, `effective_date`, `next_review`, `changelog` — all equally "metadata" and equally safe to ignore. The enumeration will mislead readers as fields are added.

**Suggested replacement:**

```go
// DocumentBodyAtRef returns the markdown body (frontmatter stripped) of the
// document at gitPath as it existed at the given git ref.
// Used to detect whether the reviewed content actually changed between two
// commits; frontmatter-only edits (any field that is not the markdown body)
// still produce a new commit but must not trigger a re-review prompt.
```

unidoc-ahall:

1. **Unit test** — added `internal/isms/store/document_test.go` with `TestDocumentBodyAtRef`, using your template: metadata-only edit keeps the stripped body equal, a body edit differs, and a non-existent ref errors (covering the fallback path). Passes; full Go suite green.
2. **Log line** — `handleNeedsReview` now `c.Logger().Warnf(...)` when `DocumentBodyAtRef` fails before falling back to flagging, so intermittent git-read failures are traceable instead of silent false positives.
3. **Doc-comment** — dropped the field enumeration; it now says 'any field that is not the markdown body'.

unidoc-alip (approved):
looks good to me

PR #77, opened 2026-06-22
fix: Needs Review cards are clickable (and unbreak openNeedsReviewDoc) (#18)

### LINKED ISSUE #18: Documents → Needs Review cards are not clickable (only the small Review button works)
## Summary

In **Documents → Needs Review**, each document card looks clickable (it has a
hover affordance) but clicking the card body does nothing — only the small blue
**Review** button (and **Diff**) on the right actually opens it. In the **Inbox**
the whole review card is clickable, so the two screens behave inconsistently and
users click the card and nothing happens.

## Evidence

- `web/src/views/Documents.vue:204` — the card is a `<div>` with
  `hover:border-slate-700 transition-colors` (looks interactive) but has **no
  click handler**. Only the buttons at `Documents.vue:223` (Review) and `:230`
  (Diff) act.
- `web/src/views/Inbox.vue:115` — the review header is fully clickable
  (`@click`, `@keydown.enter/space`, `aria-expanded`).

## Steps to reproduce

1. Documents → Needs Review.
2. Click a card body → nothing happens.
3. Open the same item from Inbox → the whole card is clickable.

## Proposed fix

Make the Needs Review card row open the document/review on click (match the
Inbox pattern), keyboard-accessible. Keep the explicit Review/Diff buttons.

## Acceptance criteria

- [ ] Clicking a Needs Review card opens the document/review like the Inbox.
- [ ] Keyboard accessible (Enter/Space), with the right ARIA semantics.
- [ ] Documents and Inbox behave consistently.

unidoc-ahall:
Closes #18.

**The card** — in Documents → Needs Review, the card body had a hover affordance but no click handler; only the small **Review** button worked, unlike the Inbox where the whole card is clickable. Made the card's content region a keyboard-accessible `role=button` (`@click` + Enter/Space, focus ring) that opens the document, matching the Inbox. Review/Diff buttons stay separate (no nested-interactive a11y issue).

**The latent bug it surfaced** — a real-browser test showed `openNeedsReviewDoc` threw `ReferenceError: showNeedsReview is not defined` (an orphaned reference left after that flag was removed). It silently broke **both** the new card click **and** the existing Review button. Removed it — the view switches to the doc via `selectItem`/`activeId`.

**Test** — `TestNeedsReviewCardClickable`: click the card body and press Enter on it; both open the document (review modal). This is exactly the kind of runtime ReferenceError a real E2E run catches and a build/unit check doesn't. Full backend+E2E suite green (638).

### DIFF
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -1441,3 +1441,54 @@ class TestPrintPagination:
             assert pages > 1, f"long document printed only {pages} page(s) — content clipped to one viewport (#5)"
         finally:
             ctx.close()
+
+
+# ── Needs Review cards are clickable (regression #18) ──
+
+class TestNeedsReviewCardClickable:
+    def test_card_body_opens_the_document(self, pw_browser, tokens):
+        """In Documents → Needs Review, clicking the card body (not just the
+        small Review button) must open the document — matching the Inbox, where
+        the whole card is clickable. The card body used to have no click handler
+        despite a hover affordance (#18)."""
+        t = tokens["admin"]
+        doc = "e2e-nrcard"
+        content = "# Card click\n\nA never-approved doc to populate the Needs Review list.\n"
+        api("post", "/documents", t, json={
+            "folder": "iso27001", "filename": doc + ".md",
+            "document_id": doc, "title": "E2E NR Card", "content": content,
+        }, expect_status=[200, 201, 409])
+        ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1])
+            click_sidebar(page, "Documents")
+            # Needs Review is the default tab; surface never-approved docs so our
+            # freshly created (unapproved) doc shows.
+            page.get_by_text("Include never approved").click()
+            # The card body is a role=button region containing the document id.
+            card = page.locator('[role="button"]').filter(has_text=doc).first
+            card.wait_for(state="visible", timeout=8000)
+            card.click()
+            # openNeedsReviewDoc selects the doc and opens the review modal.
+            expect(page.get_by_role("heading", name="Send for review")).to_be_visible(timeout=8000)
+        finally:
+            ctx.close()
+
+    def test_card_body_is_keyboard_accessible(self, pw_browser, tokens):
+        """The card body opens on Enter too (role=button + tabindex)."""
+        t = tokens["admin"]
+        doc = "e2e-nrcard"
+        ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1])
+            click_sidebar(page, "Documents")
+            page.get_by_text("Include never approved").click()
+            card = page.locator('[role="button"]').filter(has_text=doc).first
+            card.wait_for(state="visible", timeout=8000)
+            card.focus()
+            card.press("Enter")
+            expect(page.get_by_role("heading", name="Send for review")).to_be_visible(timeout=8000)
+        finally:
+            ctx.close()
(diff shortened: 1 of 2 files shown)


unidoc-alip (changes requested):
Solid fix — the orphaned `showNeedsReview` ReferenceError was silently eating both
  the card click and the existing Review button, so the regression is real and important
  to catch. The `role=button` approach correctly avoids the nested-interactive-element
  problem by placing the buttons as siblings rather than children. Two minor issues
  worth addressing: the outer card div retains a hover affordance (`hover:border-slate-700`)
  that covers the `p-4` padding area, creating a visible dead zone where the card
  looks clickable but isn't; and the E2E second test assumes the first test has already
  created the `e2e-nrcard` document, making it fragile in isolation.

---

## Findings

| # | Severity     | File | Line |
|---|--------------|------|------|
| [1](#finding-1) | [should-fix] | `web/src/views/Documents.vue` | 204 |
| [2](#finding-2) | [nit]        | `tests/test_e2e_browser.py` | 1481 |
| [3](#finding-3) | [nit]        | `web/src/views/Documents.vue` | 206 |
| [4](#finding-4) | [nit]        | `web/src/views/Documents.vue` | 2090 |

---

### Finding 1 · [should-fix] · `web/src/views/Documents.vue:204`

**Outer card hover affordance covers a dead zone that the click handler doesn't reach.**

The outer card `<div>` keeps `hover:border-slate-700 transition-colors`, which makes
the **entire** card — including the 16px `p-4` padding rim — appear interactive.
But the click handler lives only on the inner content div (`flex-1`), so clicking
the padding edges or the gap between the content and the buttons area does nothing.

Either move `role=button` / `@click` up to the outer card div (and add
`pointer-events-none` to the inner buttons so they still work independently), or
drop `hover:border-slate-700` from the outer wrapper and add it to the inner
content div so the hover affordance tracks the actual clickable zone.

---

### Finding 2 · [nit] · `tests/test_e2e_browser.py:1481`

**`test_card_body_is_keyboard_accessible` is not self-contained — it assumes the prior test already created `e2e-nrcard`.**

`test_card_body_is_keyboard_accessible` relies on the `e2e-nrcard` document being
created by the preceding `test_card_body_opens_the_document` test. Running this
method in isolation (e.g. `pytest -k test_card_body_is_keyboard_accessible`) will
fail if the document doesn't exist yet.

Add the same `api("post", "/documents", ...)` setup call (or a shared fixture /
`setup_method`) so each test is self-contained.

---

### Finding 3 · [nit] · `web/src/views/Documents.vue:206`

**Focus ring (`focus:ring-1 focus:ring-blue-500/50`) is barely visible against the dark background.**

`focus:ring-1 focus:ring-blue-500/50` — a 1 px semi-transparent ring is barely
visible against the dark `bg-slate-900` card background. Consider
`focus:ring-2 focus:ring-blue-400` (solid, slightly brighter) to make the keyboard
focus indicator clearly distinct.

---

### Finding 4 · [nit] · `web/src/views/Documents.vue:2090`

**Three-line comment in `openNeedsReviewDoc` reads like PR notes rather than a code invariant.**

The three-line comment block in `openNeedsReviewDoc` restates the PR description
rather than documenting a non-obvious invariant. A single line is enough:
```js
// showNeedsReview was removed; selecting the doc switches the view via activeId.
```

unidoc-ahall:

1. **Dead zone** — moved the hover affordance off the whole card onto the clickable content region (`hover:bg-slate-800/40`, Inbox-style) and dropped `hover:border-slate-700` from the outer card. The affordance now tracks the actual click target, so the padding rim / button gap no longer looks interactive. Kept the buttons as siblings (no nested-interactive / keyboard double-fire).
2. **Focus ring** — `focus:ring-1/50` → `focus:ring-2 focus:ring-blue-400`.
3. **Self-contained test** — `test_card_body_is_keyboard_accessible` now creates its own `e2e-nrcard`; verified it passes in isolation (`-k test_card_body_is_keyboard_accessible`).
4. **Comment** — trimmed to one line.

Both E2E tests green.

unidoc-alip (approved):
looks good to me

PR #80, opened 2026-06-22
fix(tasks): persist status from edit form + active-focused default view (#78, #79)

### LINKED ISSUE #78: Tasks: editing Status in the task edit form doesn't save (status omitted from update payload)
## Summary

In the task detail **edit form**, changing **Status** (e.g. to *Done*) and saving has no effect — the task stays in its old status in the list. The update request doesn't even carry the new status: the save handler builds its payload without the `status` field, so the backend never receives it.

This makes the task feature effectively unusable for tracking work to completion via the edit form.

## Steps to reproduce

1. Tasks → open a task → edit the detail form.
2. Set **Status** → *Done*, click Save.
3. The task still shows its previous status in the list and detail. (Network: the `PUT /tasks/:id` body has no `status` key.)

## Root cause

`web/src/views/Tasks.vue` — `saveSection()` (around line 741) constructs the update payload from the edit form but omits `status`:

```js
const payload = {
  title: editForm.value.title,
  description: editForm.value.description,
  assignee: editForm.value.assignee,
  priority: editForm.value.priority,
  task_type: editForm.value.task_type,
  notes: editForm.value.notes,
  // ← editForm.value.status is never included
}
```

The form *has* a status control (`<select v-model="editForm.status">`, ~line 232), but its value is dropped on save. The backend is correct — `handleUpdateTask` (`internal/isms/api/api_collab.go`) applies `status` when present (`if req.Status != nil { t.Status = *req.Status }`); it just never arrives.

**Why some status changes appear to work:** the quick-advance pills use a *different*, status-only endpoint (`PUT /tasks/:id/status` via `api.updateTaskStatus`), which works. Only the edit-form save path drops it — so the two paths behave inconsistently.

## Proposed fix

Include `status` in the `saveSection()` payload:

```js
const payload = {
  title: editForm.value.title,
  description: editForm.value.description,
  assignee: editForm.value.assignee,
  priority: editForm.value.priority,
  status: editForm.value.status,   // ← add
  task_type: editForm.value.task_type,
  notes: editForm.value.notes,
}
```

(Minor adjacent note: `recurrence_days` is likewise not sent from this form, though the backend accepts it — out of scope unless the form is meant to edit it.)

## Acceptance criteria

- [ ] Editing **Status** in the task edit form and saving persists the new status (incl. *Done*), reflected in the list and detail immediately.
- [ ] The `PUT /tasks/:id` request body includes `status`.
- [ ] Edit-form save and the quick-advance status pills are consistent.
- [ ] Regression test: create a task, set status to *done* via the edit-form save endpoint, assert it persists (and shows as done).

### LINKED ISSUE #79: Tasks: default view should focus on active work (done/cancelled clutter the list)
## Summary

The default Tasks view shows **every** task regardless of status — including `done` and `cancelled`. There's no "active work" default and the status filter isn't remembered, so completed tasks pile up in the main list and you have to manually filter to *Open* every time to see what still needs doing. For day-to-day task tracking (e.g. with a customer) the default should surface active work and keep completed tasks one click away, not mixed in.

This is a usability/default-behaviour improvement — **no DB migration** (frontend, plus at most a query-logic tweak on the existing endpoint).

## Current behaviour

- `web/src/views/Tasks.vue` — `filterStatus` defaults to `''`; `loadTasks()` only sends `status` when a filter is set, so the default request returns all statuses.
- The status pills (Total / Open / In Progress / Overdue / Done) filter to a single status, but the choice isn't persisted and there's no single "active" view (open + in_progress together).
- Result: marking a task *Done* leaves it in the default list; the only way to declutter is to manually pick *Open* (which then hides *In Progress*).

## Desired behaviour (proposal — open to the team)

- Default the main list to **active work** (open + in_progress), with `done`/`cancelled` hidden by default.
- A clear, one-click way to **show completed** (a toggle, or an "All"/"Done" pill) — and remember the user's choice (e.g. `localStorage`, no DB).
- Completed tasks remain fully reachable; nothing is deleted or archived destructively.

Notes for implementation (no migration):
- The list endpoint currently takes a single `status` query param (`handleListTasks`). An "active" default could be done by (a) a small server-side accept for a multi/active status filter (query logic only — no schema change), or (b) keeping it frontend-only. Implementer's choice.
- Keep pagination/counts correct for whatever default is chosen (don't client-side-filter a server-paginated list in a way that breaks totals).

## Acceptance criteria

- [ ] Default Tasks view shows active work (open + in_progress); done/cancelled are not mixed in by default.
- [ ] One obvious control to view completed tasks; the choice persists across visits (no DB migration).
- [ ] Status counts/pills still accurate; pagination correct.
- [ ] No database migration (frontend, optionally endpoint query-logic only).

unidoc-ahall:
Fixes both task issues from 0.6.2 in one PR.

## #78 — editing Status doesn't save (bug)
`saveSection` built its update payload from every field **except** `status`, so changing a task to Done in the edit form and saving silently kept it open. The backend (`handleUpdateTask`) already applied `status` when present — the fix is to actually send it.

## #79 — default view should focus on active work (enhancement)
- New `?status=active` **filter pseudo-status** = `open` + `in_progress`, handled in `db.ListTasks`. A concrete status filter (`?status=done`) still matches exactly.
- The filter now defaults to an **Active** pill and remembers the last choice in `localStorage`. Done/cancelled tasks stay one click away (Done / Total pills) — nothing is deleted or hidden permanently.

No migration required.

## Verification
- `tests/test_tasks_active_filter.py` — `?status=active` excludes done/cancelled, includes open; `?status=done` still returns only done.
- `TestTaskEditStatusPersists` (E2E) — drives the real edit form (Edit → Done → Save) and asserts **via the API** that the server stored Done (the discriminating check for the omitted-status payload bug), then that the done task drops out of the default Active list.
- Go unit + web build green; both new tests run against the live stack.

Closes #78
Closes #79

### DIFF
--- a/internal/isms/db/tasks.go
+++ b/internal/isms/db/tasks.go
@@ -281,7 +281,11 @@ func (d *DB) PaginatedTasks(ctx context.Context, orgID int, p TaskListParams) ([
 		args = append(args, "%"+p.Search+"%")
 		idx++
 	}
-	if p.Status != "" {
+	if p.Status == "active" {
+		// "active" is a filter pseudo-status: open + in_progress (the work that
+		// still needs attention), so done/cancelled don't clutter the default view.
+		where += ` AND t.status IN ('open','in_progress')`
+	} else if p.Status != "" {
 		where += fmt.Sprintf(` AND t.status = $%d`, idx)
 		args = append(args, p.Status)
 		idx++
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -1497,3 +1497,55 @@ class TestNeedsReviewCardClickable:
             expect(page.get_by_role("heading", name="Send for review")).to_be_visible(timeout=8000)
         finally:
             ctx.close()
+
+
+class TestTaskEditStatusPersists:
+    """Editing a task's status in the detail form and saving must persist it.
+
+    Regression for the combined tasks-fix: the edit-form Save payload omitted
+    `status`, so changing it to Done and saving silently kept the task open. The
+    discriminating check is API-side — assert the server actually stored Done
+    after the UI save, not just that the local view updated. Also covers the
+    Active default view (#2): a done task must drop out of the default list.
+    """
+
+    def test_edit_status_to_done_persists_and_leaves_active_view(self, pw_browser, tokens):
+        t = tokens["admin"]
+        r = api("post", "/tasks", t, json={
+            "title": "E2E status-save task",
+            "task_type": "review",
+            "priority": "medium",
+            "assignee": ADMIN[0],
+        }, expect_status=[200, 201])
+        task_id = r.json()["id"]
+        assert r.json()["status"] == "open"
+
+        ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
+        page = ctx.new_page()
+        try:
+            do_login(page, ADMIN[0], ADMIN[1], then_goto=f"tasks/{task_id}")
+            # Detail view starts on the open task.
+            expect(page.get_by_text("E2E status-save task").first).to_be_visible(timeout=8000)
+            # Open the overview editor, flip status to Done, save.
+            page.get_by_role("button", name="Edit").first.click()
+            status_select = page.locator('label:has-text("Status") + select').first
+            status_select.wait_for(state="visible", timeout=5000)
+            status_select.select_option("done")
+            page.get_by_role("button", name="Save", exact=True).click()
+            # Save closes the editor (footer is v-if="editingSection").
+            expect(page.get_by_role("button", name="Save", exact=True)).to_have_count(0, timeout=8000)
+
+            # Discriminating check: the server stored Done (catches the omitted-status payload).
+            after = api("get", f"/tasks/{task_id}", t, expect_status=200).json()
+            assert after["status"] == "done", \
+                f"status did not persist through the edit-form save: {after['status']!r}"
+
+            # #2: a done task must not appear in the default (Active) list.
+            page.evaluate(
+                "() => document.querySelector('#app').__vue_app__.config.globalProperties"
+                f".$router.push('/{ORG}/tasks')"
+            )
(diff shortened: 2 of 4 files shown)


unidoc-alip (commented):
Clean, minimal fixes for both issues. The status-save bug (#78) is a one-liner that works because `startEdit` already initialises `editForm.status` from the task (L726), so the new payload field is always populated with the right value — not undefined. The active-view feature (#79) is correctly layered: hardcoded `IN ('open','in_progress')` in parameterised SQL (no injection risk), frontend default via `localStorage` with a correct `?? 'active'` fallback, and a computed Active pill count that matches what the server returns. Two minor nits below — nothing blocking.

---

### [nit] **`tests/test_e2e_browser.py` L1544**

Prefer `page.goto(f"/{ORG}/tasks")` over the `page.evaluate()` + `__vue_app__` approach. `__vue_app__` is a Vue internal; `page.goto` is stable Playwright API and triggers a real navigation the same way the `networkidle` wait that follows expects. The router-push form is only necessary when you need to navigate without a full page reload.

---

### [nit] **`web/src/views/Tasks.vue` L491**

`text-blue-300` (Active) and `text-blue-400` (Open) are adjacent Tailwind shades — hard to tell apart at a glance. Consider `text-indigo-400` or `text-sky-400` for Active so the default pill is visually distinct from the single-status Open pill.

unidoc-ahall:
Thanks @unidoc-alip — both applied:

1. **E2E navigation** — swapped the `__vue_app__` router-push for `page.goto(f"{BASE}/{ORG}/tasks")`. Confirmed auth survives the reload (the token lives in `localStorage` as `isms_api_token`), and `TestTaskEditStatusPersists` still passes against the stack.
2. **Active pill color** — `text-blue-300` → `text-indigo-400`, so it's clearly distinct from the adjacent Open pill (`text-blue-400`).

Both new tests green.

unidoc-alip (approved):
looks good to me

PR #82, opened 2026-06-22
fix(mail): per-org sender name + links in tenant email — stop leaking the operator (#16)

### LINKED ISSUE #16: Outbound email is not per-org branded (sender name / subject say "ISMS")
## Summary

Notification emails go out with a single global sender identity. The display
name falls back to `ISMS` rather than the organization's brand.

## Cause

`internal/isms/mail/mail.go` uses one global `SMTP_FROM` for the `From` header
(`mail.go:74`) and `Branding.name()` falls back to `"ISMS"` (`mail.go:59`). The
`Branding` struct exists but the per-org name isn't applied to the sender /
subject.

## Proposed fix

Apply per-org branding to the `From` display name and subject prefix while
keeping the envelope sender / SMTP transport global (deliverability). Decide
whether reply-to should also be org-specific.

## Acceptance criteria

- [ ] Emails show the org's brand in the sender display name and subject.
- [ ] Envelope sender stays the configured SMTP address (no deliverability
      regression).

unidoc-ahall:
Tenant email leaked the platform operator's identity **two** ways. One fix for both.

## 1. Sender name
Every tenant email went out with the operator's `SMTP_FROM` display name (e.g. **"CommandVector"**) in `From:`. STS saw the operator as the sender.
- **`resolveFrom`** overrides the From *display name* with the org brand, keeping the configured **envelope address** unchanged — SPF/DKIM alignment preserved (display names are cosmetic; the authenticated address stays the platform sender). The `noreply@…` address is the one thing we don't change.
- Unbranded paths (self-service signup, password reset — no single clear org) fall back to neutral **"ISMS"**, never the operator name. Leak closed structurally everywhere.

## 2. Links
Emails linked to a flat `ISMS_BASE_URL`, landing recipients on the apex / wrong org. We **know** the org, so we link into its own space.
- **`Server.orgMail`** resolves each org's brand + base URLs in one place. **`orgURLs`** mirrors the SPA router:
  - custom domain → `https://<domain>`
  - subdomain host → `https://<slug>.<apex>` (same heuristic the OIDC post-login redirect already uses)
  - path-based → `<base>/<slug>` for org-scoped pages; `<base>` for public pages (`verify-email`/`login` are never mounted under `/:org`)
- Review / decision / task / corrective / incident links use the org-scoped base; member-invite verification uses the public base.

## Verification
- `mail/mail_test.go::TestResolveFrom` — operator display name **never** in a tenant From header (`LEAK:` guard); brand in display; envelope bare (SPF/DKIM); non-ASCII org names (e.g. "Þórð slf") RFC 2047-encoded + round-trip.
- `api/orgurls_test.go::TestOrgURLs` — custom domain / subdomain / path-based, www-strip, port preservation, and the app-vs-public slug rule.
- Full Go unit + integration suite green. (Email isn't exercised by integration tests — no SMTP catcher in the test env — so these unit tests are the durable regression.)

No migration. Out of scope: a mail-catcher (mailpit) integration test that would assert a full rendered From + link end-to-end — a small devenv addition, worth a follow-up.

Closes #16

### DIFF
--- a/internal/isms/api/api_auth.go
+++ b/internal/isms/api/api_auth.go
@@ -666,8 +666,8 @@ func (s *Server) handleInviteUser(c echo.Context) error {
 		return echo.NewHTTPError(http.StatusInternalServerError, "email not configured (SMTP_HOST)")
 	}
 
-	baseURL := os.Getenv("ISMS_BASE_URL")
-	if err := s.mailer.SendVerification(req.Email, req.Name, baseURL, token); err != nil {
+	m := s.orgMail(ctx, orgID)
+	if err := s.mailer.SendVerificationBranded(req.Email, req.Name, m.PublicURL, token, m.Branding); err != nil {
 		return echo.NewHTTPError(http.StatusInternalServerError, "sending email: "+err.Error())
 	}
 
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -363,8 +363,8 @@ func (s *Server) handleForwardReview(c echo.Context) error {
 
 		// Send email notification to reviewer
 		if s.mailer != nil && s.mailer.Enabled() {
-			baseURL := os.Getenv("ISMS_BASE_URL")
-			_ = s.mailer.SendReviewRequest(reviewer, reviewer, actor, review.DocumentID, review.Title, review.Version, baseURL, id, req.Message)
+			m := s.orgMail(ctx, orgID)
+			_ = s.mailer.SendReviewRequestBranded(reviewer, reviewer, actor, review.DocumentID, review.Title, review.Version, m.AppURL, id, req.Message, m.Branding)
 		}
 	}
 
@@ -970,8 +970,8 @@ func (s *Server) handleReviewApprove(c echo.Context) error {
 	} else {
 		// Normal flow: email the review author (human or mixed)
 		if s.mailer != nil && s.mailer.Enabled() {
-			baseURL := os.Getenv("ISMS_BASE_URL")
-			_ = s.mailer.SendReviewDecision(review.RequestedBy, review.RequestedBy, actor, review.DocumentID, review.Title, review.Version, req.Decision, baseURL)
+			m := s.orgMail(ctx, orgID)
+			_ = s.mailer.SendReviewDecisionBranded(review.RequestedBy, review.RequestedBy, actor, review.DocumentID, review.Title, review.Version, req.Decision, m.AppURL, m.Branding)
 		}
 	}
 
@@ -1959,8 +1959,8 @@ func (s *Server) handleCreateTask(c echo.Context) error {
 
 	// Email assignee if set
 	if t.Assignee != "" && s.mailer != nil && s.mailer.Enabled() {
-		baseURL := os.Getenv("ISMS_BASE_URL")
-		_ = s.mailer.SendTaskAssigned(t.Assignee, t.Assignee, t.CreatedBy, t.Title, t.Priority, baseURL)
+		m := s.orgMail(ctx, orgID)
+		_ = s.mailer.SendTaskAssignedBranded(t.Assignee, t.Assignee, t.CreatedBy, t.Title, t.Priority, m.AppURL, m.Branding)
 	}
 
 	return c.JSON(http.StatusCreated, t)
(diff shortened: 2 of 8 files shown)


unidoc-alip (changes requested):
## Summary

The core fix is solid: `resolveFrom` cleanly separates the envelope address (SPF/DKIM-critical) from the display name, and `orgURLs` correctly mirrors the SPA router's subdomain/path-based heuristic. Tests for both are well-structured and the `LEAK:` guard in `TestResolveFrom` is a good regression anchor.

Two minor issues worth cleaning up: one gofmt violation in the resubmit loop and a redundant DB query inside a per-reviewer loop. `SendOTPCode` also retains the unbranded `Send()` path, but since it is currently dead code this is low risk for this PR.

---

## Findings

### Finding 1 [should-fix] `orgMail` called N times inside a per-reviewer loop

**File:** `internal/isms/api/api_collab.go` line 3555

`s.orgMail(ctx, orgID)` is called on every iteration of the `for _, a := range allAssignments` loop (line 3546) with the same `orgID`, issuing 2 DB queries per reviewer (`GetOrganization` + `GetOrgSetting`). For a review resubmitted to 3 reviewers that is 6 round-trips to fetch the same data.

**PoC** — the loop at lines 3546–3558:

```go
for _, a := range allAssignments {
    ...
    if s.mailer != nil && s.mailer.Enabled() {
    m := s.orgMail(ctx, orgID)   // 2 DB calls on every iteration
    _ = s.mailer.SendReviewRequestBranded(a.Reviewer, ...)
    }
}
```

**Suggested fix** — hoist `orgMail` above the loop:

```go
var m orgMailCtx
if s.mailer != nil && s.mailer.Enabled() {
    m = s.orgMail(ctx, orgID)
}
for _, a := range allAssignments {
    ...
    if s.mailer != nil && s.mailer.Enabled() {
        _ = s.mailer.SendReviewRequestBranded(a.Reviewer, a.Reviewer, actor, docID, title, version, m.AppURL, existingReview.ID, req.Message, m.Branding)
    }
}
```

---

### Finding 2 [nit] gofmt violation in the resubmit mailer block

**File:** `internal/isms/api/api_collab.go` lines 3555–3556

The body of the `if s.mailer != nil && s.mailer.Enabled()` block is indented at the same level as the `if` keyword (4 tabs) instead of one level deeper (5 tabs). `gofmt -l` flags the file and `gofmt -d` shows exactly this:

```diff
            if s.mailer != nil && s.mailer.Enabled() {
-           m := s.orgMail(ctx, orgID)
-           _ = s.mailer.SendReviewRequestBranded(...)
+               m := s.orgMail(ctx, orgID)
+               _ = s.mailer.SendReviewRequestBranded(...)
            }
```

The code is functionally correct (braces define scope in Go, not indentation), but `gofmt` would reformat it. The equivalent block at lines 3595–3598 is correctly indented — looks like a copy-paste slip.

**Suggested fix:** run `gofmt -w internal/isms/api/api_collab.go`, or manually indent lines 3555–3556 to 5 tabs.

---

### Finding 3 [nit] `SendOTPCode` retains the unbranded `Send()` path

**File:** `internal/isms/mail/mail.go` line 255

`SendOTPCode` calls `m.Send()` with an empty `fromName`, which keeps the operator's `SMTP_FROM` display name verbatim — the same leak this PR fixes everywhere else. There are no callers in the codebase today (the function is dead code), but when OTP login gets wired up it will silently reintroduce the issue.

**PoC** — `mail.go:255`:

```go
func (m *Mailer) SendOTPCode(to, name, code string) error {
    ...
    return m.Send(to, "Login code", body)  // empty fromName → operator display name leaks
}
```

**Suggested fix** — add a `Branding` parameter, consistent with every other `Send*` function in this PR:

```go
func (m *Mailer) SendOTPCode(to, name, code string, b Branding) error {
    body := fmt.Sprintf(`...`, name, code)
    return m.sendAs(to, fmt.Sprintf("%s — Login code", b.name()), body, b.name())
}
```

The call site would pass `s.orgMail(ctx, orgID).Branding` just like the other branded send calls.

unidoc-ahall:

1. **`orgMail` in the resubmit loop (should-fix)** — hoisted above the `for _, a := range allAssignments` loop, guarded by the same `mailer.Enabled()` check, so it's one resolve (2 DB queries) per resubmit instead of per reviewer.
2. **gofmt (nit)** — fixed the indentation in that block and gofmt'd the new `orgurls_test.go`. Heads-up: `api_collab.go` will still show in `gofmt -l` because the repo has **pre-existing** drift (`gofmt -l ./internal ./cmd` = 58 files, and CI doesn't gate on it) — my added/changed lines are clean, and I've filed a separate ticket to run `gofmt -w` repo-wide + add a CI gate so this stops leaking into reviews. Didn't want to bundle a 58-file reformat into the mail fix.
3. **`SendOTPCode` (nit)** — now takes a `Branding` and sends via `sendAs` like the rest; the call site will pass `s.orgMail(ctx, orgID).Branding` when OTP login is wired up. No callers today, so no behaviour change now.

Build + both unit tests green.

unidoc-alip (approved):
looks good to me

Issue #83, opened 2026-06-22
Run gofmt across the repo + add a gofmt CI gate


unidoc-ahall:
# Run gofmt across the repo + add a gofmt CI gate

## Why
`gofmt -l ./internal/ ./cmd/` currently lists **58 files** that aren't gofmt-clean (mostly struct-field alignment and import grouping). CI only runs `go build` + `go test`, so formatting drift accumulates silently and then surfaces as noise in unrelated PRs — it came up in the #82 (per-org email) review, where a real indentation regression was hard to separate from the pre-existing drift.

## What
1. One mechanical commit: `gofmt -w ./...` (and `goimports -w ./...` if we want import grouping normalised too). **No logic changes** — review by confirming `git diff` is whitespace/import-order only.
2. Add a CI step that fails on unformatted code, e.g.:
   ```yaml
   - name: gofmt
     run: test -z "$(gofmt -l .)" || { gofmt -l .; echo "run gofmt -w ."; exit 1; }
   ```
   So drift can't reaccumulate.

## Acceptance criteria
- `gofmt -l .` is empty repo-wide.
- CI fails when any committed file isn't gofmt-clean.
- The reformat commit is isolated (no behavioural changes; build + full test suite green).

## Notes
- Keep the reformat as its own PR so it doesn't tangle with feature diffs.
- Not urgent — purely hygiene + preventing future review noise.

PR #84, opened 2026-06-23
fix(cli): stop dropping declared flags — audit --summary + add-command relations (#49, #52; +#19 test)

### LINKED ISSUE #19: Needs Review keeps listing documents that are already approved
## Summary

Documents that have already been approved can stay in (or reappear in) the
**Needs Review** list, with no useful action to take on them.

## Root cause

`handleNeedsReview` (`internal/isms/api/server.go:1589`) flags a document when
its current git commit ≠ the last approved review's `merge_commit` (falling back
to `commit_hash`). This false-positives even right after approval in two ways:

- **(a) frontmatter-only change** — e.g. changing the document owner bumps the
  commit hash with no content change → flagged as "Changed". *Tracked separately
  as #3.*
- **(b) review with no `merge_commit`** — falls back to `commit_hash` (the
  pre-review commit), which never equals the post-merge commit, so the document
  shows "Changed" **permanently**. `merge_commit` is recorded now
  (`api_collab.go:1133`) but older reviews have it empty — see the comment at
  `server.go:1584` ("falls back to commit_hash for older reviews").

## Proposed fix (resolves both this and #3)

Compare a **normalized content hash** (body + meaningful frontmatter, excluding
volatile fields like owner) instead of the raw git commit hash. That stops both
frontmatter-only edits *and* merge-commit drift from producing false positives.
Optionally backfill `merge_commit` for legacy approved reviews.

## Acceptance criteria

- [ ] An approved, unchanged document does not appear in Needs Review.
- [ ] Changing only the owner (or other volatile frontmatter) does not flag a
      document as needing review.
- [ ] Documents approved by the legacy path (empty `merge_commit`) no longer
      show as perpetually changed.

## Related

- #3 (owner-change false needs-review) — same root cause family; the
  content-hash fix covers both.

### LINKED ISSUE #49: CLI: audit complete drops the required --summary
## Problem

`isms audit complete` declares `--summary` as **required** but only sends the status — the summary is silently dropped. As a result `isms audit show` always displays an empty summary.

- `auditCompleteCmd` (`cmd/isms/audit.go`) calls `UpdateAuditStatus(id, "completed")` only; the parsed `summary` is never sent.
- `audit show` reads `audit.Summary`, which stays empty because completion never sets it.

## Fix direction

Send status + summary together in one `PUT /audits/:id` — `handleUpdateAudit` routes the status through `UpdateAuditStatus` and persists the summary in the same call.

## Acceptance criteria

- `audit complete --summary "..."` persists the summary; `audit show` displays it.
- Regression test asserting complete-with-summary sets both status and summary.

## Notes

The rest of the audit CLI workflow (programme → create → start → assess → finding → complete) is functional; this is the one gap.

### LINKED ISSUE #52: CLI: add commands declare relation flags that are silently dropped
## Problem

Several `isms ... add` commands declare relation flags, parse them, but never include them in the API payload — so the user thinks they linked things that were never sent. Verified in code:

- `incident add --risks/--assets/--systems` — flags declared (`cmd/isms/incident.go:89-92`) but the `db.Incident{}` struct built at :44-56 omits them.
- `legal add --documents` — declared (`cmd/isms/legal.go:81`) but never sent.
- `risk add --assets/--documents/--interested-parties` — declared (`cmd/isms/risk.go:121-123`) but never sent.

## Fix direction

Either wire the flags into the API payload so they persist, or remove them. Audit the other `add` commands for the same class of silently-ignored flag.

## Acceptance criteria

- No `add` command declares a flag it silently drops — each declared flag is either sent and persisted, or removed.
- Spot-check covered by a test where practical.

unidoc-ahall:
Two CLI bugs where a declared flag was collected and silently dropped.

## #49 — `audit complete` drops the required `--summary`
The flag was declared (and `MarkFlagRequired`) but the command only called `UpdateAuditStatus(id, "completed")` — the summary went nowhere. It now sends `summary` + `status` through the audit update endpoint, which routes the status transition through the same `UpdateAuditStatus` path.

## #52 — `add` commands drop relation flags
`incident add` (`--risks`/`--assets`/`--systems`), `risk add` (`--assets`/`--documents`) and `legal add` (`--documents`) declared relation flags, but the client posted the bare entity struct — which has no `references` field — so they were dropped. The client create methods (`AddRisk`/`CreateIncident`/`CreateLegal`) now take a `[]Reference` and send it as the API's existing `references` field (the server already turns those into bidirectional entity references). Commands build the references from the flags via a shared `buildRefs` helper (`risk`/`asset`/`system`/`document` types).

## Tests
`cmd/isms/cli_relations_test.go` runs each command against an httptest server and asserts the outbound request body:
- `audit complete` → body carries `summary` + `status`.
- `incident/risk/legal add` → body carries the right `{type,id}` references.
- `buildRefs` unit-tested (flag → references, blanks skipped).

All green (`go test ./cmd/isms`), full Go unit suite green, build clean.

## Also: #19 needs-review regression test (test-only)
`tests/test_needs_review_lifecycle.py` rides along. #19 ("Needs Review keeps listing approved docs") needed **no product change** — it was already fixed by #3 (merge_commit baseline + body comparison). Verified against the stack: approved+unchanged → not listed; body change → flagged; re-approval → cleared; never-reviewed → surfaced then cleared after review. This locks that lifecycle in. Uses unique doc ids so it's idempotent on the persistent test stack (verified by running twice + the full `just test` suite).

> Note: `gofmt -l` lists a few touched files (`main.go`, `audit.go`, `risk.go`, `client.go`) — that's the **pre-existing** repo-wide drift tracked in #83, not introduced here; my added lines are gofmt-clean.

Closes #49
Closes #52
Closes #19

### DIFF
--- a/cmd/isms/audit.go
+++ b/cmd/isms/audit.go
@@ -404,7 +404,14 @@ func auditCompleteCmd() *cobra.Command {
 			}
 
 			c := requireAPI()
-			if err := c.UpdateAuditStatus(id, "completed"); err != nil {
+			// Send the auditor summary alongside the status transition. The
+			// update endpoint routes the status change through UpdateAuditStatus
+			// and persists the summary in the same call — previously the
+			// (required) --summary flag was collected and silently dropped (#49).
+			if _, err := c.UpdateAudit(id, map[string]interface{}{
+				"status":  "completed",
+				"summary": summary,
+			}); err != nil {
 				return err
 			}
 			fmt.Printf("Audit #%d completed.\n", id)
--- a/cmd/isms/audit.go
+++ b/cmd/isms/audit.go
@@ -404,10 +404,7 @@ func auditCompleteCmd() *cobra.Command {
 			}
 
 			c := requireAPI()
-			// Send the auditor summary alongside the status transition. The
-			// update endpoint routes the status change through UpdateAuditStatus
-			// and persists the summary in the same call — previously the
-			// (required) --summary flag was collected and silently dropped (#49).
+			// summary was silently dropped before #49; send it atomically with status.
 			if _, err := c.UpdateAudit(id, map[string]interface{}{
 				"status":  "completed",
 				"summary": summary,
--- a/cmd/isms/cli_relations_test.go
+++ b/cmd/isms/cli_relations_test.go
@@ -1,0 +1,155 @@
+package main
+
+import (
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"reflect"
+	"testing"
+)
+
+// recordedReq captures one outbound CLI API call.
+type recordedReq struct {
+	method string
+	path   string
+	body   map[string]interface{}
+}
+
+// cliServer spins an httptest server that records every request body and replies
+// with a minimal JSON object the client can unmarshal. Returns the recorder.
+func cliServer(t *testing.T) (*httptest.Server, *[]recordedReq) {
+	t.Helper()
+	var got []recordedReq
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		raw, _ := io.ReadAll(r.Body)
+		var body map[string]interface{}
+		_ = json.Unmarshal(raw, &body)
+		got = append(got, recordedReq{method: r.Method, path: r.URL.Path, body: body})
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"id":1,"identifier":"X-001","title":"x"}`))
+	}))
+	t.Setenv("ISMS_API_URL", srv.URL)
+	t.Setenv("ISMS_API_KEY", "test-token")
+	return srv, &got
+}
+
+// refSet extracts the {type:[ids]} relations from a recorded create body.
+func refSet(t *testing.T, body map[string]interface{}) map[string][]string {
+	t.Helper()
+	out := map[string][]string{}
+	refs, ok := body["references"].([]interface{})
+	if !ok {
+		return out
+	}
+	for _, r := range refs {
+		m := r.(map[string]interface{})
+		typ, _ := m["type"].(string)
+		id, _ := m["id"].(string)
+		out[typ] = append(out[typ], id)
+	}
+	return out
+}
+
+func TestBuildRefs(t *testing.T) {
+	got := buildRefs(
+		refSpec{"risk", []string{"RISK-1", " RISK-2 ", ""}},
+		refSpec{"asset", []string{"A-1"}},
+		refSpec{"system", nil},
+	)
+	want := []client_Reference{
+		{"risk", "RISK-1"}, {"risk", "RISK-2"}, {"asset", "A-1"},
+	}
+	if !reflect.DeepEqual(toPlain(got), want) {
+		t.Errorf("buildRefs = %+v, want %+v", toPlain(got), want)
+	}
+}
+
+// client_Reference / toPlain decouple the assertion from the client package's
+// import path while still comparing type+id.
+type client_Reference struct{ Type, ID string }
+
+func toPlain(refs interface{}) []client_Reference {
+	v := reflect.ValueOf(refs)
+	out := make([]client_Reference, 0, v.Len())
+	for i := 0; i < v.Len(); i++ {
+		e := v.Index(i)
+		out = append(out, client_Reference{
+			Type: e.FieldByName("Type").String(),
+			ID:   e.FieldByName("ID").String(),
+		})
+	}
+	return out
+}
+
+// #49: `audit complete` must send the (required) --summary, not silently drop it.
+func TestAuditCompleteSendsSummary(t *testing.T) {
+	_, got := cliServer(t)
+	cmd := auditCmd()
+	cmd.SetArgs([]string{"complete", "5", "--summary", "Quarterly audit closed"})
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if len(*got) == 0 {
+		t.Fatal("no request sent")
+	}
+	last := (*got)[len(*got)-1]
+	if last.body["summary"] != "Quarterly audit closed" {
+		t.Errorf("summary not sent: body=%v", last.body)
+	}
+	if last.body["status"] != "completed" {
+		t.Errorf("status not sent: body=%v", last.body)
+	}
+}
+
+// #52: `incident add` must send its declared relation flags as references.
+func TestIncidentAddSendsRelations(t *testing.T) {
+	_, got := cliServer(t)
+	cmd := incidentCmd()
+	cmd.SetArgs([]string{"add", "--title", "Phish", "--description", "phishing wave",
+		"--severity", "high",
+		"--risks", "RISK-1,RISK-2", "--assets", "A-1", "--systems", "SYS-9"})
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	refs := refSet(t, (*got)[0].body)
+	for typ, want := range map[string][]string{
+		"risk": {"RISK-1", "RISK-2"}, "asset": {"A-1"}, "system": {"SYS-9"}} {
+		if !reflect.DeepEqual(refs[typ], want) {
+			t.Errorf("%s refs = %v, want %v (full=%v)", typ, refs[typ], want, refs)
+		}
+	}
+}
+
+// #52: `risk add` must send --assets and --documents as references.
+func TestRiskAddSendsRelations(t *testing.T) {
+	_, got := cliServer(t)
+	cmd := riskCmd()
+	cmd.SetArgs([]string{"add", "--title", "R", "--owner", "o@x.io",
+		"--likelihood", "3", "--impact", "4",
+		"--assets", "A-1,A-2", "--documents", "iso27001-5-1"})
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	refs := refSet(t, (*got)[0].body)
+	if !reflect.DeepEqual(refs["asset"], []string{"A-1", "A-2"}) {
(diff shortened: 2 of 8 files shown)


unidoc-alip (changes requested):
> 1 must-fix · 1 should-fix · 2 nits

The core fixes for #49 and #52 are solid: `audit complete` now sends `summary` alongside `status` atomically via `UpdateAudit` (which routes the status transition through `UpdateAuditStatus` server-side), and the `buildRefs` + anonymous-struct-embedding pattern cleanly wires relation flags for `incident add`, `risk add`, and `legal add` without touching the API contract. The #19 regression test is a clean addition.

One gap blocks fully closing #52: `risk add --interested-parties` is still declared but silently dropped — the PR's own acceptance criterion ("no `add` command declares a flag it silently drops") isn't met for that flag.

---

## Findings

| # | Severity | Location |
|---|----------|----------|
| 1 | **must-fix** | `cmd/isms/risk.go:92` |
| 2 | **should-fix** | `cmd/isms/cli_relations_test.go:87,107,126,145` |
| 3 | nit | `cmd/isms/cli_relations_test.go:72` |
| 4 | nit | `cmd/isms/audit.go:407` |

---

### 1. `risk add --interested-parties` still silently dropped — **must-fix**

**`cmd/isms/risk.go:92`**

`--interested-parties` is still silently dropped. #52's acceptance criterion requires every declared flag to be "either sent and persisted, or removed." `interestedParties` is declared at line 44 and registered at line 125 but absent from `buildRefs(...)`. The backend `riskCreateRequest` has no `interested_parties` field and no Go handler for it — the field only exists in `openapi.yaml`. The flag must be removed until the backend supports it.

**Proof of concept**

Flag is declared and parsed but never forwarded:

```go
// risk.go:44
interestedParties []string

// risk.go:125
cmd.Flags().StringSliceVar(&interestedParties, "interested-parties", nil, "Interested parties (comma-separated)")

// risk.go:92-95 — interestedParties is ABSENT from buildRefs
result, err := c.AddRisk(r, buildRefs(
    refSpec{"asset", assets},
    refSpec{"document", linkedDocs},
    // interestedParties: missing
))
```

Backend has no handler — `grep -rn "interested" internal/isms/api/ | grep -v openapi` returns nothing. Only `openapi.yaml:3266` mentions it: `interested_parties: { type: array, items: { type: string } }`. No Go struct, no DB column, no handler.

Paste this test to confirm the silent drop:

```go
func TestRiskAddInterestedPartiesDropped(t *testing.T) {
    srv, got := cliServer(t)
    defer srv.Close()
    cmd := riskCmd()
    cmd.SetArgs([]string{"add", "--title", "R", "--owner", "o@x.io",
        "--likelihood", "3", "--impact", "4",
        "--interested-parties", "legal,hr"})
    _ = cmd.Execute()
    refs, _ := (*got)[0].body["references"].([]interface{})
    for _, r := range refs {
        if r.(map[string]interface{})["type"] == "interested_party" {
            t.Fatal("should not have sent interested_party — backend does not support it")
        }
    }
    t.Log("confirmed: --interested-parties is silently swallowed")
    // test passes — proving the drop
}
```

**Suggested fix**

Remove the variable and flag registration. Re-add both once the backend adds the field.

```diff
--- a/cmd/isms/risk.go
+++ b/cmd/isms/risk.go
 	var (
 		...
-		interestedParties            []string
 		owner                        string
 		...
 	)

 	cmd.Flags().StringSliceVar(&linkedDocs, "documents", nil, "Linked document IDs (comma-separated)")
-	cmd.Flags().StringSliceVar(&interestedParties, "interested-parties", nil, "Interested parties (comma-separated)")
 	cmd.Flags().StringVar(&owner, "owner", "", "Risk owner")
```

---

### 2. `httptest.Server` never closed in test helpers — **should-fix**

**`cmd/isms/cli_relations_test.go:87` (also 107, 126, 145)**

`cliServer` returns an `*httptest.Server` that is discarded with `_` in all four test functions. The server's TCP listener stays open for the test binary's lifetime. Add `defer srv.Close()`.

**Proof of concept**

`cliServer` signature returns the server as the first value:

```go
// cli_relations_test.go:21
func cliServer(t *testing.T) (*httptest.Server, *[]recordedReq)
```

All four call sites discard it:

```go
// lines 87, 107, 126, 145
_, got := cliServer(t)
```

Each `httptest.NewServer` binds an ephemeral port. Without `Close()`, all stay bound until the process exits. Running with a high `-count` makes the accumulation measurable:

```bash
go test -count=50 -run 'TestAuditComplete|TestIncident|TestRisk|TestLegal' ./cmd/isms/ &
PID=$!; sleep 1
lsof -p $PID 2>/dev/null | grep LISTEN | wc -l  # 200 lingering listeners
wait $PID
```

With `defer srv.Close()`, the count stays at ≤1 at any moment.

**Suggested fix**

```diff
--- a/cmd/isms/cli_relations_test.go
+++ b/cmd/isms/cli_relations_test.go
 func TestAuditCompleteSendsSummary(t *testing.T) {
-	_, got := cliServer(t)
+	srv, got := cliServer(t)
+	defer srv.Close()

 func TestIncidentAddSendsRelations(t *testing.T) {
-	_, got := cliServer(t)
+	srv, got := cliServer(t)
+	defer srv.Close()

 func TestRiskAddSendsRelations(t *testing.T) {
-	_, got := cliServer(t)
+	srv, got := cliServer(t)
+	defer srv.Close()

 func TestLegalAddSendsRelations(t *testing.T) {
-	_, got := cliServer(t)
+	srv, got := cliServer(t)
+	defer srv.Close()
```

---

### 3. `toPlain` reflection wrapper is unnecessary and fragile — nit

**`cmd/isms/cli_relations_test.go:72`**

`toPlain` uses `reflect.FieldByName` to avoid importing `client.Reference`, but the comment's premise is wrong — test files in `package main` can import `isms.sh/internal/isms/client` directly (the package is already imported by `main.go`). The reflection approach is fragile: `FieldByName` on a missing field returns the zero `reflect.Value` and `.String()` returns `""` silently, with no compile-time error. Drop `toPlain`/`client_Reference` and compare `[]client.Reference` directly.

**Proof of concept**

`toPlain` at lines 72-82:

```go
Type: e.FieldByName("Type").String(),
ID:   e.FieldByName("ID").String(),
```

If `client.Reference.Type` is renamed to `Typ`, `FieldByName("Type")` silently returns the zero value, producing:

```
buildRefs = [{ RISK-1} { RISK-2} { A-1}], want [{risk RISK-1} {risk RISK-2} {asset A-1}]
```

No indication that a field rename caused it. The direct comparison would instead produce a compile error at the `want` literal — failing loudly and immediately.

**Suggested fix**

```diff
--- a/cmd/isms/cli_relations_test.go
+++ b/cmd/isms/cli_relations_test.go
 import (
 	"encoding/json"
 	"io"
 	"net/http"
 	"net/http/httptest"
 	"reflect"
 	"testing"
+
+	"isms.sh/internal/isms/client"
 )

-// client_Reference / toPlain decouple the assertion from the client package's
-// import path while still comparing type+id.
-type client_Reference struct{ Type, ID string }
-
-func toPlain(refs interface{}) []client_Reference {
-	v := reflect.ValueOf(refs)
-	out := make([]client_Reference, 0, v.Len())
-	for i := 0; i < v.Len(); i++ {
-		e := v.Index(i)
-		out = append(out, client_Reference{
-			Type: e.FieldByName("Type").String(),
-			ID:   e.FieldByName("ID").String(),
-		})
-	}
-	return out
-}

 func TestBuildRefs(t *testing.T) {
 	got := buildRefs(...)
-	want := []client_Reference{
-		{"risk", "RISK-1"}, {"risk", "RISK-2"}, {"asset", "A-1"},
-	}
-	if !reflect.DeepEqual(toPlain(got), want) {
-		t.Errorf("buildRefs = %+v, want %+v", toPlain(got), want)
+	want := []client.Reference{
+		{Type: "risk", ID: "RISK-1"}, {Type: "risk", ID: "RISK-2"}, {Type: "asset", ID: "A-1"},
+	}
+	if !reflect.DeepEqual(got, want) {
+		t.Errorf("buildRefs = %+v, want %+v", got, want)
 	}
 }
```

---

### 4. Verbose comment in `auditCompleteCmd` — nit

**`cmd/isms/audit.go:407`**

The four-line comment restates what the calling code already shows and appends historical bug context that belongs in the commit message. Trim to one line.

**Proof of concept**

```go
// Send the auditor summary alongside the status transition. The
// update endpoint routes the status change through UpdateAuditStatus
// and persists the summary in the same call — previously the
// (required) --summary flag was collected and silently dropped (#49).
if _, err := c.UpdateAudit(id, map[string]interface{}{
    "status":  "completed",
    "summary": summary,
}); err != nil {
```

Lines 1-2 restate what `UpdateAudit` with `status`+`summary` keys already shows. Lines 3-4 add bug history already captured in commit `6ea6ded`.

**Suggested fix**

```diff
--- a/cmd/isms/audit.go
+++ b/cmd/isms/audit.go
-			// Send the auditor summary alongside the status transition. The
-			// update endpoint routes the status change through UpdateAuditStatus
-			// and persists the summary in the same call — previously the
-			// (required) --summary flag was collected and silently dropped (#49).
+			// summary was silently dropped before #49; send it atomically with status.
 			if _, err := c.UpdateAudit(id, map[string]interface{}{
```

unidoc-ahall:

1. **`--interested-parties` (must-fix)** — removed the flag and its variable. You're right: no backend field/handler exists (only `openapi.yaml`), so it was a guaranteed silent drop. Re-add when the backend gains the field. #52's invariant (no declared flag silently dropped) now holds.
2. **Leaking test servers (should-fix)** — `srv, got := cliServer(t)` + `defer srv.Close()` in all four tests.
3. **`toPlain` reflection (nit)** — dropped it; the test imports `client` and compares `[]client.Reference` directly, so a field rename now fails at compile time instead of silently.
4. **Verbose comment (nit)** — trimmed to one line.

Build + all five CLI tests green. (`risk.go`/`audit.go` still show in `gofmt -l` — pre-existing repo drift tracked in #83, not from these changes; my changed lines are clean.)

unidoc-alip (approved):
looks good to me

Issue #27, opened 2026-06-07
Disabling OTP requires no re-authentication


unidoc-ahall:
## Summary
Disabling TOTP requires no current TOTP code, no password, and no confirmation — a single click on a valid session removes 2FA.

## Rationale
A hijacked or stolen session can silently downgrade account security. Standard practice requires step-up re-authentication before disabling 2FA. This must be enforced server-side.

## Evidence
- `Settings.vue:70-72` binds a single button to `disableOTP()` (383-397) -> `DELETE /api/v1/auth/otp`, no confirmation/code/password.
- `handleOTPDisable` (`api_auth.go:585-598`) reads session email, calls `db.ClearOTP()`, returns 200; no body parsed, no step-up middleware on the route.

## Acceptance criteria
- Disabling OTP requires a valid current TOTP code or password, verified server-side.
- The UI shows a confirmation dialog and a code/password prompt.
- Test covers rejection without valid re-auth and success with it.

Issue #24, opened 2026-06-07
Contributor can change incident/change-request status via the general edit endpoint


unidoc-ahall:
## Summary
The general edit endpoints for incidents and change requests allow the `contributor` role and route a `status` field through the status-transition functions with no secondary manager/admin check, bypassing the RBAC enforced on the dedicated status endpoints.

## Rationale
The rule 'only managers/admins may approve/reject an incident or change request' is enforced on the dedicated endpoint and (for changes) the status-button UI, but a contributor can set the status via the general edit form / a direct `PUT` call. This is an authorization bypass, not just an inconsistency.

## Evidence
- `handleUpdateIncident` (`internal/isms/api/api_incidents.go:262`) and `handleUpdateChange` (`internal/isms/api/api_collab.go:2351`) both `requireRole(c, 'admin','manager','contributor')` and route `status` through `UpdateIncidentStatus`/`UpdateChangeRequestStatus` (change status sub-path `api_collab.go:2396-2402`) with no extra guard.
- Dedicated endpoints `handleUpdateIncidentStatus` (`api_incidents.go:446`) and `handleUpdateChangeStatus` (`api_collab.go:2466`) restrict to admin/manager.
- `Changes.vue:219` exposes a status `<select>` in the contributor-reachable edit form; incidents UI blocks it but the API does not.

## Acceptance criteria
- A contributor calling the general edit endpoint with a `status` change is rejected with 403 (or `status` is ignored and must go through the dedicated endpoint).
- Behavior is identical between the general-edit and dedicated-status paths for all roles.
- `Changes.vue` status `<select>` is hidden/disabled for contributors.
- Tests cover contributor attempts via both endpoints for incidents and change requests.

_Found by the codebase gap analysis (dimension B — consistency)._

PR #85, opened 2026-06-23
fix(security): status changes manager/admin-only + OTP disable needs re-auth (#24, #27)

unidoc-ahall:
Two security fixes, conflict-free with the open CLI PR.

## #24 — RBAC bypass on status via the general edit endpoint
`handleUpdateIncident` and `handleUpdateChange` route a status change through the dedicated status-transition functions, but only required the **contributor** role — while `PUT /incidents/:id/status` and `PUT /changes/:id/status` require **admin/manager**. So a contributor could change status through the general edit endpoint. The general handlers now require admin/manager **when the status actually changes**; non-status edits are unaffected.

## #27 — disabling OTP required no re-authentication
`handleOTPDisable` cleared the OTP secret straight from the session — a hijacked session could silently remove 2FA. It now requires a **valid current OTP code**. Frontend: the Settings disable action prompts for the code; `deleteJSON` gained an optional body.

## Tests — `tests/test_status_rbac_otp.py`
- Contributor blocked (403) from status changes on both `PUT /incidents/:id` and `PUT /changes/:id`; non-status edits still 200; manager status changes still 200.
- OTP disable rejected without a code (400) and with a wrong code (401), accepted with a valid code (200). Uses a unique throwaway user per run + RFC 6238 TOTP, so it's idempotent on the persistent stack.

Verified against the live stack with `just test` (run twice for idempotency).

### DIFF
--- a/internal/isms/api/api_auth.go
+++ b/internal/isms/api/api_auth.go
@@ -583,11 +583,28 @@ func (s *Server) handleOTPVerify(c echo.Context) error {
 // --- Self-service: OTP disable ---
 
 func (s *Server) handleOTPDisable(c echo.Context) error {
+	var req otpVerifyRequest
+	if err := c.Bind(&req); err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "invalid request")
+	}
+
 	ctx := c.Request().Context()
 	email := getUserEmail(c)
 	user, err := s.db.GetUserByEmail(ctx, email)
 	if err != nil {
 		return echo.NewHTTPError(http.StatusUnauthorized, "user not found")
+	}
+
+	// Re-authenticate with the current OTP code before removing the second
+	// factor — otherwise a hijacked session could silently disable 2FA (#27).
+	if user.OTPSecret == nil || *user.OTPSecret == "" {
+		return echo.NewHTTPError(http.StatusBadRequest, "OTP is not enabled")
+	}
+	if req.Code == "" {
+		return echo.NewHTTPError(http.StatusBadRequest, "current OTP code is required to disable OTP")
+	}
+	if !verifyTOTP(*user.OTPSecret, req.Code) {
+		return echo.NewHTTPError(http.StatusUnauthorized, "invalid OTP code")
 	}
 
 	if err := s.db.ClearOTP(ctx, user.ID); err != nil {
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -2396,6 +2396,11 @@ func (s *Server) handleUpdateChange(c echo.Context) error {
 	// Route status changes through the dedicated transition function so that
 	// approved_at / implemented_at are cleared correctly on reverse transitions.
 	if req.Status != nil && *req.Status != old.Status {
+		// Changing status is a manager/admin action — match the dedicated status
+		// endpoint so the general edit endpoint can't be used as an RBAC bypass (#24).
+		if err := requireRole(c, "admin", "manager"); err != nil {
+			return err
+		}
 		if err := s.db.UpdateChangeRequestStatus(ctx, orgID, id, *req.Status, getUserEmail(c)); err != nil {
 			return pgxHTTPError(err)
 		}
--- a/internal/isms/api/api_incidents.go
+++ b/internal/isms/api/api_incidents.go
@@ -324,6 +324,12 @@ func (s *Server) handleUpdateIncident(c echo.Context) error {
 	// Route status changes through the dedicated transition function so that
 	// contained_at / resolved_at / closed_at are cleared correctly on reverse transitions.
 	if req.Status != nil && *req.Status != existing.Status {
+		// Changing status is a manager/admin action. The dedicated status
+		// endpoint already enforces this; the general edit endpoint must apply
+		// the same rule so it can't be used as an RBAC bypass (#24).
+		if err := requireRole(c, "admin", "manager"); err != nil {
+			return err
+		}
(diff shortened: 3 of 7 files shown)


unidoc-alip (changes requested):
Both API-layer security fixes are correct and well-targeted. The RBAC guard in `handleUpdateIncident` and `handleUpdateChange` is placed exactly inside the `status != old.status` branch, so non-status edits by contributors are unaffected. The OTP re-auth is sound: `verifyTOTP` accepts a ±1 window, the frontend sends `Content-Type: application/json` via `getHeaders()` so Echo's `Bind` parses the DELETE body correctly, and the test matrix covers no-code / wrong-code / valid-code paths.

One acceptance criterion from #24 is missing: the status `<select>` in `Changes.vue` is still visible and interactive for contributors even though the API now rejects their request with 403. They'll get a confusing silent error with no indication the control is off-limits.

---

## Findings

| [1](#finding-1) | [should-fix] | `web/src/views/Changes.vue` | 219 |
| [2](#finding-2) | [nit] | `web/src/views/Settings.vue` | 72 |
| [3](#finding-3) | [nit] | `tests/test_status_rbac_otp.py` | 55 |

---

### Finding 1 · [should-fix] · `web/src/views/Changes.vue:219`

**Status `<select>` is not hidden for contributors**

The status `<select>` has no role guard, so contributors see and can interact with it. When they submit a status change the API now correctly returns 403, but the UI shows a generic error with no explanation.

**Proof of concept:**

`Incidents.vue:219` gates the Edit button with `v-if="canWrite && !editingSection"` where `canWrite` is `admin || manager` (line 574). Contributors never enter the edit form, so they never see the status select.

`Changes.vue:182` gates the Edit button with `v-if="canCreate && !editingSection"` where `canCreate` includes contributors (line 464). Contributors click Edit, the full form renders, and the status `<select>` at line 219 has no guard at all.

Reproduction: log in as a contributor → open any change request → click Edit in the Overview tab → status dropdown is fully interactive → select a new status and save → API returns 403 → UI shows a generic network error.

**Fix:**

Add `canWrite` to `Changes.vue` matching the pattern in `Incidents.vue`:

```js
// script section, alongside the existing canCreate
const canWrite = computed(() => userRole.value === 'admin' || userRole.value === 'manager')
```

Wrap the status block (lines 217–227) with it:

```vue
<div v-if="canWrite">
  <label class="block text-xs font-medium text-slate-500 mb-1">Status</label>
  <select v-model="editForm.status"
    class="w-full bg-slate-800 border border-slate-700 rounded-lg px-3 py-2 text-sm text-slate-200 focus:outline-none focus:ring-1 focus:ring-blue-500">
    <option value="proposed">Proposed</option>
    <option value="approved">Approved</option>
    <option value="rejected">Rejected</option>
    <option value="in_progress">In Progress</option>
    <option value="implemented">Implemented</option>
    <option value="closed">Closed</option>
  </select>
</div>
```

---

### Finding 2 · [nit] · `web/src/views/Settings.vue:72`

**OTP disable input accepts non-digit characters**

`type="text"` with `inputmode="numeric"` is only a keyboard hint on mobile; on desktop any characters can be typed. The button disable guard only checks `disableCode.length !== 6`, so typing `abcdef` enables the button and sends a guaranteed-wrong code to the server.

**Proof of concept:**

```
type "abcdef" into the input
→ disableCode.length === 6 → button enables
→ deleteJSON('/api/v1/auth/otp', { code: 'abcdef' })
→ verifyTOTP returns false → 401 "invalid OTP code"
```

There's no user-visible indication that the field must be digits.

**Fix:**

Add an `@input` handler to strip non-digits in place:

```vue
<input v-model="disableCode" type="text" maxlength="6" inputmode="numeric"
  placeholder="000000"
  @input="disableCode = disableCode.replace(/\D/g, '').slice(0, 6)"
  class="w-28 bg-slate-800 border border-slate-700 rounded-lg px-3 py-2 text-sm text-white tracking-widest text-center focus:outline-none focus:border-red-500" />
```

---

### Finding 3 · [nit] · `tests/test_status_rbac_otp.py:55`

**Hand-rolled RFC 6238 TOTP**

`_totp()` re-implements TOTP inline (HMAC-SHA1, 30 s window, 6 digits). The implementation is correct and matches the server's `verifyTOTP` logic, so this is informational only.

**Proof of concept:**

`tests/requirements.txt` lists only `pytest>=9.0`, `requests>=2.31`, and `playwright>=1.49` — no `pyotp`. The hand-rolled version is the only option without adding a new dependency. The server's `verifyTOTP` accepts ±1 window, so edge-of-window timing races in the test are covered server-side.

**Fix:**

No change required now. If `pyotp` is ever added to the test dependencies, replace `_totp()` with:

```python
import pyotp
pyotp.TOTP(secret_b32).now()
```

unidoc-ahall:
Thanks @unidoc-alip — addressed:

1. **Changes.vue status select (should-fix)** — added `canWrite` (`admin`/`manager`) and gated the status `<select>` with it, matching `Incidents.vue`. Contributors keep their non-status edits (consistent with what #24 allows at the API). The broader "contributor shouldn't edit changes at all" is the #23 role realignment, tracked separately.
2. **OTP disable input (nit)** — added `@input` that strips non-digits (`replace(/\\D/g, '').slice(0, 6)`), so a wrong non-numeric code can't be sent.
3. **Hand-rolled TOTP (nit)** — agreed, leaving as-is; will swap to `pyotp.TOTP(secret).now()` if/when pyotp is added to the test deps.

Web build green.

unidoc-alip (approved):
looks good to me

PR #87, opened 2026-06-23
fix(suggestions): default task assignee to the applier on apply (#86)

### LINKED ISSUE #86: Applying a task suggestion with no assignee fails with a raw SQL error
# Applying a task suggestion with no assignee fails with a raw SQL error

## What happens
A manager opens a suggestion that proposes a **task**, clicks apply/approve, and gets:

```
apply failed: apply: ERROR: null value in column "assignee_id" of relation "tasks" violates not-null constraint (SQLSTATE 23502)
```

The apply fails and the raw database error is shown to the user.

## Root cause
`tasks.assignee_id` is `NOT NULL`, but a suggested task usually carries no assignee. `applyTaskCreate` inserted the task as-is, so an empty assignee resolved to a NULL `assignee_id` and tripped the constraint.

## Fix
Default the assignee to the **manager applying the suggestion** when the suggestion has none. They can reassign it afterwards in edit. (No schema change — tasks still require an assignee by design.)

## Acceptance
- Applying a task-create suggestion with no assignee succeeds (no SQL error) and the task is assigned to the applier.
- Regression test covering apply-without-assignee.

Related: surfaced alongside the suggestion UX (suggestions should read as "under review") and the role realignment (#23) — tracked separately.

unidoc-ahall:
## The bug
Applying a suggestion that creates a **task** failed with a raw SQL error when the suggestion had no assignee:
```
null value in column "assignee_id" of relation "tasks" violates not-null constraint
```
`tasks.assignee_id` is NOT NULL; `applyTaskCreate` inserted the suggested task with an empty assignee.

## Fix
Default the assignee to the **manager applying** the suggestion when none is given — they can reassign in edit. No schema change (tasks still require an assignee by design).

## Test
`tests/test_suggestion_apply_task.py` — create a task suggestion with no assignee, apply it, assert 200 (no SQL error) and that the task is assigned to the applier. Unique title per run, idempotent on the persistent stack.

Closes #86

### DIFF
--- a/internal/isms/api/api_suggestions.go
+++ b/internal/isms/api/api_suggestions.go
@@ -1136,9 +1136,7 @@ func applyTaskCreate(ctx context.Context, tx pgx.Tx, s *Server, orgID int, sg *d
 	}
 	if t.Priority == "" { t.Priority = "medium" }
 	if t.TaskType == "" { t.TaskType = "general" }
-	// Tasks require an assignee (tasks.assignee_id is NOT NULL). A suggested task
-	// often has none, which previously surfaced as a raw SQL constraint error on
-	// apply. Default to the manager applying it — they can reassign in edit.
+	// tasks.assignee_id is NOT NULL; default to the applier when the suggestion carries none.
 	if t.Assignee == "" {
 		t.Assignee = actor
 	}
--- a/internal/isms/api/api_suggestions.go
+++ b/internal/isms/api/api_suggestions.go
@@ -1136,6 +1136,12 @@ func applyTaskCreate(ctx context.Context, tx pgx.Tx, s *Server, orgID int, sg *d
 	}
 	if t.Priority == "" { t.Priority = "medium" }
 	if t.TaskType == "" { t.TaskType = "general" }
+	// Tasks require an assignee (tasks.assignee_id is NOT NULL). A suggested task
+	// often has none, which previously surfaced as a raw SQL constraint error on
+	// apply. Default to the manager applying it — they can reassign in edit.
+	if t.Assignee == "" {
+		t.Assignee = actor
+	}
 	if err := db.CreateTaskTx(ctx, tx, orgID, &t); err != nil {
 		return "", err
 	}
--- a/tests/test_suggestion_apply_task.py
+++ b/tests/test_suggestion_apply_task.py
@@ -1,0 +1,48 @@
+"""Regression: applying a task-create suggestion with no assignee must not blow
+up with a raw SQL constraint error.
+
+tasks.assignee_id is NOT NULL, but a suggested task often carries no assignee.
+applyTaskCreate used to insert it as-is, surfacing
+  "null value in column assignee_id ... violates not-null constraint"
+to the manager. It now defaults the assignee to the applier, who can reassign
+in edit afterwards.
+"""
+import uuid
+
+import requests
+from conftest import ADMIN_EMAIL
+
+
+def test_apply_task_suggestion_without_assignee_defaults_to_applier(api_url, admin_headers):
+    title = f"apply-task-noassignee {uuid.uuid4().hex[:8]}"
+
+    # A task suggestion with NO assignee — the exact repro.
+    r = requests.post(f"{api_url}/suggestions", headers=admin_headers, json={
+        "entity_type": "task",
+        "suggestion_type": "create",
+        "title": title,
+        "rationale": "follow-up proposed during review",
+        "payload": {
+            "title": title,
+            "description": "Investigate the reported issue",
+            "priority": "high",
+            "task_type": "general",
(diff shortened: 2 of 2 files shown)


unidoc-alip (changes requested):
Tight, correct fix. The six-line change in `applyTaskCreate` is the right place to default the assignee, consistent with how `applyIncidentCreate` handles the same situation (it unconditionally sets `Assignee: actor`). The transaction safety, changelog, and notification paths are all unaffected. One test-quality note worth considering before merge: the regression test fetches the task via a full `GET /tasks?limit=2000` scan — using `?q=<title>&limit=5` is more targeted (the list endpoint supports ILIKE search). Comment verbosity in Go is a nit.

---

## Findings

| [1](#finding-1) | [should-fix] | `tests/test_suggestion_apply_task.py` | 44 |
| [2](#finding-2) | [nit] | `internal/isms/api/api_suggestions.go` | 1139 |

---

### Finding 1 · [should-fix] · `tests/test_suggestion_apply_task.py:44`

**Task lookup uses `limit=2000` full scan; use the `?q=` search filter instead.**

#### Verification

`server.go:586` registers `GET /tasks/:id` → `handleGetTask` (`api_collab.go:2011`), which calls `strconv.Atoi(c.Param("id"))` and rejects non-integers. The apply response returns `applied_entity_id` as the string identifier (e.g. `"TASK-3"`), not the numeric DB id — so `GET /tasks/TASK-3` would return 400. There is no `GET /tasks/by-identifier/:ident` route.

`handleListTasks` (`api_collab.go:1852`) accepts `?q=` which maps to `PaginatedTasks` (`db/tasks.go:279-281`):

```go
where += fmt.Sprintf(` AND (t.title ILIKE $%d OR COALESCE(t.description,'') ILIKE $%d)`, idx, idx)
args = append(args, "%"+p.Search+"%")
```

Since the title contains a `uuid4().hex[:8]` suffix, `?q=<title>` matches exactly one row.

#### Proof of concept

```python
# Current (fragile): if the org accumulates >2000 tasks, the target task may fall
# outside the page and the assertion fails with "task not found" — not the actual bug.
r = requests.get(f"{api_url}/tasks", headers=admin_headers, params={"limit": 2000})
rows = r.json().get("data", r.json())
match = [t for t in rows if t.get("title") == title]
```

#### Suggested fix

```python
# Use the search filter — fetches at most 5 rows matching the unique title.
r = requests.get(f"{api_url}/tasks", headers=admin_headers, params={"q": title, "limit": 5})
rows = r.json().get("data", r.json())
match = [t for t in rows if t.get("title") == title]
assert match, "task created by the applied suggestion not found"
assert match[0]["assignee"] == ADMIN_EMAIL, \
    f"expected assignee to default to the applier {ADMIN_EMAIL}, got {match[0].get('assignee')!r}"
```

> Note: the original plan suggested `GET /tasks/{applied_entity_id}` — that was incorrect; `handleGetTask` requires an integer id. This corrects it.

---

### Finding 2 · [nit] · `internal/isms/api/api_suggestions.go:1139`

**Collapse the three-line comment to a single line.**

#### Verification

Lines 1139–1141 read:

```go
// Tasks require an assignee (tasks.assignee_id is NOT NULL). A suggested task
// often has none, which previously surfaced as a raw SQL constraint error on
// apply. Default to the manager applying it — they can reassign in edit.
```

The WHY (NOT NULL constraint, previous SQL error) is non-obvious without checking the migration, so a comment is warranted. The third sentence ("they can reassign in edit") adds no information not already obvious from reading the fix.

#### Proof of concept

```
Current (3 lines, 195 chars):
  // Tasks require an assignee (tasks.assignee_id is NOT NULL). A suggested task
  // often has none, which previously surfaced as a raw SQL constraint error on
  // apply. Default to the manager applying it — they can reassign in edit.

Tightened (1 line, 73 chars — same information density):
  // tasks.assignee_id is NOT NULL; default to the applier when the suggestion carries none.
```

#### Suggested fix

```go
// tasks.assignee_id is NOT NULL; default to the applier when the suggestion carries none.
if t.Assignee == "" {
    t.Assignee = actor
}
```

unidoc-ahall:
Thanks @unidoc-alip — both addressed:

1. **Test task lookup (should-fix)** — switched from `GET /tasks?limit=2000` to `?q=<title>&limit=5`, matching on the uuid-suffixed unique title via the list endpoint's ILIKE filter. (And yes — you're right that `GET /tasks/{applied_entity_id}` wouldn't work; `handleGetTask` wants the integer id, not the `TASK-N` identifier.)
2. **Comment (nit)** — collapsed to one line: `// tasks.assignee_id is NOT NULL; default to the applier when the suggestion carries none.`

Build green.

unidoc-alip (approved):
looks good to me

PR #89, opened 2026-06-23
fix(rbac): enforce contributor=suggest-only, reader=read-only (#23)

unidoc-ahall:
Enforces the role model that README (#23) already documents but the code never enforced.

## Model
- **Contributor** — proposes/reports **through suggestions**, comments; no direct create/edit/apply (same path as an AI agent).
- **Reader** — read-only; cannot suggest.
- **Ownership exception** — you may act on what's yours: a task *assignee* may advance their task's status; a document *owner* may confirm its periodic review.

## Backend
- `handleCreateIncident`/`handleUpdateIncident`, `handleCreateTask`, `handleCreateChange`/`handleUpdateChange` — contributor dropped, admin/manager only.
- `handleUpdateTaskStatus` — keeps contributor **only for a task assigned to them**; managers/admins may change any task.
- `handleConfirmDocumentReview` — already gates contributor-confirm on document ownership; unchanged.

## Frontend
- `SuggestNewButton` hidden for readers (via `useSession` role) — fixes the original "reader sees Suggest, gets 403" report.

## Tests
- `tests/test_contributor_role.py` rewritten: contributor cannot create incident/change/task directly (403) but **can** create suggestions; reader **cannot** suggest (403). Existing register/admin/delete forbids retained.
- `tests/test_status_rbac_otp.py`: the #85-era "contributor can still edit non-status incident fields" test is superseded — under #23 a contributor cannot edit incidents directly at all (403).

## Docs
`auditor-guide.md` + `suggestions.md` aligned with the model.

Verified with `just test` against the restarted stack.

Out of scope (follow-up): the richer "suggestions read as under review" UX.

Closes #23

PR #90, opened 2026-06-23
fix(cli): review approve uses the approval handler (not a dead status path) + document completion (#51, #54)

### LINKED ISSUE #51: CLI: review approve hits a dead endpoint and bypasses the approval flow
## Problem

`isms review approve` is broken: it POSTs to `/v1/approvals`, a route that is **not registered**, then force-sets the review status to `approved` directly — bypassing the entire approval flow.

Verified in code:
- `reviewApproveCmd` calls `c.AddApproval()` → POST `/v1/approvals` (`internal/isms/client/client.go:509-512`). No such route exists in `server.go` (only `POST /reviews/:id/approve` at :480 and `GET /documents/:docId/approvals` at :499).
- It then calls `UpdateReviewStatus(id, "approved")` (`cmd/isms/review.go:140-157`), skipping `handleReviewApprove` (`POST /reviews/:id/approve`) which enforces the self-approval block, assignment check, duplicate-action guard, `decision_log` write, and the approval-policy flow (`require_human`, `auto_merge`).

So a CLI approval is recorded incorrectly (or 404s) and the governance trail / policy enforcement is skipped entirely.

## Fix direction

`review approve` should call the real `POST /v1/reviews/:id/approve` endpoint, so it goes through the same logic as the web UI. Remove the dead `AddApproval`/`/v1/approvals` path.

## Acceptance criteria

- `review approve` calls `POST /v1/reviews/:id/approve`.
- The decision is written to `decision_log` and the approval-policy flow runs.
- Self-approval / assignment / duplicate-action guards apply the same as the web UI.
- Regression test covering CLI approval through the real endpoint.

### LINKED ISSUE #54: Surface & document shell completion (already works via cobra)
## Summary

`isms completion bash|zsh|fish|powershell` already works — cobra (v1.10.2) provides it and `DisableDefaultCmd` is not set anywhere in `cmd/isms`. It's just undiscovered/undocumented. Surface it; consider dynamic completion later.

## Scope

- Document how to enable shell completion in the README / CLI docs (a one-liner per shell). cobra's own `completion <shell> --help` already prints install steps.
- (Optional, later) dynamic completion for entity IDs / enums via `ValidArgsFunction` calling the API.

## Acceptance criteria

- Completion setup is documented and discoverable for at least bash + zsh.

## Notes

Near-zero effort for the static part — it already exists. The "command that writes bashrc" idea is just wrapping/documenting the existing `isms completion bash`.

unidoc-ahall:
Two CLI items, conflict-free with the open #89.

## #51 — `review approve` bypassed the approval flow
The command called `AddApproval` (/approvals) then `UpdateReviewStatus(id, "approved")` → `PUT /reviews/:id/status`. But that endpoint **only accepts `closed`** (all other transitions go through dedicated handlers), so approve hit a dead path and side-stepped the real governance (decision log, content hash, atomic status transition in `handleReviewApprove`). It now calls the dedicated **POST /reviews/:id/approve** via a new `client.ApproveReview`, the same path the web UI uses.

## #54 — shell completion
Cobra already provides `isms completion <shell>`; added a README **Shell completion** section with load/install one-liners for bash/zsh.

## Tests
- `cmd/isms/cli_relations_test.go::TestReviewApproveUsesApprovalEndpoint` — `review approve 7` makes exactly **one** request, `POST /v1/reviews/7/approve` with `decision=approved` (no `PUT /status`, no `/approvals`).
- Verified `isms completion bash` emits a completion script (exit 0).
- `go test ./cmd/isms` green; build clean.

## Also: flaky-test fix (found running the suite)
`test_incident_reopen_clears_closure_metadata` was intermittently failing with "cannot resolve incident: 1 open corrective action(s) still linked". Root cause: `TestCrossEntityLinkage` created a CA→incident reference using the incident's numeric PK (`INC-<id>`) instead of its identifier. Identifiers are a monotonic sequence, so `INC-<pk>` is a **dangling** reference; once the sequence climbs to that value a later fresh incident inherits the dangling open-CA link and can't resolve — flaky and CI-seed-dependent. Fixed to reference the incident's real identifier. (Bundled here since it's a one-line test fix found during this PR's suite run.)

> `gofmt -l` may list `review.go`/`client.go` — pre-existing repo drift (#83), not from these changes.

Closes #51
Closes #54

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,23 @@ isms document list
 isms risk list
 ```
 
+### Shell completion
+
+The CLI ships completion for bash, zsh, fish, and PowerShell:
+
+```bash
+isms completion bash        # print the script
+source <(isms completion bash)   # load it for the current shell
+
+# install permanently (bash, Linux):
+isms completion bash | sudo tee /etc/bash_completion.d/isms >/dev/null
+
+# zsh:
+isms completion zsh > "${fpath[1]}/_isms"
+```
+
+Run `isms completion --help` for per-shell install instructions.
+
 ### Setup AI Agent
 
 ```bash
--- a/cmd/isms/cli_relations_test.go
+++ b/cmd/isms/cli_relations_test.go
@@ -164,4 +164,7 @@ func TestReviewApproveUsesApprovalEndpoint(t *testing.T) {
 	if r.body["decision"] != "approved" {
 		t.Errorf("decision not sent as 'approved': body=%v", r.body)
 	}
+	if r.body["comment"] != "LGTM" {
+		t.Errorf("comment not forwarded: body=%v", r.body)
+	}
 }
--- a/cmd/isms/cli_relations_test.go
+++ b/cmd/isms/cli_relations_test.go
@@ -142,3 +142,26 @@ func TestLegalAddSendsRelations(t *testing.T) {
 		t.Errorf("document refs = %v (full=%v)", refs["document"], refs)
 	}
 }
+
+// #51: `review approve` must go through the dedicated approval handler
+// (POST /reviews/:id/approve), not PUT /reviews/:id/status (which the server
+// rejects for anything but "closed") — and must not side-step it via /approvals.
+func TestReviewApproveUsesApprovalEndpoint(t *testing.T) {
+	srv, got := cliServer(t)
+	defer srv.Close()
+	cmd := reviewCmd()
+	cmd.SetArgs([]string{"approve", "7", "--comment", "LGTM"})
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	if len(*got) != 1 {
+		t.Fatalf("expected exactly one request, got %d: %+v", len(*got), *got)
+	}
+	r := (*got)[0]
+	if r.method != "POST" || r.path != "/v1/reviews/7/approve" {
+		t.Errorf("approve hit %s %s, want POST /v1/reviews/7/approve", r.method, r.path)
+	}
+	if r.body["decision"] != "approved" {
+		t.Errorf("decision not sent as 'approved': body=%v", r.body)
+	}
+}
--- a/cmd/isms/review.go
+++ b/cmd/isms/review.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 
 	"isms.sh/internal/isms/client"
-	"isms.sh/internal/isms/db"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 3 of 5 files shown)


unidoc-alip (changes requested):
Both fixes are correct and well-scoped. The approval routing fix is exactly right — `ApproveReview` goes through the real governance path, the test proves exactly one request reaches `POST /reviews/:id/approve`, and the flaky-test fix for `TestCrossEntityLinkage` closes a subtle identifier-vs-PK mismatch. One leftover and two minor test/style gaps below.

---

## Findings

### Finding 1 [should-fix] — `internal/isms/client/client.go:521` — `AddApproval` is dead code that posts to a non-existent route

`client.AddApproval` has zero callers after this PR replaces its only call site with `ApproveReview`. It also POSTs to `/v1/approvals`, which is not registered in `server.go` (the only "approvals" route is `GET /documents/:docId/approvals`). Every call would 404.

**Proof:**
```
$ grep -rn "\.AddApproval\b" . --include="*.go"
# (no output)

$ grep -n "approvals" internal/isms/api/server.go
577:    api.GET("/documents/:docId/approvals", s.handleListApprovals)
```

**Fix:** Delete `client.go:519–524`:
```go
// --- Approvals ---

func (c *Client) AddApproval(approval *db.Approval) error {
    _, err := c.post("/v1/approvals", approval)
    return err
}
```

---

### Finding 2 [nit] — `cmd/isms/review.go:139` — comment narrates removed code, not current code

Three of the five comment lines describe what the old code did wrong. That context is already in the PR description and commit message. Only the hidden constraint (status endpoint only accepts `"closed"`) is worth keeping inline.

**Current (lines 139–143):**
```go
// Go through the dedicated approval handler so the decision log,
// content hash and status transition happen atomically. The old code
// hit PUT /reviews/:id/status with "approved", which the server
// rejects (status endpoint only allows "closed"), and side-stepped the
// approval governance (#51).
```

**Suggested:**
```go
// status endpoint only accepts "closed"; all other transitions go through dedicated handlers.
```

---

### Finding 3 [nit] — `cmd/isms/cli_relations_test.go:164` — test doesn't assert the comment reaches the server

The test sets `--comment LGTM` but only checks `r.body["decision"]`. The comment flows through `c.ApproveReview(id, "approved", comment)` into the POST body — if that mapping broke, this test wouldn't catch it. `cliServer` already captures the full body, so the fix is one line.

**Fix:** Add after the existing `decision` assertion:
```go
if r.body["comment"] != "LGTM" {
    t.Errorf("comment not forwarded: body=%v", r.body)
}
```

unidoc-ahall:
Thanks @unidoc-alip — all three:

1. **Dead AddApproval (should-fix)** — deleted. You're right: zero callers after #51, and `/v1/approvals` isn't a registered route so it would've 404'd. (`db.AddApprovalTx`/`db.AddApproval` are the real persistence path and are untouched.)
2. **Comment (nit)** — trimmed to the one durable constraint: `// The status endpoint only accepts "closed"; other transitions go through dedicated handlers`.
3. **Comment assertion (nit)** — the test now also asserts `body["comment"] == "LGTM"`, so a break in the comment mapping would be caught.

`go test ./cmd/isms` green; build clean.

unidoc-alip (approved):
looks good to me

PR #91, opened 2026-06-24
fix: cron backfill writes changelog; evidence download works on local blob backend (#50, #31)

### LINKED ISSUE #31: Evidence download silently fails on the local (file) blob backend
## Summary

Downloading check-in evidence silently fails when the blob backend is **local
(file)** — the default for self-hosted deployments. It works on the S3 backend.

## Root cause

`handleDownloadEvidence` (`internal/isms/api/api_objectives.go:878`) returns two
different response shapes:

- **S3 backend:** JSON `{url, title, content_type}` (a presigned URL).
- **Local backend:** the raw file bytes served directly (`LocalStore.URL`
  returns an error, so the handler falls through to serve bytes).

The frontend `downloadEvidence` (`web/src/api.js:439`) always uses `fetchJSON`,
so on the local backend it tries to parse raw file bytes as JSON and fails
silently. On S3 the response is JSON, so it works there.

## Proposed fix

Client handles both shapes: if the response is JSON with a `url`, open it;
otherwise treat the response as a file blob and download it. (Or make the server
contract consistent across backends.) Test both backends.

## Acceptance criteria

- [ ] Evidence downloads on the local/file backend — no silent failure.
- [ ] Still works on the S3 backend (presigned URL).
- [ ] Covered by a test against the file backend.

No schema change.

### LINKED ISSUE #50: Cron manager backfill skips changelog entries

unidoc-ahall:
Two data/storage bugs, conflict-free with the open #89/#90.

## #50 — cron backfill skipped the changelog
The manager backfills missing `next_review` on suppliers and systems, but did so via `UpdateSupplier`/`UpdateSystem` with **no changelog entry** — automated changes never reached the audit trail. Now logs them with `DiffFields` + `LogChanges` (actor `system`), the same pattern every API update handler uses.

## #31 — evidence download broken on the local (file) backend
`handleDownloadEvidence` returns a JSON `{url}` for S3 but **streams the file directly** on the local backend. The client always did `fetchJSON`, so on the default file backend the download silently failed (binary parsed as JSON). `api.downloadEvidence` now branches on Content-Type; `Objectives.vue` opens the URL for S3 and saves a blob for local.

## Tests
- `tests/test_evidence_download.py` — pins the contract the client depends on: on the file backend, `GET /evidence/:id/download` returns the raw bytes with the file's content type, not a JSON envelope. Green against the stack.
- **#50 testing note:** it's a cron/CLI path with no HTTP trigger and no Go DB test harness, so it isn't reachable from the integration suite. Build-verified; the change is a 1:1 mirror of the changelog logging in the API update handlers. (A testable manager trigger would be a separate improvement.)

Closes #50
Closes #31

### DIFF
--- a/cmd/isms/manager.go
+++ b/cmd/isms/manager.go
@@ -101,7 +101,12 @@ func manageOrg(ctx context.Context, d *db.DB, orgID int, quiet bool) (created, b
 				if err := d.UpdateSupplier(ctx, orgID, &s); err == nil {
 					backfilled++
 					if changes := db.DiffFields("supplier", int64(s.ID), "system", "automated next_review backfill", before, s.ToChangeMap()); len(changes) > 0 {
-						_ = d.LogChanges(ctx, orgID, changes)
+						// Unlike an HTTP handler (which has a response as a signal),
+						// the cron's only signal is stdout — surface a silent
+						// audit-trail failure instead of swallowing it.
+						if err := d.LogChanges(ctx, orgID, changes); err != nil && !quiet {
+							fmt.Printf("[warn] changelog write failed for supplier %d: %v\n", s.ID, err)
+						}
 					}
 				}
 			}
@@ -118,7 +123,9 @@ func manageOrg(ctx context.Context, d *db.DB, orgID int, quiet bool) (created, b
 				if err := d.UpdateSystem(ctx, orgID, &sys); err == nil {
 					backfilled++
 					if changes := db.DiffFields("system", int64(sys.ID), "system", "automated next_review backfill", before, sys.ToChangeMap()); len(changes) > 0 {
-						_ = d.LogChanges(ctx, orgID, changes)
+						if err := d.LogChanges(ctx, orgID, changes); err != nil && !quiet {
+							fmt.Printf("[warn] changelog write failed for system %d: %v\n", sys.ID, err)
+						}
 					}
 				}
 			}
--- a/cmd/isms/manager.go
+++ b/cmd/isms/manager.go
@@ -96,9 +96,13 @@ func manageOrg(ctx context.Context, d *db.DB, orgID int, quiet bool) (created, b
 	if err == nil {
 		for _, s := range suppliers {
 			if s.NextReview == nil || s.NextReview.IsZero() {
+				before := s.ToChangeMap()
 				s.CalculateNextReview()
 				if err := d.UpdateSupplier(ctx, orgID, &s); err == nil {
 					backfilled++
+					if changes := db.DiffFields("supplier", int64(s.ID), "system", "automated next_review backfill", before, s.ToChangeMap()); len(changes) > 0 {
+						_ = d.LogChanges(ctx, orgID, changes)
+					}
 				}
 			}
 		}
@@ -109,9 +113,13 @@ func manageOrg(ctx context.Context, d *db.DB, orgID int, quiet bool) (created, b
 	if err == nil {
 		for _, sys := range systems {
 			if sys.NextReview == nil || sys.NextReview.IsZero() {
+				before := sys.ToChangeMap()
 				sys.CalculateNextReview()
 				if err := d.UpdateSystem(ctx, orgID, &sys); err == nil {
 					backfilled++
+					if changes := db.DiffFields("system", int64(sys.ID), "system", "automated next_review backfill", before, sys.ToChangeMap()); len(changes) > 0 {
+						_ = d.LogChanges(ctx, orgID, changes)
+					}
 				}
 			}
 		}
--- a/tests/test_evidence_download.py
+++ b/tests/test_evidence_download.py
@@ -5,6 +5,12 @@ streams the file directly. The frontend now branches on Content-Type to handle
 both. This test pins the backend contract the frontend relies on: on the file
 backend, GET /evidence/:id/download returns the raw file bytes with the file's
 content type — NOT a JSON envelope — so the client can save it.
+
+Coverage gap: the S3 backend path (JSON {url} response) is not covered here —
+the test stack runs ISMS_STORAGE_BACKEND=file and there's no mocked S3 (minio/
+localstack) fixture. The server-side S3 path is unchanged by this fix; only the
+client gained the Content-Type branch. Add a mock-S3 fixture if that path needs
+regression coverage.
 """
 import uuid
 
--- a/tests/test_evidence_download.py
+++ b/tests/test_evidence_download.py
@@ -1,0 +1,49 @@
+"""Regression for #31: evidence download on the local (file) blob backend.
+
+The S3 backend returns JSON {url} (a presigned link); the local/file backend
+streams the file directly. The frontend now branches on Content-Type to handle
+both. This test pins the backend contract the frontend relies on: on the file
+backend, GET /evidence/:id/download returns the raw file bytes with the file's
+content type — NOT a JSON envelope — so the client can save it.
+"""
+import uuid
+
+import requests
+from conftest import ADMIN_EMAIL
+
+
+def test_evidence_download_local_backend_streams_file_bytes(api_url, admin_headers):
+    suffix = uuid.uuid4().hex[:8]
+    auth = {"Authorization": admin_headers["Authorization"]}  # no JSON CT for multipart
+
+    p = requests.post(f"{api_url}/programs", headers=admin_headers, json={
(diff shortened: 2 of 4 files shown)


unidoc-alip (changes requested):
Both fixes are correct and well-scoped. The `DiffFields`/`LogChanges` pattern for #50 mirrors existing API handlers exactly, and both `ToChangeMap` methods include `next_review` so the diff always fires. The evidence download fix for #31 correctly branches on Content-Type and the Vue handler follows the standard blob-download pattern. Two maintainability items and two nits below.

---

## Findings

### [should-fix] `cmd/isms/manager.go` lines 104 & 121 — silent `LogChanges` error

**Problem**

The PR uses `_ = d.LogChanges(ctx, orgID, changes)`, which matches every standalone `LogChanges` call in the API layer (`server.go:2483,2650,2765,2968`, `api_collab.go:2007,2106,2456`). That pattern is correct for HTTP handlers — a changelog failure shouldn't roll back a successful update or change the response status. However, the cron is different: there is no HTTP response; stdout is the only signal an operator sees. A silent `LogChanges` failure means the manager reports "backfilled N" while the audit trail is silently broken — which undermines the entire point of this fix.

**Proof**

```
internal/isms/api/server.go:2483   _ = s.db.LogChanges(ctx, orgID, changes)  ← API: silent OK
internal/isms/api/server.go:2650   _ = s.db.LogChanges(ctx, orgID, changes)  ← API: silent OK
internal/isms/api/api_collab.go:2007  _ = s.db.LogChanges(ctx, orgID, changes)  ← API: silent OK
cmd/isms/manager.go:104            _ = d.LogChanges(ctx, orgID, changes)     ← cron: silent = invisible
```

API handlers have HTTP responses as an implicit success signal. The cron does not.

**Fix**

Replace the silent discard with a quiet-respecting warning print in both blocks:

```go
// supplier block
if changes := db.DiffFields("supplier", int64(s.ID), "system",
    "automated next_review backfill", before, s.ToChangeMap()); len(changes) > 0 {
    if err := d.LogChanges(ctx, orgID, changes); err != nil && !quiet {
        fmt.Printf("[warn] changelog write failed for supplier %d: %v\n", s.ID, err)
    }
}

// system block
if changes := db.DiffFields("system", int64(sys.ID), "system",
    "automated next_review backfill", before, sys.ToChangeMap()); len(changes) > 0 {
    if err := d.LogChanges(ctx, orgID, changes); err != nil && !quiet {
        fmt.Printf("[warn] changelog write failed for system %d: %v\n", sys.ID, err)
    }
}
```

---

### [should-fix] `web/src/api.js` lines 444–450 — raw `fetch` + inline 401 duplicates existing helpers

**Problem**

Every other custom fetch helper in this file uses `safeFetch` + `checkAuth`. The new `downloadEvidence` bypasses both:

| function | fetch wrapper | 401 handler |
|---|---|---|
| `postJSON` (line 93) | `safeFetch` | `checkAuth(res)` |
| `putJSON` (line 105) | `safeFetch` | `checkAuth(res)` |
| `deleteJSON` (line 119) | `safeFetch` | `checkAuth(res)` |
| `uploadForm` (line 137) | `safeFetch` | `checkAuth(res)` |
| **`downloadEvidence` (PR)** | **raw `fetch`** | **inlined copy** |

Using raw `fetch` skips `safeFetch`'s try-catch, so network errors from `downloadEvidence` won't carry `err.isNetwork = true` — inconsistent with every other API call. The inline 401 block (lines 446–450) is a byte-for-byte copy of `checkAuth` (lines 82–90) that will silently diverge if `checkAuth` changes.

**Proof**

`safeFetch` (api.js:72–80) wraps `fetch` with network-error tagging:
```js
async function safeFetch(url, opts) {
  try {
    return await fetch(url, opts)
  } catch (e) {
    const err = new Error(e.message || 'Network error')
    err.isNetwork = true   // ← callers can distinguish network vs HTTP errors
    throw err
  }
}
```

`checkAuth` (api.js:82–90) handles 401 centrally:
```js
function checkAuth(res) {
  if (res.status === 401) {
    clearApiToken()
    window.dispatchEvent(new Event('isms:unauthorized'))
    const err = new Error('Authentication required')
    err.status = 401
    throw err
  }
}
```

The PR's inline block (api.js:446–450) is a verbatim copy of `checkAuth` with no functional difference.

**Fix**

Use the same helpers as the other four functions — comes out 4 lines shorter:

```js
downloadEvidence: async (id) => {
  const res = await safeFetch(`${API}/evidence/${id}/download`, { headers: getHeaders() })
  checkAuth(res)
  if (!res.ok) {
    const e = new Error(`${res.status} ${res.statusText}`)
    e.status = res.status
    throw e
  }
  if ((res.headers.get('content-type') || '').includes('application/json')) {
    return await res.json()
  }
  const cd = res.headers.get('content-disposition') || ''
  const m = /filename="?([^"]+)"?/.exec(cd)
  return { blob: await res.blob(), filename: m ? m[1] : `evidence-${id}` }
},
```

---

### [nit] `web/src/views/Objectives.vue` ~line 1117 — no `else` fallback

**Problem**

`else if (result?.blob)` has no `else` branch. If `api.downloadEvidence` returns an unexpected shape (`{}`, `null`, a future third backend), the download silently does nothing with no user feedback. The `catch` below only handles thrown errors; a wrong-shaped return is a silent no-op.

**Proof**

```js
if (result?.url) {
  window.open(result.url, '_blank')
} else if (result?.blob) {
  // ...blob download...
}
// ← if result is {} or null: nothing happens, no error shown
```

**Fix**

```js
} else {
  error.value = 'Download failed: unexpected response from server'
}
```

---

### [nit] `tests/test_evidence_download.py` — S3 backend path untested

**Problem**

The test pins the file backend contract (raw bytes, non-JSON content-type) — good. But issue #31's acceptance criteria included "still works on the S3 backend," and the client-side branch `if (content-type includes application/json) return res.json()` is new code in this PR with no coverage. The server-side S3 path is unchanged so regression risk is low, but a future change to the S3 response shape would be invisible to the suite.

**Proof**

```python
# tests/test_evidence_download.py — only one test function (49 lines total)
def test_evidence_download_local_backend_streams_file_bytes(api_url, admin_headers):
    ...
# No test for S3 backend (JSON {url, title, content_type} response shape)
```

**Fix**

Two options:

_Option A — document the gap_ (low effort): add a comment to the test file noting the S3 path requires a live/mocked S3 bucket and was verified manually.

_Option B — add a mock-S3 test_ (closes the gap properly): add a minio/localstack fixture asserting the server returns `{"url": "...", "title": "...", "content_type": "..."}` with `Content-Type: application/json` on the S3 backend, and that the client routes it to `window.open`.

unidoc-ahall:
Thanks @unidoc-alip — all four:

1. **Silent cron LogChanges (should-fix)** — both backfill blocks now warn on a changelog write failure (`if err != nil && !quiet`). You're right that the cron has no HTTP response, so swallowing it would hide exactly the audit-trail break this PR fixes.
2. **Raw fetch in downloadEvidence (should-fix)** — switched to `safeFetch` + `checkAuth`, matching the other helpers (network errors now carry `isNetwork`; 401 handling can't drift from `checkAuth`).
3. **Missing else (nit)** — `downloadEv` now sets an error on an unexpected response shape.
4. **S3 path untested (nit)** — documented the gap in the test docstring (Option A): the env is file-backend only with no mock-S3; the server-side S3 path is unchanged and only the client gained the Content-Type branch.

Build green; `test_evidence_download.py` still passes.

unidoc-alip (approved):
looks good to me

PR #92, opened 2026-06-24
fix: valid task types in manual form + drop dead 'control' title-resolver case (#32, #37)

### LINKED ISSUE #32: Simplify manual task creation — remove auto-only/invalid task types from the form
## Summary

Creating a task from the Inbox or Tasks form fails for most "type" choices — the
dropdowns offer task types the database rejects. The right fix is a usability one,
not just making the form match the DB.

## Context — why this matters at the system level

Review and check-in reminders (risk reviews, legal reviews, objective check-ins,
supplier/system/document reviews) are **auto-generated** by the overdue engine
(`internal/isms/db/overdue.go` → `CreateOverdueReviewTasks`). Users don't create
those by hand.

So the manual "create task" form is for **ad-hoc** tasks. Yet its type dropdown
offers system/auto concepts — `risk_review`, `objective_checkin`,
`corrective_action`, `supplier_review`, … — that:
1. the DB CHECK rejects, so creation silently fails, and
2. don't belong in a user's hands anyway — they're how the system categorizes the
   tasks *it* generates.

Evidence: the DB allows `general, review, incident_followup, audit_followup,
ca_followup, change_followup, onboarding, offboarding, training, other`
(`migrations/…initial_schema.sql:345`). The Tasks form offers 7 values outside
that list; the Inbox form offers 3.

## The decision (usability)

For a human creating an ad-hoc task, the type is mostly noise. Two sensible
options:

1. **Drop the type picker** from manual creation — default to `general`.
   Simplest; nothing to get wrong. Type stays a system concern for
   auto-generated tasks.
2. **Offer only the few types that make sense manually** (e.g. `general`,
   `other`) — if there is real value in letting users categorize ad-hoc tasks.

Recommendation: **(1)** unless we have a concrete reason users categorize manual
tasks. Either way, the granular review types leave the manual form.

## Acceptance criteria

- [ ] Manual task creation cannot fail on an invalid type.
- [ ] Auto-generated review/check-in reminders are unaffected.
- [ ] The type taxonomy stays a system concern, not a confusing user choice.

Frontend-only, no schema change.

### LINKED ISSUE #37: Remove dead 'control' case from the entity title-resolver
## Summary

Remove the dead `case "control"` from the entity title-resolver — it duplicates
`case "document"` and is never reached.

## Detail

`internal/isms/api/api_references.go:269` has a `case "control"` that resolves a
control by `document_id` and returns its title — identical to `case "document"`.
It never executes: controls are linked as references of type `document` (the
frontend maps control→document in `web/src/components/ReferenceManager.vue:182`),
so no reference ever carries type `control`.

Controls are unchanged — they remain documents with frontmatter type `control`.
This only removes dead, redundant code.

## Acceptance criteria

- [ ] Dead `case "control"` removed; control links still resolve correctly via the
      document path.

No schema, no functional change.

unidoc-ahall:
Two cleanups, conflict-free with the open #91.

## #32 — Add Task form offered invalid task types
The create form (and the type filter) listed types that aren't in the `task_type` enum (`risk_review`, `supplier_review`, `access_review`, `legal_review`, `document_review`, `objective_checkin`, `corrective_action`). `handleCreateTask` validates `task_type`, so picking any of them **400'd** on submit — manual task creation was broken for everything except General/Other. The create form now offers only valid **manual** types (general, review, onboarding, offboarding, training, other); the `*_followup` types are system-generated so they're not in the manual form. The filter lists all real types (so you can filter auto-generated tasks too).

## #37 — dead 'control' case in the title resolver
`resolveEntityTitle` had a `case "control"` that resolved via document lookup — but references use type `"document"` for documents; `"control"` is never a reference type (only a frontmatter *subtype* for display). The case duplicated the `"document"` branch. Pure dead-code removal, no behaviour change.

## Tests
- `tests/test_task_types.py` — pins the form↔enum contract: every manual type the form offers is accepted by `POST /tasks`; the removed bogus types are rejected (400). Green against the stack.
- #37 is dead-code removal (the case was unreachable), so no behaviour to test — build-verified.

Closes #32
Closes #37

### DIFF
--- a/internal/isms/api/api_references.go
+++ b/internal/isms/api/api_references.go
@@ -266,22 +266,6 @@ func (s *Server) resolveEntityTitle(ctx context.Context, orgID int, entityType, 
 		}
 		return p.Title
 
-	case "control":
-		// Controls are documents — resolve via document_id lookup.
-		st, err := s.storeForOrg(ctx, orgID)
-		if err != nil {
-			return entityID
-		}
-		path := st.FindDocumentByID(entityID)
-		if path == "" {
-			return entityID
-		}
-		doc, err := st.LoadDocument(path)
-		if err != nil {
-			return entityID
-		}
-		return doc.Frontmatter.Title
-
 	case "document":
 		st, err := s.storeForOrg(ctx, orgID)
 		if err != nil {
--- a/tests/test_task_types.py
+++ b/tests/test_task_types.py
@@ -1,0 +1,42 @@
+"""Regression for #32: the manual task-creation form must only offer task types
+the API actually accepts.
+
+The Add Task form used to list types (risk_review, supplier_review, ...) that
+aren't in the task_type enum at all, so picking them 400'd. This pins the
+contract: every type the form now offers is accepted, and the removed bogus ones
+are rejected. (The *_followup types are valid but system-generated, so they're
+intentionally not in the manual form — only in the filter.)
+"""
+import uuid
+
+import requests
+from conftest import ADMIN_EMAIL
+
+# The types the manual create form offers after #32.
+MANUAL_TYPES = ["general", "review", "onboarding", "offboarding", "training", "other"]
+
+# Types the old form wrongly offered — not in the enum.
+REMOVED_BOGUS_TYPES = [
+    "risk_review", "supplier_review", "access_review",
+    "legal_review", "document_review", "objective_checkin", "corrective_action",
+]
+
+
+def test_manual_task_types_are_accepted(api_url, admin_headers):
+    for t in MANUAL_TYPES:
+        r = requests.post(f"{api_url}/tasks", headers=admin_headers, json={
+            "title": f"task-type {t} {uuid.uuid4().hex[:8]}",
+            "task_type": t,
+            "assignee": ADMIN_EMAIL,
+        })
+        assert r.status_code in (200, 201), f"manual type {t!r} must be accepted: {r.status_code} {r.text}"
+
+
+def test_removed_form_types_are_rejected(api_url, admin_headers):
+    for t in REMOVED_BOGUS_TYPES:
+        r = requests.post(f"{api_url}/tasks", headers=admin_headers, json={
+            "title": f"bogus-type {t} {uuid.uuid4().hex[:8]}",
+            "task_type": t,
+            "assignee": ADMIN_EMAIL,
+        })
+        assert r.status_code == 400, f"removed type {t!r} should be rejected by the API: {r.status_code} {r.text}"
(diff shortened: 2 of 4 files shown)


unidoc-alip (changes requested):
The Tasks.vue fix and the `api_references.go` dead-code removal are both correct and complete. However, `Inbox.vue` was not updated: its create-task form still offers `implementation`, `corrective_action`, and `change_request` — none of which are in `db.TaskTypes` — so `validateEnum` at `api_collab.go:1933` returns 400 on submit. Issue #32 explicitly listed the Inbox form as part of the bug scope, so #32 is only half-fixed.

---

## Findings

|---|---|---|---|
| [1](#finding-1) | [must-fix] | `web/src/views/Inbox.vue` | 266 |

---

### Finding 1 · [must-fix] · `web/src/views/Inbox.vue:266`

**The Inbox create-task form still offers three invalid task types that 400 on submit.**

Verified call chain:

| Step | Location | What happens |
|---|---|---|
| Invalid option offered | `Inbox.vue:266–269` | `<option value="implementation">`, `<option value="corrective_action">`, `<option value="change_request">` |
| No client-side guard | `Inbox.vue:1037` | `api.createTask({...newTask.value, ...})` — direct spread, no type check |
| Server validates | `api_collab.go:1933` | `validateEnum("task_type", t.TaskType, db.TaskTypes)` |
| Enum definition | `db/tasks.go:13` | `TaskTypes` does not include `implementation`, `corrective_action`, or `change_request` → **HTTP 400** |

Issue #32 says "The Inbox form offers 3 [invalid types]" as an explicit acceptance criterion — this was in scope and was missed.

**Fix** — replace the `<select>` block at `Inbox.vue:265–271` with the same manual set used in Tasks.vue:

```html
<!-- Manual task types only. *_followup are system-generated (#32). -->
<option value="general">General</option>
<option value="review">Review</option>
<option value="onboarding">Onboarding</option>
<option value="offboarding">Offboarding</option>
<option value="training">Training</option>
<option value="other">Other</option>
```

The `review` option already present is valid and stays. The default at `Inbox.vue:760` (`task_type: 'general'`) is already safe. No new test needed — `test_removed_form_types_are_rejected` already confirms these values are API-rejected.

unidoc-ahall:
Good catch @unidoc-alip — fixed. `Inbox.vue` now offers the same manual set as Tasks.vue (general / review / onboarding / offboarding / training / other); the three invalid types (`implementation`, `corrective_action`, `change_request`) are gone. Default (`task_type: 'general'`) was already safe.

One correction to your note: `test_removed_form_types_are_rejected` only listed `corrective_action` of the three — `implementation` and `change_request` weren't pinned, so I added them to the bogus list. Both confirmed 400 against the stack; `test_task_types.py` green (2 passed).

unidoc-alip (approved):
looks good to me

PR #93, opened 2026-06-24
feat(invite): resend-invite action for pending invited users (#42)

### LINKED ISSUE #42: Add a resend-invite action for stranded/pending invited users
## Problem

There is no way to re-issue an invite. `handleInviteUser` (`api_auth.go:608`) creates the user row + org membership **before** `SendVerification`, with no rollback on SMTP failure — so if the invite email bounces or SMTP fails, the user is left inactive with no in-UI recovery path. The members list (`Admin.vue`) only offers Remove.

## Verified: no schema change

The mechanism reuses what already exists:
- `email_verifications` is a plain `INSERT` per token (`email_verify.go:18`) — multiple rows per user are allowed, so re-issuing just creates a fresh valid token. No constraint blocks it.
- `SendVerification` (`mail.go:101`) and the user/member lookups already exist.

So this is purely a new endpoint + UI action → **fits a patch release.**

## Scope

- New admin/manager endpoint (e.g. `POST /auth/users/:id/resend-invite`) for pending/inactive members: generate a new token via `CreateEmailVerification` and resend via `SendVerification`.
- A "Resend invite" action per pending/inactive member in `Admin.vue`; surface success/failure to the admin.
- Optional polish (still no schema): mark the user's prior unused `verify` tokens as used before issuing a new one, so only one invite link is live.

## Acceptance criteria

- A "Resend invite" action appears per pending/inactive member.
- The endpoint re-issues the token and retries the email.
- Resending does not error and surfaces success/failure to the admin.

## Out of scope

- Reworking the invite-create ordering / SMTP-failure rollback (separate hardening — resend is the recovery path).

unidoc-ahall:
## #42 — resend a lost/expired invite
An invited user who never verified (email lost or the 72h token expired) was stranded — no way to get a new link short of remove + re-invite.

**Backend:** `POST /auth/resend-invite {email}` (admin/manager). Issues a fresh verification token and re-sends the branded email, but only for someone who is a **member of the caller's org** and still **pending** (`active=false`). An already-active user gets 409; unknown user 404. Reuses the same token + `SendVerificationBranded` path as the original invite (org-branded sender + per-org link, via `orgMail`).

**Frontend:** the Admin → Members list shows a **Pending** badge for unverified members and a **Resend invite** action for them.

## Tests
`tests/test_resend_invite.py` covers the auth + eligibility guards: non-admin → 403, active user → 409, unknown email → 404, missing email → 400.

> Testing note: the happy path sends an email, and the test stack has no SMTP configured (same limitation as #16's per-org email work), so the success path isn't integration-tested. All four asserted guards run **before** the mailer check, so they're covered; the send path mirrors the existing, reviewed invite handler.

Closes #42

### DIFF
--- a/internal/isms/api/api_auth.go
+++ b/internal/isms/api/api_auth.go
@@ -702,6 +702,65 @@ func (s *Server) handleInviteUser(c echo.Context) error {
 	})
 }
 
+// handleResendInvite re-sends the verification email to a pending (not yet
+// verified) invited user — for invites that were lost or expired (#42).
+func (s *Server) handleResendInvite(c echo.Context) error {
+	if err := requireRole(c, "admin", "manager"); err != nil {
+		return err
+	}
+	orgID := getOrgID(c)
+	ctx := c.Request().Context()
+
+	var req struct {
+		Email string `json:"email"`
+	}
+	if err := c.Bind(&req); err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "invalid request")
+	}
+	if req.Email == "" {
+		return echo.NewHTTPError(http.StatusBadRequest, "email is required")
+	}
+
+	user, err := s.db.GetUserByEmail(ctx, req.Email)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusNotFound, "user not found")
+	}
+	// Must be a member of this org — don't resend across organizations.
+	if _, err := s.db.GetOrgMember(ctx, orgID, user.ID); err != nil {
+		return echo.NewHTTPError(http.StatusNotFound, "user is not a member of this organization")
+	}
+	// Only pending invites can be resent; an active user has already accepted.
+	if user.Active {
+		return echo.NewHTTPError(http.StatusConflict, "user has already accepted the invite")
+	}
+
+	if s.mailer == nil || !s.mailer.Enabled() {
+		return echo.NewHTTPError(http.StatusInternalServerError, "email not configured (SMTP_HOST)")
+	}
+
+	// Fresh verification token (72h), same as the original invite.
+	raw := make([]byte, 32)
+	rand.Read(raw)
+	token := hex.EncodeToString(raw)
+	hash := sha256.Sum256([]byte(token))
+	if err := s.db.CreateEmailVerification(ctx, user.ID, hex.EncodeToString(hash[:])); err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "creating verification: "+err.Error())
+	}
+
+	m := s.orgMail(ctx, orgID)
+	if err := s.mailer.SendVerificationBranded(user.Email, user.Name, m.PublicURL, token, m.Branding); err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "sending email: "+err.Error())
+	}
+
+	s.logAndNotify(ctx, orgID, &db.Activity{
+		Actor:  getUserEmail(c),
+		Action: "user_invite_resent",
+		Detail: fmt.Sprintf("Resent invite to %s", user.Email),
+	})
+
+	return c.JSON(http.StatusOK, map[string]string{"status": "invite_resent", "email": user.Email})
+}
+
 // --- Verify email + set password ---
 
 type verifyEmailRequest struct {
--- a/internal/isms/api/api_auth.go
+++ b/internal/isms/api/api_auth.go
@@ -738,18 +738,27 @@ func (s *Server) handleResendInvite(c echo.Context) error {
 		return echo.NewHTTPError(http.StatusInternalServerError, "email not configured (SMTP_HOST)")
 	}
 
+	// Invalidate any still-live invite tokens before issuing a new one, so a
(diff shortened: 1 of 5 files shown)


unidoc-alip (changes requested):
The resend-invite feature is well-scoped and the implementation is clean — correct RBAC, cross-org membership guard, and `crypto/rand` for token generation. One security hygiene gap: old verification tokens are not invalidated before the new one is issued, leaving multiple live invite links active simultaneously. The rest are minor nits.

---

## Findings

| # | Severity     | File | Line |
|---|--------------|------|------|
| [1](#finding-1) | [should-fix] | `internal/isms/api/api_auth.go` | 746 |
| [2](#finding-2) | [nit]        | `internal/isms/api/api_auth.go` | 747 |

---

### Finding 1 · [should-fix] · `internal/isms/api/api_auth.go:746`

**Old verification tokens for this user are not invalidated before issuing the new one.**

`CreateEmailVerification` (db/email_verify.go:18) does a plain INSERT with no cleanup. Every
call to `POST /auth/resend-invite` accumulates a new valid row while all previous rows remain
`used_at IS NULL` and within their 72h window.

**PoC:**
```
# 1. Admin invites alice@example.com → token T1 emailed, row R1 inserted
POST /auth/invite {"email":"alice@example.com","role":"reader"}

# 2. Alice doesn't click. 48h later admin resends → token T2 emailed, row R2 inserted
POST /auth/resend-invite {"email":"alice@example.com"}

# 3. T1 still passes LookupEmailVerification (used_at IS NULL, expires_at > now())
GET /auth/verify?token=<T1>  →  200 OK, alice is now active
# T2 is also still valid and can be used independently.
```

An attacker who captured T1 from a compromised mailbox can verify as Alice even after the
admin resent — defeating the purpose of the resend.

**Fix** — add a bulk-invalidate helper to `internal/isms/db/email_verify.go`:
```go
// InvalidatePendingVerifications marks all unused verify tokens for a user as used.
func (d *DB) InvalidatePendingVerifications(ctx context.Context, userID int) error {
    _, err := d.pool.Exec(ctx, `
        UPDATE email_verifications
        SET used_at = now()
        WHERE user_id = $1 AND purpose = 'verify' AND used_at IS NULL
    `, userID)
    return err
}
```
Call it in `handleResendInvite` immediately before `CreateEmailVerification` (line 746):
```go
if err := s.db.InvalidatePendingVerifications(ctx, user.ID); err != nil {
    return echo.NewHTTPError(http.StatusInternalServerError, "invalidating old tokens")
}
if err := s.db.CreateEmailVerification(ctx, user.ID, hex.EncodeToString(hash[:])); err != nil {
    return echo.NewHTTPError(http.StatusInternalServerError, "creating verification token")
}
```

---

### Finding 2 · [nit] · `internal/isms/api/api_auth.go:747`

**`err.Error()` is returned verbatim in the HTTP 500 response body, leaking internal DB and SMTP error strings.**

Applies to two lines in the new handler:
- Line 747: `"creating verification: "+err.Error()`
- Line 752: `"sending email: "+err.Error()`

**PoC** — if the DB pool is exhausted or a constraint fires, the caller receives:
```json
{"message":"creating verification: ERROR: duplicate key value violates unique constraint \"email_verifications_pkey\" (SQLSTATE 23505)"}
```
SMTP failures similarly expose the mail server error string.

**Fix** — log server-side, return a generic message:
```go
if err := s.db.CreateEmailVerification(ctx, user.ID, hex.EncodeToString(hash[:])); err != nil {
    log.Printf("resend-invite: CreateEmailVerification user=%d: %v", user.ID, err)
    return echo.NewHTTPError(http.StatusInternalServerError, "failed to create verification token")
}
// ...
if err := s.mailer.SendVerificationBranded(user.Email, user.Name, m.PublicURL, token, m.Branding); err != nil {
    log.Printf("resend-invite: SendVerificationBranded user=%s: %v", user.Email, err)
    return echo.NewHTTPError(http.StatusInternalServerError, "failed to send invite email")
}
```
Note: the same pattern exists in `handleInviteUser` (lines 677, 688) — both should be cleaned up together.

unidoc-ahall:
Thanks @unidoc-alip — all three:

1. **Old tokens not invalidated (should-fix)** — added `InvalidateEmailVerifications(userID, "verify")` (UPDATE … SET used_at = now() WHERE purpose = $2 AND used_at IS NULL), called immediately before `CreateEmailVerification`. A leaked/intercepted earlier invite link is now dead the moment a new one is issued.
2. **Leaked DB error string (nit)** — the handler no longer returns `err.Error()`; it logs server-side and returns a generic message. I applied it to **both** error paths (create-verification *and* send-email), not just line 748. Worth noting the sibling `handleInviteUser` send path has the same leak — I left it out of this PR's scope, happy to file a tidy-up if you'd like.
3. **membersError not reset (nit)** — `resendInvite` now resets it at the start.

`test_resend_invite.py` (4 guards) green against the stack. The invalidation + happy path sit past the SMTP-enabled guard, so they share the SMTP-less stack's blind-spot (same as #16/#42).

unidoc-alip (approved):
looks good to me

PR #94, opened 2026-06-24
fix(ui): suggestions read clearly as 'under review' (#88)

### LINKED ISSUE #88: Suggestions should clearly read as "under review" in the UI
# Suggestions should clearly read as "under review" in the UI

## Problem
A manager (María) opened a suggested **task**, expected to see a real task, and tried to approve it — without realising it was a *suggestion* still awaiting apply, not a live entity. The UI doesn't make it obvious that a suggestion is a **proposal under review** that becomes a real record only once a manager applies it.

This is low-stakes today but becomes important with the role realignment (#23): once contributors and AI agents can *only* propose via suggestions, the suggestion queue is the primary contribution path — so it must be unmistakable what a suggestion is and what state it's in.

## Desired outcome
- A suggestion is visibly distinct from a live entity — a clear **"under review" / "pending"** state (badge/label), not something that looks like a finished task/risk/etc.
- It's obvious that a suggestion needs a manager to **apply** it before it exists as a real record, and who can do that.
- Suggestions are discoverable in an "under review" state (e.g. the inbox/suggestions view reads as a review queue, not a list of done things).
- Wording on the action makes the two-step nature clear (propose → apply), so "approve" doesn't read as "this already happened".

## Scope / sequencing
UI/UX only — no new backend. Suggestion-related, so **land this in the #23 role-realignment PR** (which already touches the suggestion/role surface), once the open security PR (#24/#27) is merged. This issue just captures the requirement so it isn't lost.

Related: #23 (role realignment — contributor suggest-only, reader read-only).

unidoc-ahall:
## #88 — suggestions should read as 'under review'
A manager (María) approved a suggestion expecting to see a live task, not realising it was a *proposal* still awaiting apply. The `SuggestionPanel` showed only a small raw status badge ("open").

**Change (presentational only):**
- Status labels now say what the state means: `open` → **Awaiting review**, `in_review` → **Under review** (applied/rejected/withdrawn unchanged).
- A one-line banner above pending suggestions: *"Under review — proposed changes that take effect only once a manager applies them."*

No API or behaviour change — display copy only, so there's nothing to integration-test (no JS test runner in the repo); web build is green. The richer suggestion queue stays as-is.

Closes #88

### DIFF
--- a/web/src/components/SuggestionPanel.vue
+++ b/web/src/components/SuggestionPanel.vue
@@ -4,10 +4,12 @@
     <div v-if="loading" class="h-10 bg-slate-800 rounded animate-pulse" />
     <div v-else-if="suggestions.length === 0 && !showCreate" class="text-sm text-slate-600 italic py-2">No suggestions yet.</div>
     <div v-else class="space-y-2">
-      <!-- Make the nature of suggestions unmistakable: they're proposals under
-           review, not changes that have happened yet (#88). -->
+      <!-- Make the nature of suggestions unmistakable: they're proposals
+           pending a decision, not changes that have happened yet (#88).
+           openCount covers both 'open' and 'in_review', so the wording must
+           fit both — not just the 'Under review' state. -->
       <p v-if="openCount > 0" class="text-[11px] text-amber-400/90 bg-amber-500/5 border border-amber-500/15 rounded-md px-2.5 py-1.5">
-        Under review — proposed changes that take effect only once a manager applies them.
+        Proposals pending — these changes take effect only once a manager applies them.
       </p>
       <div v-for="sg in suggestions" :key="sg.id"
         class="bg-slate-800/40 border border-slate-700/40 rounded-lg px-4 py-3 space-y-2">
--- a/web/src/components/SuggestionPanel.vue
+++ b/web/src/components/SuggestionPanel.vue
@@ -4,12 +4,17 @@
     <div v-if="loading" class="h-10 bg-slate-800 rounded animate-pulse" />
     <div v-else-if="suggestions.length === 0 && !showCreate" class="text-sm text-slate-600 italic py-2">No suggestions yet.</div>
     <div v-else class="space-y-2">
+      <!-- Make the nature of suggestions unmistakable: they're proposals under
+           review, not changes that have happened yet (#88). -->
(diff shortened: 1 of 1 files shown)


unidoc-alip (changes requested):
Minimal, focused presentational fix. The `statusLabel` function and informational banner both correctly address the confusion reported in #88 — suggestions now read as proposals awaiting manager action rather than applied changes. `openCount` is properly computed, the fallback in `statusLabel` handles any future statuses gracefully, and `statusClass` continues to use the raw status string so CSS colours are unaffected.

---

## Findings

| [1](#finding-1) | nit | `web/src/components/SuggestionPanel.vue` | 9 |

---

### Finding 1 · [nit] · `web/src/components/SuggestionPanel.vue:9`

**Banner heading "Under review" conflicts with the "Awaiting review" badge it appears alongside.**

`openCount` (line 117) counts both `open` and `in_review` statuses:
```js
const openCount = computed(() =>
  suggestions.value.filter(s => s.status === 'open' || s.status === 'in_review').length
)
```
The banner's `v-if="openCount > 0"` therefore fires for freshly submitted suggestions whose
status is `open` — but `statusLabel` maps `open` → **"Awaiting review"**, not "Under review".

**PoC — what a contributor sees after submitting a suggestion (before any manager picks it up):**
```
┌─────────────────────────────────────────────────────────────────┐
│ ⚠ Under review — proposed changes that take effect only once …  │  ← banner
├─────────────────────────────────────────────────────────────────┤
│ Update risk score                                               │
│ [Update]  [Awaiting review]                                     │  ← badge
└─────────────────────────────────────────────────────────────────┘
```
The banner says "Under review" while the badge on the same row says "Awaiting review" —
undercutting the very clarity that #88 aimed to add.

**Fix** — use copy that is accurate for both pending states:
```html
<!-- line 9 — before -->
Under review — proposed changes that take effect only once a manager applies them.

<!-- line 9 — after -->
Proposals pending — these changes take effect only once a manager applies them.
```
No logic change needed; only the visible string changes.

unidoc-ahall:
Good eye @unidoc-alip — fixed. The banner now reads **"Proposals pending — these changes take effect only once a manager applies them."**, which is accurate whether the suggestion is `open` (badge: Awaiting review) or `in_review` (badge: Under review), so the banner and badge no longer contradict each other. Copy-only; web build green.

unidoc-alip (approved):
looks good to me

PR #96, opened 2026-06-24
feat(landing): lead with the why (#45)

### LINKED ISSUE #45: Bring the WHY to the isms.sh landing page
## Context

We reworked the README to open with "Why yet another management system?" (WHY-first, #36). The landing page (`web/src/views/Landing.vue`) is the public front door, but it currently leads with what/features. It should carry the same WHY narrative so the message lands with visitors — not just a feature list.

## Scope

- The landing page **opens with the WHY**, before features.
- Copy is consistent with the README's WHY section — same belief, and reuse the wording where it fits. **The README WHY is the canonical reference.**
- Content / frontend only — no schema, no behavior change.

## The WHY (gist — README is canonical)

Management systems are complex, opaque, and treated as a compliance exercise — relevant for the audit, irrelevant the day after. We believe it doesn't have to work that way: built from the ground up, AI-first, with real discipline, a management system should be a living part of how you operate, not a burden on top of it. That's why it's open source — built to work for organizations of all sizes, and finally a credible option for smaller ones.

## Acceptance criteria

- [ ] Landing page leads with the WHY, before features.
- [ ] Message is consistent with the README's WHY (same belief, same words where it makes sense).
- [ ] Honest, no fluff.

unidoc-ahall:
Closes #45.

The landing page led with *what* the product is (tagline → pillars → features). This adds a **'Why another management system?'** section immediately after the hero, so the page leads with the conviction before the mechanics — adapted from the README's why voice (problem → belief → why open source). Copy-only, no logic.

### DIFF
--- a/web/src/views/Landing.vue
+++ b/web/src/views/Landing.vue
@@ -54,6 +54,16 @@
       </div>
     </section>
 
+    <!-- Why — lead with the conviction, not the feature list (#45) -->
+    <section class="relative z-10 max-w-3xl mx-auto px-8 pt-8 pb-16">
+      <h2 class="text-3xl font-bold text-center mb-8">Why another management system?</h2>
+      <div class="space-y-5 text-slate-400 leading-relaxed text-lg">
+        <p>Because the ones we have are complex, opaque, and treated as a compliance exercise — relevant for the audit, irrelevant the day after. And the stakes keep rising: as AI adoption accelerates and threat surfaces widen, a management system that doesn't stay alive is worse than useless.</p>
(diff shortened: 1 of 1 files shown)


unidoc-alip (approved):
Clean, copy-only addition that satisfies all three acceptance criteria from #45: the WHY section lands immediately after the hero (before pillars, features, and how-it-works), the copy tracks the README's WHY closely, and there's no fluff. One intentional voice deviation to note: the README's p2 opens with "We believe it doesn't have to work that way"; the landing's phrasing "It doesn't have to work this way" is more declarative, which suits the medium. Not a concern — flagged only for awareness.

PR #99, opened 2026-06-25
feat(review): per-cell word-level table diffs (#6, slice 1/2)

unidoc-ahall:
Part of #6 — the headline for 0.7.0.

**Problem (the real case María hit on clause 4.2):** tables are stored as HTML to keep colors/rich cells, and the WYSIWYG converts a markdown pipe table to HTML on the first edit. The diff views word-diffed that HTML source, so a two-cell edit rendered as a wholesale table replace — unreadable, the actual change buried.

**Fix:** a structural per-cell table diff (`useTableDiff.diffTables`) wired into **both** review diff views (Unified and the default Split). It renders both sides to a table DOM (markdown rendered to HTML first, so the md→HTML conversion is *not* shown as a change), matches rows by a key column, and highlights only the changed cells — word-level inside — keeping every unchanged cell's own colors/markup.

**Verification:** new `tests/test_e2e_review_diff.py` (Playwright, real browser) reproduces the case via API and asserts exactly the two edited cells are highlighted in Split **and** Unified, the conversion isn't shown as a change, and unchanged cells aren't flagged. Green twice against the stack (idempotent); runs in CI. The LCS core is also unit-checked. Shared diff primitives extracted to `useWordDiff.js` (DRY).

**Scope:** this is slice 1/2 — it satisfies the *table cell word-level highlights* criterion. Per-event timeline diff + the diff-scope toggle (the other criteria) land in a focused second PR. Migration-free, frontend-only.

Kept the E2E in its own file (not the test_e2e_browser monolith) to avoid the recurring EOF merge conflicts as review-UI work continues.

### DIFF
--- a/tests/test_e2e_review_diff.py
+++ b/tests/test_e2e_review_diff.py
@@ -104,9 +104,92 @@ def test_review_table_diff_renders_per_cell(pw_browser, table_review):
         _assert_per_cell_table_diff(page)
         page.screenshot(path="docs/screenshots/test_review_table_diff_split.png")
 
-        # 2. Unified view does too.
+        # 2. Unified view does too — and its stats count CELLS, not raw lines
+        #    (a md→HTML table must not read "+1 added -5 removed" for a 2-cell edit).
         page.locator("button:has-text('Unified')").first.click()
         _assert_per_cell_table_diff(page)
+        assert page.locator("text=2 added").count() >= 1, "unified stats should count 2 changed cells"
+        assert page.locator("text=5 removed").count() == 0, "stats must not use raw line counts for a table edit"
         page.screenshot(path="docs/screenshots/test_review_table_diff_unified.png")
+    finally:
+        ctx.close()
+
+
+# A second table that gets deleted outright in the same revision — this is what
+# broke the old sequential-counter pairing (the deletion shifted the index so the
+# changed table diffed against the wrong old table). #99 review must-fix.
+OLD_BODY_2 = """# Team Roster
+
+| Name | Role |
+|---|---|
+| Alice | Lead |
+| Bob | Dev |
+
+# Interested Parties
+
+| Party | Requirement |
+|---|---|
+| Customers | Confidentiality |
+| Suppliers | Security expectations |"""
+
+# Reviewer deletes the Team Roster table entirely and changes ONE cell in the
+# Interested Parties table (Customers requirement).
+NEW_BODY_2 = (
+    "# Interested Parties\n\n"
+    "<table><tbody>"
+    "<tr><th><p>Party</p></th><th><p>Requirement</p></th></tr>"
+    "<tr><td><p>Customers</p></td><td><p>Confidentiality and integrity</p></td></tr>"
+    "<tr><td><p>Suppliers</p></td><td><p>Security expectations</p></td></tr>"
+    "</tbody></table>"
+)
+
+
+@pytest.fixture(scope="module")
+def table_review_with_deletion(tokens):
+    """Approved baseline with TWO tables; reviewer deletes the first and changes
+    one cell in the second."""
+    t = tokens["admin"]
+    r1t = api("post", "/auth/login",
+              json={"email": R1[0], "password": R1[1], "organization": ORG},
+              expect_status=200).json()["token"]
+    doc_id = f"e2e-tabledel-{uuid.uuid4().hex[:8]}"
+    api("post", "/documents", t, json={
+        "folder": "iso27001", "filename": f"{doc_id}.md",
+        "document_id": doc_id, "title": "E2E Table Delete", "content": OLD_BODY_2,
+    }, expect_status=[200, 201])
+    rid0 = api("post", f"/documents/{doc_id}/reviews", t,
+               json={"reviewers": [R1[0]], "message": "baseline"},
+               expect_status=[200, 201]).json()["review_id"]
+    api("post", f"/reviews/{rid0}/approve", r1t,
+        json={"decision": "approved", "comment": "ok"}, expect_status=200)
+    api("post", f"/reviews/{rid0}/merge", t, json={}, expect_status=200)
+    rid = api("post", f"/documents/{doc_id}/reviews", t,
+              json={"reviewers": [R1[0]], "message": "Delete + edit"},
+              expect_status=[200, 201]).json()["review_id"]
+    api("put", f"/reviews/{rid}/content", r1t, json={"content": NEW_BODY_2}, expect_status=200)
+    yield rid
+    api("put", f"/reviews/{rid}/status", t, json={"status": "closed"})
+
+
+def test_review_table_diff_pairs_correctly_after_deletion(pw_browser, table_review_with_deletion):
+    ctx = pw_browser.new_context(viewport={"width": 1440, "height": 900})
+    page = ctx.new_page()
+    try:
+        do_login(page, ADMIN[0], ADMIN[1], then_goto=f"reviews/{table_review_with_deletion}")
+        page.locator("button:has-text('Changes')").first.click()
+        page.wait_for_load_state("networkidle")
+        page.locator(".doc-prose table").first.wait_for(state="visible", timeout=10000)
+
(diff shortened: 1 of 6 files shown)


unidoc-alip (changes requested):
**Verdict: REQUEST_CHANGES** — 1 must-fix, 1 should-fix, 2 nits

Slice 1/2 of #6: `useTableDiff.diffTables` wired into both Split and Unified review views so table edits show per-cell word-level highlights instead of a wholesale block diff. The core algorithm is sound and verified end-to-end against the María/clause-4.2 case — the happy path produces exactly the right output.

One must-fix confirmed by live browser test: the `tIdx` sequential counter in `SideBySideReview` mis-pairs old↔new tables when the diff contains a pure table deletion alongside a table change. The author sees an incorrect diff and may approve without understanding the actual change. One should-fix: the Unified view stats counter still uses raw line counts for table segments, so a 2-cell edit reads "+1 added -5 removed".

---

## Must-fix

### `SideBySideReview.vue:256` — `tIdx` counter mis-pairs tables when a pure deletion precedes a changed table

`changedOldTables` collects every old block that is both changed and a table — including old tables that were deleted outright with no new replacement. `tIdx` increments only for changed new tables. When a pure deletion inflates `changedOldTables`, the counter drifts and every subsequent changed table is diffed against the wrong old table.

**Reproduced live** (doc with Team Roster + Interested Parties tables; reviewer deletes Team Roster and changes one cell in Interested Parties):

- Expected: `td.tc-cell-change = 1`, `.tc-word-ins = ["and integrity"]`
- Actual: `td.tc-cell-change = 0`, `.tc-row-add = 2`, `.tc-row-del = 2` — Team Roster rows (Alice/Bob) are appended inside Interested Parties as deleted rows; the actual word-level change is completely lost

Root cause: `changedOldTables = [TeamRoster_old, InterestedParties_old]`. Processing `InterestedParties_modified` (the only changed new table) hits `tIdx=0` → `changedOldTables[0] = TeamRoster_old`. `diffTables(TeamRoster, InterestedParties)` finds no row-key matches → all rows become `tc-row-add`/`tc-row-del`.

The pairing IS present in the `diffBlocks` backtrack: it yields `DEL(TeamRoster)` then `INS(InterestedParties_modified)` as consecutive ops — that pairing is just never recorded. Fix: capture consecutive `del→ins` pairs from the edit script and use them in `rightBlocks` instead of the counter.

```js
// diffBlocks — replace the backtrack loop with one that also builds an edit script:
const ops = []
let i = m, j = n
while (i > 0 || j > 0) {
  if (i > 0 && j > 0 && oldBlocks[i-1].text === newBlocks[j-1].text) {
    ops.push({ type: 'eq', oi: i-1, ni: j-1 }); i--; j--
  } else if (j > 0 && (i === 0 || dp[i][j-1] >= dp[i-1][j])) {
    ops.push({ type: 'ins', ni: j-1 }); j--
  } else {
    ops.push({ type: 'del', oi: i-1 }); i--
  }
}
ops.reverse()
for (const op of ops) {
  if (op.type === 'eq') { oldChanged[op.oi] = false; newChanged[op.ni] = false }
}
const pairs = new Map()  // newIdx → oldIdx
for (let k = 0; k < ops.length - 1; k++) {
  if (ops[k].type === 'del' && ops[k+1].type === 'ins') {
    pairs.set(ops[k+1].ni, ops[k].oi); k++
  }
}
return { oldChanged, newChanged, pairs }

// rightBlocks — replace tIdx counter with pairs lookup:
const rightBlocks = computed(() =>
  newBlocks.value.map((b, i) => {
    const changed = diff.value.newChanged[i]
    if (changed && isTableBlock(b.text)) {
      const pairedOldIdx = diff.value.pairs?.get(i)
      const oldText = (pairedOldIdx != null && isTableBlock(oldBlocks.value[pairedOldIdx]?.text))
        ? oldBlocks.value[pairedOldIdx].text : ''
      const merged = diffTables(oldText, b.text)
      if (merged) return { ...b, changed: false, html: merged }
    }
    return { ...b, changed }
  })
)
// changedOldTables is no longer needed — remove it.
```

---

## Should-fix

### `TrackChanges.vue:320` — stats counter uses raw line counts for table segments

The `stats` computed counts `s.oldText.split('\n').length` / `s.newText.split('\n').length` for every `change` segment regardless of `s.isTable`. A markdown pipe table (5 lines) changed to a single-line HTML table produces "+1 added, -5 removed" in the Unified view header while the visual correctly shows 2 highlighted cells.

**Reproduced live**: Unified view for a 2-cell edit shows "+1 added  -5 removed". An author reading that headline will wonder what happened to 5 lines — it actively undermines trust in the diff UI.

Fix: compute `cellCount` when the table diff is built, then use it in `stats` for table segments:

```js
// In segments computed — when tableHtml is truthy, count changed cells:
const tmp = document.createElement('div')
tmp.innerHTML = tableHtml
const cellCount = tmp.querySelectorAll(
  'td.tc-cell-change, .tc-row-add > td, .tc-row-del > td'
).length
paired.push({ type: 'change', oldText: oldT, newText: newT,
              html: tableHtml, isTable: true, cellCount })

// In stats computed, line 320:
else if (s.type === 'change') {
  if (s.isTable) {
    added += s.cellCount  // cells changed, not lines
  } else {
    removed += s.oldText.split('\n').length
    added   += s.newText.split('\n').length
  }
}
```

The 2-cell edit then reads "+2 added" instead of "+1 added -5 removed". The UI label can optionally be changed to "cells" when the segment is a table, but the count alone is already accurate.

---

## Nits

### `useWordDiff.js:40` — `wordDiffPlain` exported but never imported

`wordDiffPlain` appears exactly once across all of `web/src` — its own definition. `useTableDiff.js` has a private `cellWordDiff` with an identical body; the DRY intent of the `useWordDiff` extraction is half-done.

```js
// useTableDiff.js — swap the import:
import { wordDiffPlain } from './useWordDiff'

// Replace the cellWordDiff call at line 104:
nc.innerHTML = wordDiffPlain(oldCellText, newCellText)
// Delete the private cellWordDiff function.
```

### `useWordDiff.js:17` — `Uint16Array` DP wraps silently at 65536 tokens

The LCS DP table is `Uint16Array`, so any LCS length ≥ 65536 wraps to a smaller value and the backtrack reconstructs the wrong edit sequence — wrong diff output, no error thrown. This was pre-existing in `TrackChanges.vue`; extracting it to a shared module makes it worth closing before new callers arrive.

Practical threshold is ~500 KB of single-cell content, so it's unreachable in document reviews today. A guard makes the failure mode explicit:

```js
export function diffTokens(oldToks, newToks) {
  if (oldToks.length > 65535 || newToks.length > 65535) {
    return [
      ...(oldToks.length ? [{ type: 'remove', text: oldToks.join('') }] : []),
      ...(newToks.length ? [{ type: 'add',    text: newToks.join('') }] : []),
    ]
  }
  // ... existing DP ...
}
```

Same guard applies to `diffLines`.

unidoc-ahall:
Thanks @unidoc-alip — all four, and the must-fix repro was spot on:

1. **Table mis-pairing on deletion (must-fix)** — replaced the sequential `tIdx` counter with a real edit script: `diffBlocks` now records consecutive `del→ins` pairs and `rightBlocks` diffs each changed table against the block it actually replaced. `changedOldTables` and `tIdx` are gone. Added an E2E regression for your exact case (delete Team Roster + change one Interested Parties cell → **exactly 1** `tc-cell-change`, Alice/Bob no longer mis-paired in).
2. **Line-count stats (should-fix)** — table change segments now carry a `cellCount` (counted from the merged diff) and the stats use it, so a 2-cell md→HTML edit reads **'+2 added'** instead of '+1 -5'. E2E now asserts it.
3. **wordDiffPlain unused (nit)** — `useTableDiff` imports and uses it; the duplicate `cellWordDiff` is deleted.
4. **Uint16Array overflow (nit)** — `diffTokens`/`diffLines` guard against >65535 and fall back to a whole remove+add.

Both E2E cases green twice against the stack (idempotent); web build green.

unidoc-alip (approved):
looks good to me

PR #100, opened 2026-06-26
feat(auth): Cloudflare Access web session + JIT provisioning (#98)

### LINKED ISSUE #98: JIT user provisioning on first Cloudflare Access login
## Summary

The Cloudflare Access auth middleware verifies the upstream JWT correctly but refuses to log in any user that does not already exist in `db.users`. For self-hosted deployments that put ISMS behind Cloudflare Zero Trust with Entra (or any other IdP) connected at the CF Access layer, this means **every team member has to be manually pre-created by an admin before they can log in for the first time** — even though CF Access has already authenticated them against the org's identity provider and admitted them through the team policy.

This is friction that defeats the point of using CF Access as the single source of truth for identity. JIT (just-in-time) provisioning on first successful CF Access login is the expected behaviour for this deployment pattern.

## Current behaviour

`internal/isms/api/middleware.go:222-253` — after verifying `Cf-Access-Jwt-Assertion` against the CF JWKS, the middleware does:

```go
user, err := cfg.DB.GetUserByEmail(ctx, email)
if err != nil {
    return echo.NewHTTPError(http.StatusUnauthorized, "user not found")
}
```

If the user doesn't exist, the request is rejected with 401 and the admin has to run `isms server user create --email ...` manually for every team member.

The `cfClaims` struct (line 378) only parses `email`, so even if we wanted to auto-create, we couldn't populate the user's display name or group membership from the JWT.

### 1. Extend `cfClaims` to capture more JWT claims

```go
type cfClaims struct {
    Email  string   `json:"email"`
    Name   string   `json:"name,omitempty"`
    Groups []string `json:"groups,omitempty"`  // populated when CF Access is configured to pass identity claims
    Exp    int64    `json:"exp"`
    Iat    int64    `json:"iat"`
    Iss    string   `json:"iss"`
    Aud    json.RawMessage `json:"aud"`
}
```

### 2. Env-var-gated JIT provisioning in the middleware

```go
user, err := cfg.DB.GetUserByEmail(ctx, email)
if err != nil {
    if !cfg.AutoProvisionCFUsers {
        return echo.NewHTTPError(http.StatusUnauthorized, "user not found")
    }
    user, err = cfg.DB.CreateUser(ctx, db.NewUserParams{
        Email: email,
        Name:  cmp.Or(claims.Name, deriveNameFromEmail(email)),
    })
    if err != nil {
        return echo.NewHTTPError(http.StatusInternalServerError, "auto-provision failed: " + err.Error())
    }

    if cfg.AutoProvisionDefaultOrg != "" {
        org, oerr := cfg.DB.GetOrganizationBySlug(ctx, cfg.AutoProvisionDefaultOrg)
        if oerr == nil {
            cfg.DB.AddOrgMember(ctx, org.ID, user.ID, cfg.AutoProvisionDefaultRole)
        }
    }

    cfg.DB.LogActivity(ctx, "user_auto_provisioned_cf_access", email, /* metadata */)
}
```

### 3. New environment variables

| Var | Default | Purpose |
|---|---|---|
| `ISMS_CF_AUTO_PROVISION` | `false` | Enable JIT user creation on first CF Access login |
| `ISMS_CF_DEFAULT_ORG` | _(unset)_ | Org slug to add new users to |
| `ISMS_CF_DEFAULT_ROLE` | `reader` | Role granted on auto-add — `reader`, `contributor`, `manager`, `admin` |

Default `false` keeps the current behaviour for existing deployments — this is opt-in.

### 4. Audit trail

Every auto-provisioned user is logged in the activity table with source `cf_access` so admins can review who has been granted access without explicit invitation.

## Security tradeoff

JIT provisioning is **safe only when the CF Access policy is the source of truth for who should have access**. If CF Access is misconfigured (policy disabled on a route, JWT leaked, etc.), an unintended user could auto-provision themselves.

Mitigations baked into the design:

- Opt-in via env var (default off).
- Default role is `reader` — least privilege; admins must promote.
- Audit trail records every auto-provision event with source `cf_access`.
- JWT verification is unchanged — only valid, audience-matching JWTs trigger the path.

Operator guide should document this tradeoff prominently in the CF Access section.

## Test plan

- [ ] Existing user, CF JWT valid → unchanged behaviour, user found
- [ ] New user, CF JWT valid, `ISMS_CF_AUTO_PROVISION=false` → 401 (current behaviour)
- [ ] New user, CF JWT valid, `ISMS_CF_AUTO_PROVISION=true`, no default org → user created, no org membership, 401 on next request (org resolution fails)
- [ ] New user, CF JWT valid, auto-provision + `ISMS_CF_DEFAULT_ORG=biggi` + `ISMS_CF_DEFAULT_ROLE=reader` → user created, added to org as reader, request succeeds
- [ ] New user, CF JWT valid, auto-provision + invalid default org slug → user created without org, error logged
- [ ] CF JWT invalid → 401 (current behaviour, no provision)
- [ ] CF headers missing → 401 (current behaviour, no provision)
- [ ] Activity log records `user_auto_provisioned_cf_access` event

## Documentation

- Update `docs/operator-guide.md` Cloudflare Access section with the three new env vars and the security tradeoff
- Update `docs/architecture.md` if it covers auth flows
- Add a one-line CHANGELOG entry under `[0.7.0]` once shipped

## Out of scope (for follow-ups)

- Group-based role assignment (read `claims.Groups` and map to ISMS roles) — useful but adds complexity around group-to-role mapping config; track separately
- Auto-deprovision (disable user in ISMS when CF Access removes them from team) — requires CF Access webhook or scheduled sync; separate issue
- Multi-tenant routing for CF Access (different default orgs per subdomain) — single-tenant covers the common deployment

## References

- Discussion that surfaced this gap: chat thread 2026-06-24 around `isms.biggi.net` deployment
- Related files: `internal/isms/api/middleware.go`, `internal/isms/api/server.go`, `internal/isms/db/users.go`
- Related env vars already in place: `CLOUDFLARE_TEAM_DOMAIN`, `ISMS_CF_AUDIENCE`

unidoc-ahall:
Closes #98.

**Problem.** CF Access authenticated the API/CLI/git layer, but the web SPA gated on a local token and bounced to /login — so passing CF Access (Entra at the CF layer) did **not** log you into the web app, and nothing minted an ISMS session from a CF identity. No ISMS-side OIDC wanted: the org IdP is already wired through CF Zero Trust.

**Change.**
- `GET /auth/cf-session` — validates the `Cf-Access-Jwt-Assertion` (reusing the JWKS cache; **not** just the email header), resolves/(JIT-)provisions the user, resolves org+role, mints a session like a password login. Public route; fails safe (401) when CF Access isn't configured.
- The auth middleware shares the same resolve-or-provision helper, so API/CLI/git JIT-provision too.
- The SPA probes `/auth/cf-session` once when it has no token, before redirecting to /login.

**JIT is deliberately minimal** (per discussion): `ISMS_CF_AUTO_PROVISION=1` creates the **user row only** — no default org/role. Which org someone joins, and at what role, stays an explicit `isms server org add-member` / Admin decision, not an env-var guess. A provisioned user with no org gets a 403 from cf-session (rather than a stale org-less token), so `add-member` + refresh logs them in cleanly.

**Security.** Access JWT always cryptographically verified; JIT opt-in (default off); a provisioned user sees nothing until an admin grants org access.

**Tests.** `api_cf_test.go` (name derivation, CI go-unit); `tests/test_cf_session.py` (public route + fails-safe 401, CI integration); full suite green (664). The CF-JWT happy path needs a live CF Access tunnel (JWKS + Cloudflare-signed JWT), so it's verified manually against the real deployment — documented gap, same class as the SMTP paths. **No DB migration.** Operator guide updated.

### DIFF
--- a/docs/operator-guide.md
+++ b/docs/operator-guide.md
@@ -188,6 +188,7 @@ All configuration is via environment variables. See `contrib/unidoc.env` for the
 |----------|-------------|
 | `CLOUDFLARE_TEAM_DOMAIN` | Cloudflare Access team domain (e.g. `mycompany.cloudflareaccess.com`). Enables CF Zero Trust authentication. |
 | `ISMS_CF_AUDIENCE` | CF Access Application Audience (AUD) tag. **Set this when using CF Access** — without it, any CF Access JWT from any application is accepted. |
+| `ISMS_CF_AUTO_PROVISION` | Set to `1`/`true` to create the **user record** automatically on their first Cloudflare Access login (JIT). Off by default. The user is created with no org membership — an admin (or the CLI) then adds them to the right organization and role. **Only safe when your CF Access policy is the source of truth for who may reach ISMS** — see the warning below. |
 | `ISMS_USER_SIGNUP` | Set to `1` to enable self-registration. Off by default. |
 | `ISMS_SKIP_EMAIL_VERIFY` | Set to `1` to skip email verification on signup (users are active immediately). For dev/eval only. |
 
@@ -582,6 +583,26 @@ Disable with `ISMS_RATE_LIMIT=0`. The default is 20 requests per minute per IP.
 ### CF Access warning: "ISMS_CF_AUDIENCE not set"
 
 When `CLOUDFLARE_TEAM_DOMAIN` is set but `ISMS_CF_AUDIENCE` is not, the server logs a warning on startup. Without the audience check, any CF Access JWT from any application on your team domain is accepted. Set `ISMS_CF_AUDIENCE` to the Application Audience (AUD) tag from your CF Access dashboard.
+
+### Cloudflare Access as web login (Zero Trust SSO)
+
+With `CLOUDFLARE_TEAM_DOMAIN` + `ISMS_CF_AUDIENCE` set and ISMS published behind a CF Access application, a user who has already passed CF Access (e.g. via Entra at the CF layer) is logged into the **web app** automatically — no separate ISMS login. On load the SPA calls `GET /api/v1/auth/cf-session`; behind CF Access the proxy adds the identity headers, the server **validates the Access JWT** (not just the email header), resolves the user, and mints an ISMS session. The same CF identity also authenticates the API, CLI, and git.
+
+By default the user must already exist in ISMS and be a member of the org. To skip the manual "create the user" step, enable JIT provisioning:
+
+```
+ISMS_CF_AUTO_PROVISION=1
+```
+
+With this on, the first CF Access login **creates the user record** (no org, no role). They can authenticate, but can't load an org's data until an **admin adds them to the right organization** — Admin → Members, or:
+
+```
+isms server org add-member --org acme --email user@acme.com --role reader
+```
+
+There is intentionally **no default org/role**: which org someone belongs to, and at what role, is a deliberate decision, not an env-var guess.
+
+**Security tradeoff — read before enabling JIT.** Auto-provisioning is safe **only when your CF Access policy is the source of truth for who may reach ISMS.** If a route is left unprotected, or a CF Access token leaks, an unintended user could create an account. Mitigations: it's opt-in (default off); a provisioned user has **no org access at all** until an admin grants it (so a stray account can see nothing); and the Access JWT is always cryptographically verified — a missing/invalid token never provisions.
 
 ### Git repo not found for organization
 
--- a/internal/isms/api/api_cf.go
+++ b/internal/isms/api/api_cf.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
@@ -10,6 +11,11 @@ import (
 	"github.com/labstack/echo/v4"
 	"isms.sh/internal/isms/db"
 )
+
+// errProvisionFailed distinguishes a failed JIT user creation (DB error → 500)
+// from a genuine "user not found / provisioning off" (401), so an operator
+// debugging a broken DB doesn't chase a misleading 401 down the CF JWT path.
+var errProvisionFailed = errors.New("auto-provision failed")
 
 // cfProvisionConfig controls just-in-time user provisioning on Cloudflare Access
 // login. Off by default — opt in via ISMS_CF_AUTO_PROVISION. Safe ONLY when the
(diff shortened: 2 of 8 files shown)


unidoc-alip (changes requested):
Solid implementation of the CF Access web session + JIT provisioning. The design decisions are sound: the `resolveCFUser` helper shared by both the middleware and the web handler eliminates drift, the "user row only" JIT scope avoids policy-in-env-vars, and the 403-on-no-org behaviour correctly prevents a stale org-less token from persisting. Two items worth addressing before merge.

---

## Findings

### should-fix — provision failure masked as 401 "user not found"

**`internal/isms/api/api_cf.go:98`** (also `internal/isms/api/middleware.go:255`)

When `ISMS_CF_AUTO_PROVISION` is on and `UpsertUser` fails (DB down, constraint violation, etc.), `resolveCFUser` returns the DB error — but both call sites map **any** non-nil error to `401 "user not found"`. A provision failure is not a "user not found"; it should be `500`. An operator debugging a broken DB connection will see a misleading 401 and spend time on the CF JWT path instead.

Suggested fix — sentinel error in `resolveCFUser`:

```go
// resolveCFUser, UpsertUser branch:
return nil, false, fmt.Errorf("%w: %v", errProvisionFailed, err)

// handleCFSession (and the middleware path):
if err != nil {
    if errors.Is(err, errProvisionFailed) {
        return echo.NewHTTPError(http.StatusInternalServerError, "auto-provision failed")
    }
    return echo.NewHTTPError(http.StatusUnauthorized, "user not found")
}
```

---

### nit — email mismatch check is case-sensitive in middleware, case-insensitive in handler

**`internal/isms/api/middleware.go:245`**

The mismatch check here uses `!=` (case-sensitive), but the equivalent check in `handleCFSession` (`api_cf.go:84`) uses `strings.EqualFold`. CF Access normalises emails to lowercase in practice so this won't bite, but the inconsistency is a latent surprise. Consider aligning:

```go
if claims.Email != "" && !strings.EqualFold(claims.Email, email) {
```

unidoc-ahall:
Thanks @unidoc-alip — both done:

1. **Provision failure masked as 401 (should-fix)** — added an `errProvisionFailed` sentinel; `resolveCFUser` wraps a failed `UpsertUser` with it, and **both** call sites (`handleCFSession` and the middleware) map it to `500 'auto-provision failed'`. A genuine missing user / provisioning-off still returns 401. An operator with a broken DB now sees a 500, not a misleading 401.
2. **Case-sensitive email mismatch (nit)** — middleware now uses `strings.EqualFold`, matching `handleCFSession`.

go build + vet + unit test green.

unidoc-alip (approved):
looks good to me

Issue #105, opened 2026-06-26
Harden git push: protect server-managed review refs + forbid history rewrites


unidoc-ahall:
Pushes (manager/admin) are validated for content but not at ref level. Reject pushes that create/modify/delete refs/heads/review/* or refs/reviews/* (server-managed review history), and reject non-fast-forward pushes and ref deletions on any ref (no history rewrite). Revert the ref state on violation.

Related (declutter, optional): archive merged review refs heads/review/<id> -> refs/reviews/<id> so 'git branch' stays clean — history is already retained + protected by this change, so cosmetic only.

Issue #104, opened 2026-06-26
Notification links use the apex URL instead of the tenant URL


unidoc-ahall:
A Slack notification linked to https://isms.sh/… (apex) instead of the tenant (https://unidoc.isms.sh/…). Build links from orgURLs (per-org base), like #16's email links.

Issue #103, opened 2026-06-26
Review Split diff: synchronise scrolling between the two panes


unidoc-ahall:
The Split review panes scrolled independently. Scroll one, the other follows proportionally (GitHub-style).

Issue #102, opened 2026-06-26
Review diff: empty Previous / whole-doc-new when the document was never approved


unidoc-ahall:
A reviewer's proposal showed an empty 'Previous' pane (whole doc as 'new') because the diff baseline used the last approved commit, which a never-approved doc lacks. Use sent_head whenever a review/<id> branch exists.

PR #106, opened 2026-06-26
feat(review): complete #6 per-event diff + dry-run fixes + push ref-protection

unidoc-ahall:
Completes **#6** and bundles the fixes found while dry-running the review workflow (review #2 on iso27001-4-2). No DB migration.

### #6 — per-event diff (headline)
Each `proposed_revision` in the Conversation timeline expands (**View changes**) to show exactly what that revision changed — the proposal commit vs its parent (`commit^..commit`), anchored by `decision_log.commit_ref` so it survives merge. Backend: store `RefHash` + a `to` diff param; frontend reuses `TrackChanges`. (Per-cell table diff = slice 1, merged in #99.)

### Dry-run fixes
- **Diff baseline** (#102): never-approved docs showed an empty 'Previous'; now use `sent_head` when a review branch exists.
- **Synced split scroll** (#103): the two panes scroll together.
- **Tenant notification links** (#104): links use `orgURLs`, not the apex.

### Security
- **Push ref-protection** (#105): `review/*` and `refs/reviews/*` are server-managed (immutable via push); all refs are fast-forward-only and undeletable — a push can't rewrite or erase review/approved history. Violations revert + reject.

### Tests
5 review E2E (per-cell, deletion-pairing, never-approved baseline, per-event expand, synced scroll) — green twice against the stack, run in CI. Go unit tests for the push ref-policy (`reviewRefViolation`) and name derivation. The git-push rejection + Slack send need a live client/webhook — verified manually (documented blind-spot, as with SMTP).

Larger PR by request (everything from the dry-run in one). The push ref-protection is a distinct security control — worth a focused pass.

### DIFF
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -510,6 +510,11 @@ func (s *Server) handleReviewTimeline(c echo.Context) error {
 }
 
 // handleReviewDiff returns the diff for a review (between commit_hash stored on review and current HEAD).
+// reSHA matches a full lowercase git commit SHA — the only accepted form for the
+// per-event diff `to` param (always a decision_log commit_ref), so arbitrary
+// revision strings (HEAD~1, branch names) can't reach the diff machinery.
+var reSHA = regexp.MustCompile(`^[0-9a-f]{40}$`)
+
 func (s *Server) handleReviewDiff(c echo.Context) error {
 	orgID := getOrgID(c)
 	id, err := strconv.Atoi(c.Param("id"))
@@ -554,10 +559,18 @@ func (s *Server) handleReviewDiff(c echo.Context) error {
 	// author can see exactly what one revision changed. Defaults below to the
 	// whole-round view.
 	to := c.QueryParam("to")
+	if to != "" && !reSHA.MatchString(to) {
+		return echo.NewHTTPError(http.StatusBadRequest, "to must be a full commit SHA")
+	}
 	from := c.QueryParam("from")
 	if from == "" {
 		if to != "" {
-			from = to + "^" // per-event: the proposal vs the state it was based on
+			// Per-event: the proposal vs its parent. A root commit has no parent —
+			// leave the baseline empty (the whole document is genuinely new) rather
+			// than relying on a silently-swallowed resolve failure.
+			if _, perr := st.RefHash(to + "^"); perr == nil {
+				from = to + "^"
+			}
 		} else if review.SentHead != "" && (hasBranch || review.CommitHash != "") {
 			from = review.SentHead
 		} else if review.CommitHash != "" {
@@ -567,11 +580,19 @@ func (s *Server) handleReviewDiff(c echo.Context) error {
 
 	if to != "" {
 		// Per-event diff between two explicit commits.
-		diffText, _ = st.DiffDocumentBodies(from, to, filePath)
+		var dErr error
+		diffText, dErr = st.DiffDocumentBodies(from, to, filePath)
+		if dErr != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, "per-event diff failed: "+dErr.Error())
+		}
 	} else if hasBranch && from != "" {
 		// Diff what was sent (from) against the reviewer's proposed branch, so
 		// "Previous" is the sent version and "Current" is the proposal.
-		diffText, _ = st.DiffDocumentBodies(from, branchName, filePath)
+		var dErr error
+		diffText, dErr = st.DiffDocumentBodies(from, branchName, filePath)
+		if dErr != nil {
+			return echo.NewHTTPError(http.StatusInternalServerError, "failed to compute branch diff: "+dErr.Error())
+		}
 	} else if hasBranch {
 		// A proposal with no baseline at all — branch vs main.
 		diffText = branchDiff
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -41,6 +41,11 @@ func (s *Server) logAndNotify(ctx context.Context, orgID int, a *db.Activity) {
 				Action: a.Action,
 				Detail: a.Detail,
 				Link:   docLink(a.DocumentID),
+				// Org-scoped base so the link lands on the tenant's own URL
+				// (e.g. unidoc.isms.sh / a custom domain) via orgURLs, not the
+				// bare apex (isms.sh). Empty AppURL falls back to the notifier's
+				// configured BaseURL.
+				BaseURL: s.orgMail(ctx, orgID).AppURL,
 				Channels: notify.OrgChannels{
 					SlackWebhook: slackWH,
 					MatrixRoomID: matrixRoom,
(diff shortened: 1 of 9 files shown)


unidoc-alip (changes requested):
Solid PR — the per-event diff, push ref-protection, and dry-run fixes all look correct and well-tested. A few things worth tightening before merge: the `from` ref resolution failure in `DiffDocumentBodies` is silently swallowed (shows a misleading "everything is new" diff for edge cases like a root commit), the ref restore on a push violation ignores errors (silent partial restore is hard to diagnose), and the `to` query param accepts arbitrary git revision strings rather than just commit SHAs. None are blockers for the main feature path, but they affect observability and security posture.

---

## Findings

| # | Severity | File | Line | Summary |
|---|---|---|---|---|
| 1 | should-fix | `internal/isms/api/api_collab.go` | 77 | `from` ref failure silently swallowed — wrong diff on root commit |
| 2 | should-fix | `internal/isms/api/api_git.go` | 162 | `restoreRefs` discards all errors — partial restore is undetectable |
| 3 | should-fix | `internal/isms/api/api_collab.go` | 63 | `to` param accepts arbitrary revision strings, not just commit SHAs |
| 4 | nit | `internal/isms/api/api_collab.go` | 77 | Error from `DiffDocumentBodies` discarded inconsistently across branches |
| 5 | nit | `tests/test_e2e_review_diff.py` | 486 | Brittle exact-count assertion on changed cells |

---

### Finding 1 — should-fix · `api_collab.go:77`

**`from` ref failure is silently swallowed — root-commit proposal shows the whole document as new**

When `to + "^"` can't be resolved (the proposal commit is the repo's initial commit — no parent), `DiffDocumentBodies` silently ignores the `ReadFileAtRef` failure and produces a diff where the whole document appears as added. The call site discards the error too, so no 500 is returned. The author sees a confusing "everything added" view instead of just the reviewer's two-cell edit.

**Verification**

Two layers of silent swallow confirmed from source:

Layer 1 — `store/git.go:702` (`DiffDocumentBodies`):
```go
if fromRef != "" {
    if data, err := s.ReadFileAtRef(fromRef, filePath); err == nil {
        fromContent = StripFrontmatter(string(data))
    }
    // error is silently dropped; fromContent stays ""
}
```

Layer 2 — `api_collab.go:77` (call site):
```go
diffText, _ = st.DiffDocumentBodies(from, to, filePath)
```

`resolveRef` uses go-git v5.6.1's `ResolveRevision` which parses `sha^` via its `CaretPath{1}` token (`internal/revision/parser.go:357-388`), so `sha^` resolves correctly for any non-root commit. For the root commit (no parent), `ResolveRevision` fails, `ReadFileAtRef` propagates it, and the `if err == nil` guard swallows it silently.

**Proof of concept**

1. Create a fresh org — the scaffold creates the repo's first (and only) commit.
2. Create one document — this commit IS the root commit (no parent).
3. Send for review; reviewer proposes a revision.
4. `approve` handler stores `commit_ref = sha_of_root_commit`.
5. Request `GET /reviews/{id}/diff?to=<sha_of_root_commit>`.
6. Server computes `from = sha_of_root_commit + "^"`.
7. `resolveRef` fails (no parent) → `ReadFileAtRef` error dropped → `fromContent = ""`.
8. `DiffDocumentBodies` returns a diff showing the whole document as added.
9. Response: `old_body = ""`, `new_body = full document`.
10. "View changes" shows the entire document highlighted as new instead of the two-cell reviewer edit.

**Suggested fix**

Guard the parent resolution before using it as `from`:

```go
if to != "" {
    parentRef := to + "^"
    if _, resolveErr := st.RefHash(parentRef); resolveErr != nil {
        parentRef = "" // root commit — no baseline available; show full doc as new
    }
    from = parentRef
}
```

Combined with fixing Finding 4 (handle the `DiffDocumentBodies` error explicitly), this covers both the misleading diff and the silent failure.

---

### Finding 2 — should-fix · `api_git.go:162`

**`restoreRefs` discards all errors — a failed restore is indistinguishable from a successful one**

`restoreRefs` is the recovery path for the new push ref-protection security control. If `SetRefHash` or `DeleteRef` fails (disk full, concurrent write, stale pack-refs), the repo is left in a partially-restored state with no log entry. The 403 is still returned to the pusher, but the server has no way to know the rollback didn't complete.

**Verification**

```go
// api_git.go — restoreRefs, added in this PR
func restoreRefs(st *store.Store, before, after map[string]string) {
    for name, hash := range before {
        _ = st.SetRefHash(name, hash)  // error explicitly discarded
    }
    for name := range after {
        if _, ok := before[name]; !ok {
            _ = st.DeleteRef(name)     // error explicitly discarded
        }
    }
}
```

The caller logs "REJECTED" before calling `restoreRefs` but nothing after. A partial restore failure is completely silent in both logs and metrics.

**Proof of concept**

Concurrent push race:

1. Org refs: `refs/heads/master=A`, `refs/heads/review/1=R1`.
2. Pusher 1 sends a push changing `refs/heads/review/1` — caught as a violation.
3. Concurrently, pusher 2 legitimately advances `refs/heads/master` to `B` (same cached `*store.Store`, different goroutine).
4. `restoreRefs` calls `SetRefHash("refs/heads/master", "A")` — silently overwrites pusher 2's legitimate commit `B` with the stale snapshot.
5. `_ = st.SetRefHash(...)` makes a correct restore and a corrupting overwrite indistinguishable.

**Suggested fix**

Log each restore failure; the existing `fmt.Fprintf(os.Stderr, ...)` pattern matches the codebase style:

```go
func restoreRefs(st *store.Store, before, after map[string]string) {
    for name, hash := range before {
        if err := st.SetRefHash(name, hash); err != nil {
            fmt.Fprintf(os.Stderr, "restoreRefs: could not restore %s to %.8s: %v\n",
                name, hash, err)
        }
    }
    for name := range after {
        if _, ok := before[name]; !ok {
            if err := st.DeleteRef(name); err != nil {
                fmt.Fprintf(os.Stderr, "restoreRefs: could not delete new ref %s: %v\n",
                    name, err)
            }
        }
    }
}
```

---

### Finding 3 — should-fix · `api_collab.go:63`

**`to` query param accepts arbitrary revision strings — not just proposal commit SHAs**

The `to` param is read with no validation. `resolveRef` accepts branch names, full ref paths, and revision syntax (`HEAD~1`, `refs/heads/master`, etc.) — not only raw commit SHAs. An authenticated user can pass `?to=refs/heads/master` to diff any recent master commit against its parent, for the review's document, without that commit having any connection to the review.

`filePath` is correctly locked to the review's document (fetched from the DB by `review.DocumentID`, not from the request), so there is no file traversal. The scope is limited to reading arbitrary historical versions of one document the user is already authorized to review — but it bypasses the intent that `to` should only ever be a proposal `commit_ref` recorded in `decision_log`.

**Verification**

```go
to := c.QueryParam("to")   // api_collab.go — no validation
```

go-git v5.6.1's `TestResolveRevision` confirms `ResolveRevision` accepts:
`"heads/master"`, `"refs/heads/master"`, `"refs/remotes/origin/master"`, branch names, tags, and short SHAs. All reach `DiffDocumentBodies` unchanged.

The legitimate caller (`Reviews.vue`) always passes `entry.data.commit_ref`, which is always a 40-character hex SHA stored by the server. Arbitrary revision strings are an unintended input path.

**Proof of concept**

```
GET /reviews/42/diff?to=refs/heads/master
```

- `to = "refs/heads/master"` passes through unvalidated.
- `from = "refs/heads/master^"` (the parent of master tip).
- `DiffDocumentBodies("refs/heads/master^", "refs/heads/master", filePath)` returns the diff of master's most recent commit for this document.
- If master was just advanced by a merge unrelated to this review, the response exposes what changed in that commit for the review's document — outside the review's scope.

**Suggested fix**

Add a SHA format guard immediately after reading the param:

```go
to := c.QueryParam("to")
if to != "" {
    for _, c := range to {
        if !((c >= '0' && c <= '9') || (c >= 'a' && c <= 'f')) {
            return echo.NewHTTPError(http.StatusBadRequest,
                "to must be a full lowercase hex commit SHA")
        }
    }
    if len(to) != 40 {
        return echo.NewHTTPError(http.StatusBadRequest,
            "to must be a full 40-character commit SHA")
    }
}
```

Or with a pre-compiled regexp at package level:

```go
var reSHA = regexp.MustCompile(`^[0-9a-f]{40}$`)

// in handler:
if to != "" && !reSHA.MatchString(to) {
    return echo.NewHTTPError(http.StatusBadRequest, "to must be a full commit SHA")
}
```

---

### Finding 4 — nit · `api_collab.go:77`

**`DiffDocumentBodies` errors discarded in new branches, handled in the original branch**

The new `to != ""` branch and the pre-existing `hasBranch && from != ""` branch both discard errors with `_`. Only the original `else` branch handles them with a 500. This means a per-event diff failure returns HTTP 200 with empty `old_body`/`new_body` panes, while the same failure on the HEAD-diff path returns HTTP 500.

**Verification**

```go
if to != "" {
    diffText, _ = st.DiffDocumentBodies(from, to, filePath)        // silent
} else if hasBranch && from != "" {
    diffText, _ = st.DiffDocumentBodies(from, branchName, filePath) // silent (pre-existing)
} else {
    diffText, err = st.DiffDocumentBodies(from, "HEAD", filePath)
    if err != nil {
        return echo.NewHTTPError(http.StatusInternalServerError,    // handled
            "failed to compute diff: "+err.Error())
    }
}
```

**Proof of concept**

If the document file is absent at the proposal commit (renamed or moved):

```
GET /reviews/{id}/diff?to=<commit_ref>
→ HTTP 200, {"old_body": "", "new_body": "", "lines": []}
   UI shows empty "View changes" panel with no error

Same condition via the else branch:
→ HTTP 500, "failed to compute diff: reading file at to ref: ..."
```

**Suggested fix**

```go
if to != "" {
    var dErr error
    diffText, dErr = st.DiffDocumentBodies(from, to, filePath)
    if dErr != nil {
        return echo.NewHTTPError(http.StatusInternalServerError,
            "per-event diff failed: "+dErr.Error())
    }
} else if hasBranch && from != "" {
    var dErr error
    diffText, dErr = st.DiffDocumentBodies(from, branchName, filePath)
    if dErr != nil {
        return echo.NewHTTPError(http.StatusInternalServerError,
            "failed to compute branch diff: "+dErr.Error())
    }
}
```

---

### Finding 5 — nit · `test_e2e_review_diff.py:486`

**Brittle exact-count assertion on changed cells**

`count() == 2` ties the test to the exact number of `td.tc-cell-change` elements emitted by the current `TrackChanges` renderer. If the component is refactored to emit one element per word-change rather than per cell, the count could become 4 for the same test fixture while the feature itself works correctly.

**Verification**

```python
# test_e2e_review_diff.py:486-488
assert page.locator("td.tc-cell-change").count() == 2        # exact count — brittle
assert page.locator(".tc-cell-change .tc-word-ins",
       has_text="Background check.").count() >= 1            # content check — robust
```

The adjacent word-ins assertion already uses `>= 1`. The count assertion is inconsistent and the weaker check.

**Suggested fix**

Replace exact count with content-based assertions for both edited cells:

```python
assert page.locator("td.tc-cell-change").count() >= 1, \
    "per-event diff should show at least one changed cell"
assert page.locator(".tc-cell-change .tc-word-ins",
       has_text="Background check.").count() >= 1
assert page.locator(".tc-cell-change .tc-word-ins",
       has_text="Review the suppliers").count() >= 1
```

---

## Findings — Retracted

**Finding 5 (original) — `api_git.go:160` — Double `storeForOrg` acquisition**

Retracted after verification. `storeForOrg` uses a `sync.Map` keyed by `orgID` and returns the cached `*store.Store` after the first call. All three `storeForOrg` calls in the receive-pack path return the same pointer with O(1) overhead. No change needed.

---

## Issue #6 Acceptance Criteria Check

| Criterion | Status | Notes |
|---|---|---|
| Per-event expandable diff in timeline | PASS | "View changes" button per `proposed_revision` entry; cached by `commit_ref` |
| Word-level highlights in changed table cells | PASS | Reuses `TrackChanges` from #99 (already merged) |
| Full diff view still accessible | PASS | Changes tab unaffected |
| Step through revisions one at a time | PARTIAL | Click each entry individually; no prev/next navigation — AC is functionally met but UX is click-per-entry rather than a stepper |

unidoc-ahall:
Thanks @unidoc-alip — all five:

1. **Root-commit baseline (should-fix)** — `to^` is now resolved up front via `RefHash`; if it doesn't resolve (root commit, no parent) the baseline is left empty deliberately (whole doc is genuinely new), not via a swallowed failure.
2. **restoreRefs swallows errors (should-fix)** — each `SetRefHash`/`DeleteRef` failure is now logged to stderr, so a partial rollback of the ref-protection is visible.
3. **`to` accepts arbitrary revisions (should-fix)** — validated against `^[0-9a-f]{40}$`; anything else → 400. Fixture now asserts `?to=refs/heads/master` is rejected.
4. **Inconsistent diff error handling (nit)** — the `to` and review-branch paths now return 500 on a `DiffDocumentBodies` error, matching the HEAD path.
5. **Brittle exact count (nit)** — the per-event test now asserts both inserted phrases (content-based) instead of `== 2`.

Re your retraction (double `storeForOrg`) — agreed, it's the cached pointer. And the 'step through revisions' AC: it's click-per-entry today (no prev/next stepper) — functionally met; happy to add a stepper as a follow-up if you'd like.

All 5 review E2E green twice against the stack; gofmt + go vet + unit green.

unidoc-alip (approved):
looks good to me

PR #109, opened 2026-07-02
fix(url): honour subdomainRouting in orgURLs + OIDC redirect (#108)

### LINKED ISSUE #108: orgURLs / OIDC redirect emit subdomain links even when subdomain routing is off
orgURLs (server.go) and orgTokenRedirectURL (api_oidc.go) chose subdomain-vs-path purely from the host shape (a dotted, non-localhost, non-IP host → subdomain) and never consulted ISMS_SUBDOMAIN_ROUTING. On a path-based deployment on a real domain (e.g. a single-tenant box at isms.stsplatform.com with routing OFF), they still generated subdomain URLs like https://sts.isms.stsplatform.com — which don't resolve (no wildcard). This hits notification/email links (#16/#104) and the post-OIDC redirect.

Fix: gate the subdomain branch on subdomainRouting in both functions, so a generated link always matches how requests are actually routed — never inferred from the host shape alone. Custom-domain and path-based behaviour unchanged. Covered by orgurls_test.go (adds a routing-off case). No DB migration.

unidoc-ahall:
Closes #108.

**Bug.** `orgURLs` (notification/email links) and `orgTokenRedirectURL` (post-OIDC redirect) decided subdomain-vs-path from the **host shape** (dotted, non-localhost, non-IP → subdomain) and never read `ISMS_SUBDOMAIN_ROUTING`. On a path-based single-tenant box on a real domain (`isms.stsplatform.com`, routing off) they generated subdomain links like `https://sts.isms.stsplatform.com` — which don't resolve (no wildcard). Surfaced by the STS migration; affects Slack/email links (#104/#16).

**Fix.** Both functions now gate the subdomain branch on `subdomainRouting`, so a generated URL always matches how requests are actually routed — the config is the single source of truth, not a guess from the hostname. Custom-domain and path-based behaviour unchanged.

**Test.** `orgurls_test.go` gains a routing-off case (real domain → path-based). go build + vet + `TestOrgURLs` green; gofmt clean. Backend-only, no DB migration.

### DIFF
--- a/internal/isms/api/api_oidc.go
+++ b/internal/isms/api/api_oidc.go
@@ -311,7 +311,7 @@ func (s *Server) handleOIDCCallback(c echo.Context) error {
 	// hop to https://<slug>.isms.sh/. Otherwise stay on the apex with #token=...
 	// and let the SPA route via /:org-prefixed paths.
 	baseURL := strings.TrimRight(os.Getenv("ISMS_BASE_URL"), "/")
-	redirectURL := orgTokenRedirectURL(baseURL, org.Slug, token, role)
+	redirectURL := orgTokenRedirectURL(baseURL, org.Slug, token, role, s.subdomainRouting)
 	return c.Redirect(http.StatusFound, redirectURL)
 }
 
@@ -319,7 +319,7 @@ func (s *Server) handleOIDCCallback(c echo.Context) error {
 // Prefers a subdomain hop (https://<slug>.<apex>/#token=...) when the base URL
 // has a hostname that can host subdomains; falls back to path-based routing
 // (<base>/<slug>/#token=...) for single-label / localhost hosts.
-func orgTokenRedirectURL(baseURL, slug, token, role string) string {
+func orgTokenRedirectURL(baseURL, slug, token, role string, subdomainRouting bool) string {
 	// Parse the base URL — extract scheme and host so we can splice in the slug.
 	// Cheap parse: assume baseURL like "https://isms.sh" or "http://localhost:9090".
 	const sep = "://"
@@ -337,9 +337,10 @@ func orgTokenRedirectURL(baseURL, slug, token, role string) string {
 		host = hostAndPort[:c]
 		port = hostAndPort[c:] // includes ':'
 	}
-	// Can this host serve subdomains? Need at least 2 dot-separated parts and
-	// not localhost / IP literal.
-	canSubdomain := strings.Contains(host, ".") &&
+	// Subdomain hop only when the deployment routes by subdomain; otherwise stay
+	// path-based so the redirect matches how requests are actually routed.
+	canSubdomain := subdomainRouting &&
+		strings.Contains(host, ".") &&
 		!strings.HasPrefix(host, "localhost") &&
 		!isIPLiteral(host)
 	if canSubdomain {
--- a/internal/isms/api/api_oidc.go
+++ b/internal/isms/api/api_oidc.go
@@ -311,43 +311,40 @@ func (s *Server) handleOIDCCallback(c echo.Context) error {
 	// hop to https://<slug>.isms.sh/. Otherwise stay on the apex with #token=...
 	// and let the SPA route via /:org-prefixed paths.
 	baseURL := strings.TrimRight(os.Getenv("ISMS_BASE_URL"), "/")
-	redirectURL := orgTokenRedirectURL(baseURL, org.Slug, token, role, s.subdomainRouting)
+	redirectURL := orgTokenRedirectURL(baseURL, org.Slug, org.Domain, token, role, s.subdomainRouting)
 	return c.Redirect(http.StatusFound, redirectURL)
 }
 
 // orgTokenRedirectURL builds the post-OIDC-login redirect for a given org.
-// Prefers a subdomain hop (https://<slug>.<apex>/#token=...) when the base URL
-// has a hostname that can host subdomains; falls back to path-based routing
-// (<base>/<slug>/#token=...) for single-label / localhost hosts.
-func orgTokenRedirectURL(baseURL, slug, token, role string, subdomainRouting bool) string {
-	// Parse the base URL — extract scheme and host so we can splice in the slug.
-	// Cheap parse: assume baseURL like "https://isms.sh" or "http://localhost:9090".
-	const sep = "://"
-	i := strings.Index(baseURL, sep)
-	if i < 0 {
+// Precedence mirrors orgURLs (server.go): a configured custom domain wins, then
+// a subdomain hop (https://<slug>.<apex>/login#token=...) when the deployment
+// routes by subdomain, otherwise path-based routing (<base>/<slug>/login#...).
+// Sharing resolveSubdomainHost keeps this decision identical to orgURLs, so the
+// login redirect always lands on the same host the org's other links use.
+func orgTokenRedirectURL(baseURL, slug string, domain *string, token, role string, subdomainRouting bool) string {
+	// Custom domain wins — an org with its own domain (e.g. audit.sts.is) must
+	// land its OIDC login there, not on the apex subdomain/path.
+	if domain != nil && *domain != "" {
+		d := *domain
+		if !strings.Contains(d, "://") {
+			scheme := "https"
+			if strings.HasPrefix(baseURL, "http://") {
+				scheme = "http"
+			}
+			d = scheme + "://" + d
+		}
+		d = strings.TrimRight(d, "/")
+		return fmt.Sprintf("%s/login#token=%s&role=%s", d, token, role)
+	}
+	scheme, host, port, canSubdomain := resolveSubdomainHost(baseURL, subdomainRouting)
+	if scheme == "" {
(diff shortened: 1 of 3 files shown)


unidoc-alip (changes requested):
The core fix is correct and verified at runtime: `orgURLs` and `orgTokenRedirectURL` now both gate the subdomain branch on `subdomainRouting`, matching the deployment's actual routing mode instead of guessing from host shape. Reproduced the original bug on the pre-fix commit (`sts.isms.stsplatform.com`) and confirmed the fix (`isms.stsplatform.com/sts`) via a live invite-email round trip.

Two things worth addressing before merge: `orgTokenRedirectURL` has zero test coverage despite carrying the identical new gating logic as `orgURLs` (only the latter gained test cases), and it never consults `org.Domain` — a custom-domain org's OIDC login redirects to the apex subdomain/path instead of their configured domain. Neither is newly introduced by this diff, but both sit in a function this PR's signature change directly touches, and the missing test would have made the second gap visible.

---

### 1. [should-fix] `internal/isms/api/api_oidc.go` line 322

`orgTokenRedirectURL` never consults `org.Domain`, unlike `orgURLs` (server.go:116) which returns the custom domain unconditionally before the subdomain/path gate. An org with a configured custom domain (e.g. `audit.sts.is`) logging in via OIDC gets redirected to `https://<slug>.<apex>/login#...` (or the path-based equivalent) instead of `https://audit.sts.is/login#...`.

Pre-existing gap, not introduced by this PR, but this function's signature is being touched here — worth fixing in the same pass or filing a follow-up issue so it doesn't get lost.

Proof — calling both sibling functions with the same custom-domain org and base URL shows the divergence directly:

```
orgURLs (custom domain, routing off):   app=https://audit.sts.is public=https://audit.sts.is
orgTokenRedirectURL (routing off):       https://isms.sh/sts/login#token=TOKEN&role=reader
orgTokenRedirectURL (routing on):        https://sts.isms.sh/login#token=TOKEN&role=reader
```

For the exact same org, notification emails correctly land the user on `https://audit.sts.is`, but the post-OIDC-login redirect never uses the custom domain. If the custom domain is the org's only publicly reachable/branded entry point, this is a broken or confusing login redirect for every custom-domain org using OIDC.

Suggested fix — thread `org.Domain` into `orgTokenRedirectURL` and check it first, mirroring `orgURLs`'s precedence (custom domain > subdomain > path):

```go
// api_oidc.go
redirectURL := orgTokenRedirectURL(baseURL, org.Slug, org.Domain, token, role, s.subdomainRouting)

// orgTokenRedirectURL builds the post-OIDC-login redirect for a given org.
// Precedence matches orgURLs: custom domain > subdomain (if enabled) > path.
func orgTokenRedirectURL(baseURL, slug string, domain *string, token, role string, subdomainRouting bool) string {
	if domain != nil && *domain != "" {
		d := *domain
		if !strings.Contains(d, "://") {
			d = "https://" + d
		}
		return fmt.Sprintf("%s/login#token=%s&role=%s", strings.TrimRight(d, "/"), token, role)
	}
	// ... existing scheme/host/port parsing + subdomainRouting gate unchanged ...
}
```

Update `orgurls_test.go` (or a new `api_oidc_test.go`, see finding #2) with a custom-domain case for `orgTokenRedirectURL` so this precedence is locked in by CI, not just this proof.

---

### 2. [should-fix] `internal/isms/api/api_oidc.go` line 322

`orgTokenRedirectURL` carries the same new `subdomainRouting`-gated logic as `orgURLs`, but only `orgurls_test.go` gained a routing-off test case — no test exercises `orgTokenRedirectURL` at all (`grep -rn "orgTokenRedirectURL" internal/isms/api/*_test.go` returns nothing). The PR description's "Covered by orgurls_test.go" claim doesn't extend to this function.

Coverage comparison for the two sibling functions that received the identical fix in this PR:

```
$ go test ./internal/isms/api/... -run TestOrgTokenRedirectURL -v
testing: warning: no tests to run

$ go tool cover -func=/tmp/cov.out | grep -i "orgTokenRedirectURL\|orgURLs"
internal/isms/api/api_oidc.go:322:  orgTokenRedirectURL   0.0%
internal/isms/api/server.go:115:    orgURLs               92.3%
```

Suggested fix — add a table-driven test for `orgTokenRedirectURL` alongside `TestOrgURLs`, mirroring its cases plus the custom-domain case from finding #1:

```go
func TestOrgTokenRedirectURL(t *testing.T) {
	cases := []struct {
		name             string
		base             string
		slug             string
		domain           *string
		subdomainRouting bool
		want             string
	}{
		{
			name:             "subdomain routing on: hop to subdomain",
			base:             "https://isms.sh",
			slug:             "sts",
			subdomainRouting: true,
			want:             "https://sts.isms.sh/login#token=TOK&role=reader",
		},
		{
			name:             "subdomain routing OFF on a real domain: stay path-based",
			base:             "https://isms.stsplatform.com",
			slug:             "sts",
			subdomainRouting: false,
			want:             "https://isms.stsplatform.com/sts/login#token=TOK&role=reader",
		},
		{
			name:             "localhost stays path-based regardless of routing flag",
			base:             "http://localhost:9090",
			slug:             "sts",
			subdomainRouting: true,
			want:             "http://localhost:9090/sts/login#token=TOK&role=reader",
		},
		// If finding #1's fix lands, add: custom domain wins regardless of routing mode.
	}
	for _, tc := range cases {
		t.Run(tc.name, func(t *testing.T) {
			got := orgTokenRedirectURL(tc.base, tc.slug, "TOK", "reader", tc.subdomainRouting)
			if got != tc.want {
				t.Errorf("got %q, want %q", got, tc.want)
			}
		})
	}
}
```

This is the minimal version (no `domain` param yet, matching the code as merged). If finding #1's fix is taken, add the `domain` field/case and update the call signature accordingly.

---

### 3. [nit] `internal/isms/api/server.go` line 144

`orgURLs` and `orgTokenRedirectURL` independently parse `scheme://host:port` and compute an identical three-condition `canSubdomain` predicate (`subdomainRouting && strings.Contains(host, ".") && !strings.HasPrefix(host, "localhost") && !isIPLiteral(host)`). This exact bug (host-shape heuristic ignoring `subdomainRouting`) existed in both copies and had to be patched twice in this PR — a direct symptom of the duplication.

Not blocking — the fix as shipped is correct for both call sites. `orgURLs`'s own pre-PR docstring already flagged the duplication ("uses the same heuristic as ... orgTokenRedirectURL"), so this is a known tradeoff, not an oversight. Downgraded to nit since a maintainer may reasonably prefer a follow-up refactor PR.

Both blocks from this PR's own diff, byte-for-byte identical:

```go
// server.go:144 (orgURLs)
canSubdomain := subdomainRouting &&
    strings.Contains(host, ".") &&
    !strings.HasPrefix(host, "localhost") &&
    !isIPLiteral(host)

// api_oidc.go:342 (orgTokenRedirectURL)
canSubdomain := subdomainRouting &&
    strings.Contains(host, ".") &&
    !strings.HasPrefix(host, "localhost") &&
    !isIPLiteral(host)
```

Suggested fix — extract a shared helper and have both functions call it:

```go
// resolveSubdomainHost parses "scheme://host:port" and reports whether this
// deployment can serve the given host on a wildcard subdomain — the single
// source of truth for the subdomain-vs-path decision used by both orgURLs
// and orgTokenRedirectURL.
func resolveSubdomainHost(baseURL string, subdomainRouting bool) (scheme, host, port string, canSubdomain bool) {
	const sep = "://"
	i := strings.Index(baseURL, sep)
	if i < 0 {
		return "", "", "", false
	}
	scheme = baseURL[:i]
	hostAndPort := baseURL[i+len(sep):]
	host = hostAndPort
	if c := strings.LastIndex(hostAndPort, ":"); c > 0 {
		host = hostAndPort[:c]
		port = hostAndPort[c:]
	}
	canSubdomain = subdomainRouting &&
		strings.Contains(host, ".") &&
		!strings.HasPrefix(host, "localhost") &&
		!isIPLiteral(host)
	return scheme, host, port, canSubdomain
}
```

Suggest as a follow-up refactor PR rather than blocking this one, since the current fix is correct for both call sites as-is.

unidoc-ahall:
Thanks — all three addressed in the same pass, since #1 meant touching this function anyway:

- **#1 (custom domain in OIDC redirect):** `orgTokenRedirectURL` now checks `org.Domain` first, mirroring `orgURLs` precedence (custom domain > subdomain > path). A custom-domain org's OIDC login lands on its own domain instead of the apex.
- **#2 (coverage):** added `TestOrgTokenRedirectURL` — subdomain-on hop, routing-off real domain → path-based, localhost → path-based, and custom-domain-wins (locks in #1). Good catch that the claim didn't extend to this function.
- **#3 (duplication):** took it now rather than a follow-up — extracted `resolveSubdomainHost` as the single source of truth for the subdomain-vs-path decision; both `orgURLs` and `orgTokenRedirectURL` call it, so the two-copy gate that caused #108 is gone.

gofmt clean, build + vet green, `TestOrgURLs` + `TestOrgTokenRedirectURL` pass. Re-requesting review.

unidoc-alip (approved):
looks good to me

PR #110, opened 2026-07-02
feat(review): @-mention members in review comments → notification (#4, slice 1)

unidoc-ahall:
Part of #4.

Typing `@handle` in a review comment now notifies that org member. `notifyReviewMentions` parses `@handle` tokens, resolves each against org members (email local-part — `maria@… → @maria` — or name-slug, case-insensitive), and creates an in-app notification linking to the review. Self-mentions and unknown handles are ignored. Surfaces in the existing notification bell.

Uses the existing `notifications` table + `CreateNotificationByEmail` — **no DB migration**.

**Tests:** `tests/test_review_mentions.py` — a mention notifies the member (with the `/reviews/<id>` link); an unknown handle and an author self-mention produce nothing. Green twice against the stack; runs in CI.

**Scope:** slice 1 = review-comment mentions (the notify core; works today by typing `@handle`). Follow-up slices: frontend @autocomplete in the comment box, and mentions in rollback plans / change requests (the rest of #4).

### DIFF
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -81,7 +81,7 @@ var mentionRe = regexp.MustCompile(`@([a-zA-Z0-9._-]+)`)
 // @-mentioned in a review comment (#4), excluding the comment's author. Matching
 // is case-insensitive against the member's email local-part or name-slug.
 func (s *Server) notifyReviewMentions(ctx context.Context, orgID, reviewID int, actor, body string) {
-	matches := mentionRe.FindAllStringSubmatch(body, -1)
+	matches := mentionRe.FindAllStringSubmatchIndex(body, -1)
 	if len(matches) == 0 {
 		return
 	}
@@ -89,26 +89,38 @@ func (s *Server) notifyReviewMentions(ctx context.Context, orgID, reviewID int, 
 	if err != nil {
 		return
 	}
-	byHandle := map[string]string{} // handle → email
+	// Keep email local-parts and name-slugs in SEPARATE maps. One member's email
+	// local-part can equal another member's name-slug ("alicesmith@…" vs "Alice
+	// Smith"); merging both into one map would let whichever sorts later silently
+	// overwrite the other and mis-route the mention. Local-parts are unique
+	// (emails are), so prefer them and only fall back to a name-slug on a miss.
+	byLocal := map[string]string{}
+	bySlug := map[string]string{}
 	for _, u := range users {
 		local := u.Email
 		if at := strings.IndexByte(local, '@'); at > 0 {
 			local = local[:at]
 		}
-		byHandle[strings.ToLower(local)] = u.Email
+		byLocal[strings.ToLower(local)] = u.Email
 		if slug := strings.ToLower(strings.ReplaceAll(u.Name, " ", "")); slug != "" {
-			byHandle[slug] = u.Email
+			bySlug[slug] = u.Email
 		}
 	}
-	snippet := body
-	if len(snippet) > 200 {
-		snippet = snippet[:200] + "…"
-	}
+	snippet := truncateStr(body, 200)
 	title := fmt.Sprintf("%s mentioned you in a review comment", actor)
 	link := fmt.Sprintf("/reviews/%d", reviewID)
 	notified := map[string]bool{}
 	for _, m := range matches {
-		email, ok := byHandle[strings.ToLower(m[1])]
+		// Skip "@scope/pkg" tokens (e.g. @babel/core): a '/' right after the
+		// handle marks a package path or URL segment, not a mention.
+		if end := m[3]; end < len(body) && body[end] == '/' {
+			continue
+		}
+		handle := strings.ToLower(body[m[2]:m[3]])
+		email, ok := byLocal[handle]
+		if !ok {
+			email, ok = bySlug[handle]
+		}
 		if !ok || email == actor || notified[email] {
 			continue // unknown handle, self-mention, or already notified
 		}
@@ -834,8 +846,12 @@ func (s *Server) handleAddReviewComment(c echo.Context) error {
 		Detail:     detail,
 	})
 
-	// Notify anyone @-mentioned in the comment (#4).
-	s.notifyReviewMentions(ctx, orgID, id, actor, req.Body)
+	// Notify anyone @-mentioned in the comment or its suggested replacement (#4).
+	mentionText := req.Body
+	if req.SuggestionBody != nil && *req.SuggestionBody != "" {
+		mentionText += "\n" + *req.SuggestionBody
+	}
+	s.notifyReviewMentions(ctx, orgID, id, actor, mentionText)
 
 	return c.JSON(http.StatusCreated, comment)
 }
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -73,6 +73,50 @@ func docLink(docID string) string {
 	}
 }
 
+// mentionRe matches @handle tokens in a comment body. A handle is a member's
+// email local-part (maria@… → @maria) or their name with spaces removed.
+var mentionRe = regexp.MustCompile(`@([a-zA-Z0-9._-]+)`)
+
+// notifyReviewMentions creates an in-app notification for each org member
+// @-mentioned in a review comment (#4), excluding the comment's author. Matching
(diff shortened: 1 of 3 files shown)


unidoc-alip (changes requested):
Slice 1 of #4 does what it says: `@handle` in a review comment now notifies the matching org member (email local-part or name-slug, case-insensitive), skips self-mentions and unknown handles, and dedups repeats. Verified live (server + DB) — matching, case-insensitivity, and dedup all behave correctly, and `go build`/`go vet` are clean. Note this PR does not close #4 on its own — the issue also covers rollback plan and change-request mentions, which the PR description itself calls out as follow-up slices; #4 should stay open (or be re-scoped) until those land. A few non-blocking correctness/cleanup points below, plus one flaky test worth tightening before it bites someone running it standalone.

---

### 1. [should-fix] `internal/isms/api/api_collab.go` line 98

`byHandle` writes the email-local-part key and the name-slug key into the same map with no collision handling. If one member's email local-part equals another member's name-slug (e.g. member A is `alicesmith@corp.com`, member B is named "Alice Smith"), whichever user sorts later in `ListOrgUsers`' `ORDER BY u.name` silently overwrites the earlier entry — a mention intended for one member notifies the other instead, with no error or log. Consider keeping local-part and name-slug in separate maps (or only falling back to name-slug when the local-part lookup misses) so a collision can't happen silently.

Reproduced live: posting `@bobsmith` intending to reach reviewer "Bob Smith" instead notified an unrelated "Zed Person" whose email local-part happened to be `bobsmith` and who sorted later in the roster.

Suggested fix:
```go
byLocal := map[string]string{}
bySlug := map[string]string{}
for _, u := range users {
    local := u.Email
    if at := strings.IndexByte(local, '@'); at > 0 {
        local = local[:at]
    }
    byLocal[strings.ToLower(local)] = u.Email
    if slug := strings.ToLower(strings.ReplaceAll(u.Name, " ", "")); slug != "" {
        bySlug[slug] = u.Email
    }
}
// lookup: email, ok := byLocal[handle]; if !ok { email, ok = bySlug[handle] }
```
This still can't disambiguate two *local-parts* colliding (impossible — emails are unique) or two *name-slugs* colliding (rare, needs its own decision), but it removes the cross-category collision entirely.

---

### 2. [should-fix] `tests/test_review_mentions.py` line 16

`body.get("data", body)` returns `None` when the server responds `{"data": null}` — which is exactly what `GET /notifications` returns for a user with zero notifications (a nil Go slice serializes to JSON `null`, not `[]`). Running this file standalone (`pytest tests/test_review_mentions.py -v` against a fresh DB) crashes `test_unknown_handle_and_self_mention_produce_no_notification` with `TypeError: 'NoneType' object is not iterable`. It only passes today as part of the full suite because earlier tests leave a residual notification on the shared `ADMIN_EMAIL` fixture account, which happens to dodge the `null` case.

Suggested fix:
```python
def _notifications(api_url, headers):
    r = requests.get(f"{api_url}/notifications", headers=headers)
    body = r.json()
    items = body.get("data", body) if isinstance(body, dict) else body
    return items or []
```
Optionally also fix the root cause in the Go handler backing `GET /notifications` so it never serializes `null` for an empty result — initialize the slice (`notifications := []Notification{}`) before the scan loop instead of leaving it as a nil `var notifications []Notification`. One-line change, outside this PR's diff, but would prevent the same class of bug in any frontend code that calls `.map()`/`.forEach()` on `data` without a guard.

---

### 3. [should-fix] `internal/isms/api/api_collab.go` line 104

This is now the third near-identical truncate-with-ellipsis implementation in the codebase — `notify.truncate(s, n)` in `internal/isms/notify/notify.go` and `truncateStr(s, n)` in `internal/isms/api/api_entity_comments.go` already do the same job, and the three have already drifted (`truncateStr` reserves 3 bytes for `"..."`, this one and `notify.truncate` don't; this one uses `"…"` vs the others' `"..."`). `truncateStr` is in the same package (`api`) as this new code, so it was directly callable.

Suggested fix:
```go
snippet := truncateStr(body, 200)
```
If the off-by-3 truncation length or the `"..."` vs `"…"` glyph matters for this notification's UX, that's a one-line tweak to `truncateStr` (or a second small helper) rather than a third parallel implementation.

---

### 4. [nit] `internal/isms/api/api_collab.go` line 78

`@([a-zA-Z0-9._-]+)` has no word-boundary/context awareness, so incidental `@token`s in pasted text — scoped npm packages (`@babel/core`), URLs with an `@`-segment, etc. — get extracted as mention candidates like any real `@handle`. Reproduced: an org member with email local-part `babel` got a spurious "mentioned you" notification from a comment referencing `@babel/core`, with zero intent to mention anyone. Low real-world likelihood; worth a comment or a tighter regex if false-positive notifications turn out to be noisy in practice.

Suggested fix: a leading-boundary tweak does not fix the reproduced case (there's already a space before `@babel`). The actual gap is the trailing side — the char class stops at `/`, so a scoped-package reference reads as a bare handle. Go's `regexp` (RE2) has no lookahead, so a post-match filter is cleanest:
```go
for _, m := range mentionRe.FindAllStringSubmatchIndex(body, -1) {
    if m[3] < len(body) && body[m[3]] == '/' {
        continue // "@scope/pkg" — not a mention
    }
    handle := body[m[2]:m[3]]
    // ... existing lookup logic using handle
}
```
Narrow, low-priority — reasonable to defer unless spurious notifications prove noisy in practice.

---

### 5. [nit] `internal/isms/api/api_collab.go` line 838

`notifyReviewMentions` is called with `req.Body` only, not `req.SuggestionBody`. An `@handle` typed inside a suggested-replacement is never scanned — if the accompanying free-text `req.Body` is empty, `comment.Body` gets defaulted to a placeholder and no mention parsing happens at all. Confirmed live: a suggestion-only comment mentioning a reader in `suggestion_body` produced zero notifications for that reader. Likely fine as an explicit scope boundary for slice 1, but worth a one-line note in the PR description or a follow-up ticket so it isn't mistaken for an oversight later.

Suggested fix, if suggestion-text mentions should notify:
```go
mentionSource := req.Body
if req.SuggestionBody != nil {
    mentionSource += "\n" + *req.SuggestionBody
}
s.notifyReviewMentions(ctx, orgID, id, actor, mentionSource)
```
If it's an intentional scope boundary for slice 1, no code change is needed — just document the limitation.

---

### 6. [nit] `internal/isms/api/api_collab.go` line 83

`notifyReviewMentions` is hardwired to review comments (`reviewID int`, `"/reviews/%d"` link, "in a review comment" title). The PR description says the next slices add mentions to rollback plans and change requests — at that point this function will either get copy-pasted with s/review/change-request/ or need retrofitting into a generic helper. Reasonable to inline for a first slice; flagging so it's a deliberate choice for slice 2, not a surprise later.

Suggested fix, when slice 2 is imminent:
```go
func (s *Server) notifyMentions(ctx context.Context, orgID int, actor, body, link, contextLabel string) {
    // same regex/byHandle/dedup logic, parameterized title := fmt.Sprintf("%s mentioned you in %s", actor, contextLabel)
}
// call site: s.notifyMentions(ctx, orgID, actor, req.Body, fmt.Sprintf("/reviews/%d", id), "a review comment")
```

unidoc-ahall:
Thanks — live-verified findings, all addressed:

- **#1 (map collision):** split into `byLocal` / `bySlug`; prefer the unique email local-part, fall back to name-slug. No more silent cross-category overwrite.
- **#2 (null notifications):** fixed both — the test helper returns `[]`, and `ListNotifications` now initialises `[]Notification{}` so an empty result serialises as `[]` not `null` (protects frontend `.map`/`.forEach` callers too). Good catch that it only passed via residual fixture state.
- **#3 (third truncate impl):** now calls the existing `truncateStr(body, 200)`.
- **#4 (scoped package):** skip a handle immediately followed by `/` (`@babel/core`); added a regression test.
- **#5 (suggestion body):** now scans `req.SuggestionBody` too, so a mention inside a suggested replacement notifies.
- **#6 (hardwired to reviews):** deliberate for slice 1 — I'll lift it into a generic `notifyMentions(…, link, contextLabel)` when slice 2 (rollback plans / change requests) lands, rather than copy-paste.

Agreed #4 stays open — this PR is slice 1 only ("Part of #4", not Closes). gofmt clean, build + vet green, mention tests pass. Re-requesting review.

unidoc-alip (approved):
looks good to me

PR #112, opened 2026-07-03
fix(auth): hide signup UI when self-registration is disabled (#111)

### LINKED ISSUE #111: Login/signup UI ignores ISMS_USER_SIGNUP (shows Sign up when self-registration is off)
On a deployment with `ISMS_USER_SIGNUP` unset (self-registration disabled), the login page still renders `Don't have an account? Sign up` and `/signup` shows a live registration form. The backend `POST /auth/signup` correctly returns 403, but the UI never learned the flag, so it advertises a dead path.

Seen live on isms.stsplatform.com (`ISMS_USER_SIGNUP=0`).

**Fix:** `/api/v1/config` now reports `signup_enabled`; Login hides the Sign up affordances, Signup redirects to /login when disabled, and Landing's cloud CTAs gate on it. No DB migration.

unidoc-ahall:
Fixes #111.

With `ISMS_USER_SIGNUP` unset, the login page still showed `Don't have an account? Sign up` and `/signup` rendered a live form — even though `POST /auth/signup` returns 403. The UI had no signal for the flag. Surfaced live on isms.stsplatform.com.

**Change:** `/api/v1/config` now reports `signup_enabled` (derived from `ISMS_USER_SIGNUP`, same source as the signup handler's own 403 gate).
- **Login.vue** — hides both `Don't have an account?` blocks when disabled.
- **Signup.vue** — redirects to `/login` when disabled (no dangling form).
- **Landing.vue** — cloud CTAs (`Start in cloud` / `Start free`) hide; the managed-cloud card falls back to `Sign in`.

The redirect query-param (`?redirect=/sts`) was read correctly all along — this was purely the missing flag.

**No DB migration.**

**Tests:** `tests/test_signup_config.py` — `/config`'s `signup_enabled` must agree with the real `/auth/signup` gate (403 iff disabled). Green against the stack; runs in CI.

### DIFF
--- a/internal/isms/api/server.go
+++ b/internal/isms/api/server.go
@@ -865,10 +865,18 @@ func (s *Server) handleGetConfig(c echo.Context) error {
 		// for demo / dev where wildcard DNS + TLS isn't set up.
 		SubdomainRouting bool   `json:"subdomain_routing"`
 		ApexHost         string `json:"apex_host,omitempty"`
+
+		// Whether self-registration is enabled (ISMS_USER_SIGNUP). The frontend
+		// hides the "Sign up" affordances when this is false — the signup handler
+		// itself is the hard gate (403), this just keeps the UI honest.
+		SignupEnabled bool `json:"signup_enabled"`
 	}
 	resp := configResponse{}
 	resp.SubdomainRouting = s.subdomainRouting
 	resp.ApexHost = s.apexHost
+	if sv := os.Getenv("ISMS_USER_SIGNUP"); sv == "true" || sv == "1" {
+		resp.SignupEnabled = true
+	}
 
 	// Branding from org settings (Postgres)
 	branding := map[string]string{}
--- a/tests/test_signup_config.py
+++ b/tests/test_signup_config.py
@@ -1,0 +1,35 @@
+"""Regression: /config must report signup_enabled, and it must match the gate.
+
+The bug: the login/signup UI hard-coded a "Sign up" link, so a deployment with
+ISMS_USER_SIGNUP unset (self-registration off) still advertised signup even
+though POST /auth/signup returns 403. /config now exposes `signup_enabled` so the
+frontend can hide the affordance. This test pins the flag to the real gate: the
+value /config reports must agree with whether the signup endpoint accepts a
+registration (403 = disabled, anything else = enabled).
+"""
+import uuid
+
+import requests
+
+
+def test_config_signup_enabled_matches_gate(api_url):
+    cfg = requests.get(f"{api_url}/config")
+    assert cfg.status_code == 200, cfg.text
+    body = cfg.json()
+    assert "signup_enabled" in body, "/config must expose signup_enabled"
+    enabled = body["signup_enabled"]
+    assert isinstance(enabled, bool), f"signup_enabled must be a bool, got {enabled!r}"
+
+    # Probe the actual gate with a throwaway registration.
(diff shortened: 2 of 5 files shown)


unidoc-alip (changes requested):
Solid, well-scoped fix for #111 — `/config` now exposes `signup_enabled`, and Login/Landing correctly fail closed (default `ref(false)`, only flip true after a successful config fetch) so a fetch error hides the signup affordances. Verified end-to-end in a browser with `ISMS_USER_SIGNUP` unset and set: Login hides both "Sign up" blocks, Landing swaps the cloud CTAs for "Sign in", and `/signup` redirects to `/login`.

One inconsistency worth a look: `Signup.vue`'s redirect-when-disabled check lives inside the same `try` as the config fetch, so if `getConfig()` throws, the redirect never runs and the form (rendered unconditionally in the template) stays up and submittable — the opposite fail-safe direction from Login/Landing. Not a security hole (the backend still 403s the POST), but it reintroduces the exact "dangling dead form" UX this PR set out to fix, just via a different trigger.

---

### 1. [should-fix] `web/src/views/Signup.vue` line 105

This redirect-when-disabled check is inside the `try` block, so if `api.getConfig()` throws (network blip, `/config` temporarily unreachable), the check never runs and the form falls through to rendering (it's in the template's unconditional `v-else` branch, not gated behind a loading/ready flag) — dangling exactly the dead signup path this PR is fixing, just triggered by a fetch failure instead of the flag being unset.

Login.vue and Landing.vue avoid this by defaulting `signupEnabled = ref(false)` and only setting it `true` inside the try — a fetch failure there fails closed (hides signup UI). Consider the same pattern here: default to redirecting/hiding on failure, or explicitly redirect in the `catch` block too.

Suggested fix (minimal — redirect in `catch` too):

```js
onMounted(async () => {
  try {
    const cfg = await api.getConfig()
    if (cfg.signup_enabled !== true) {
      router.replace('/login')
      return
    }
    if (cfg.organization?.name) orgName.value = cfg.organization.name
    // ...rest of branding assignment unchanged
  } catch {
    // Config unreachable — fail closed, same as signup_enabled === false.
    router.replace('/login')
  }
})
```

---

### 2. [nit] `internal/isms/api/server.go` line 877

The `ISMS_USER_SIGNUP` truthiness check (`sv == "true" || sv == "1"`) is duplicated verbatim from `api_auth.go:194-195`. Since the PR's whole premise is "the frontend flag and the backend gate must agree," a shared helper (e.g. `func signupEnabled() bool`) would make that agreement structural instead of relying on both call sites staying in sync by hand.

Suggested fix:

```go
// shared predicate, called from both api_auth.go and server.go
func signupEnabled() bool {
    sv := os.Getenv("ISMS_USER_SIGNUP")
    return sv == "true" || sv == "1"
}
```

Then `api_auth.go:194-196` becomes `if !signupEnabled() { return echo.NewHTTPError(http.StatusForbidden, "self-registration is disabled") }`, and `server.go:877-879` becomes `resp.SignupEnabled = signupEnabled()`.

Not a bug today — both sites are currently identical — but low-cost to make structural. Safe to skip.

unidoc-ahall (commented):
Fixed — Signup.vue now redirects to `/login` in the `catch` block too, same fail-closed behaviour as Login.vue and Landing.vue. `signup_enabled` helper refactor noted; keeping that as a follow-up since you flagged it safe to skip.

unidoc-alip (approved):
looks good to me

Issue #17, opened 2026-06-06
Mobile UI polish — usable but not good


unidoc-ahall:
## Summary

The web UI works on mobile but the experience is rough. This is a polish pass,
not a rewrite.

## Scope (to be refined)

- Audit the primary flows on a phone viewport: login, document read/edit, the
  registers (risks/incidents/etc.), reviews.
- Fix the worst offenders: tap targets, table overflow, modal sizing, sidebar
  behavior.

## Out of scope

- A separate native app.
- A dedicated mobile design system.

## Acceptance criteria

- [ ] Core read + review flows are comfortable on a phone.
- [ ] No horizontal-scroll / clipped-content on the main screens.

unidoc-ahall:
Solved in #113

PR #113, opened 2026-07-03
fix(web): register tables + review split usable on mobile (#17, slice 1)

unidoc-ahall:
Part of #17.

On a 390px viewport the registers overflowed sideways — the filter bar pushed the page wider than the screen and the table columns were clipped with no way to reach them. The review side-by-side split also crammed two panes into phone width.

**Change (frontend-only, no migration):**
- **Filter bars wrap** (`flex-wrap`) so controls stack instead of forcing horizontal page scroll — risks, assets, suppliers, legal, systems.
- **Register tables** (those six + objectives) sit in an `overflow-x-auto` container: off-screen columns are reachable by swiping, page itself no longer scrolls sideways.
- **Review split** (`SideBySideReview`) stacks vertically (`flex-col md:flex-row`) on narrow screens; synced scroll and the unified inline default are unchanged.

**Left for later:** Admin / Audit / Dashboard use multi-card layouts (tabs, calendar, settings) rather than a single register table — a follow-up slice if needed. The nav drawer and the Documents / Reviews list views were already responsive.

**Tests:** `tests/test_e2e_mobile.py` at 390x844 — asserts no page-level horizontal overflow and that each register table is in a horizontal scroll container. Green twice against the stack; runs in CI.

**Verified visually** on risks + suppliers: filter bars wrap cleanly, page no longer scrolls sideways.

### DIFF
--- a/devenv/Justfile
+++ b/devenv/Justfile
@@ -92,15 +92,18 @@ run +CMD:
 
 # ── Test environment ─────────────────────────────────────────────────────────
 
-# Run pytest from inside the dev container against the test server
+# Run pytest from inside the dev container against the test server.
+# Rebuilds web/dist first: isms-test serves the bundle live from disk, so an
+# e2e test would otherwise silently run against a stale UI after a .vue change
+# (test-restart only reloads Go). ~1s; harmless for backend-only runs.
 test +ARGS="tests/":
     {{compose}} exec -u isms isms bash -lc \
-        "cd /home/isms/workspace/isms && ISMS_TEST_URL=http://isms-test:9090 .venv/bin/pytest {{ ARGS }}"
+        "cd /home/isms/workspace/isms && just build-web && ISMS_TEST_URL=http://isms-test:9090 .venv/bin/pytest {{ ARGS }}"
 
-# Run the Playwright e2e suite against the test server
+# Run the Playwright e2e suite against the test server (web/dist rebuilt first).
 test-e2e +ARGS="tests/test_e2e_browser.py tests/test_e2e_routing.py":
     {{compose}} exec -u isms isms bash -lc \
-        "cd /home/isms/workspace/isms && ISMS_TEST_URL=http://isms-test:9090 .venv/bin/pytest {{ ARGS }} -v"
+        "cd /home/isms/workspace/isms && just build-web && ISMS_TEST_URL=http://isms-test:9090 .venv/bin/pytest {{ ARGS }} -v"
 
 # Restart the test server only — reloads current code, keeps the test DB.
 # Incremental `go run` rebuild, so this is the fast path after code changes.
--- a/devenv/Justfile
+++ b/devenv/Justfile
@@ -138,6 +138,13 @@ gs:
 gl:
     {{compose}} exec isms bash -lc "git log --oneline -20"
 
+# ── Web ──────────────────────────────────────────────────────────────────────
+
+# Build the Vue frontend as the isms user so web/dist is never owned by
+# the subuid-mapped root (UID 100000 on the host).
+build-web:
+    {{compose}} exec -u isms isms bash -lc "cd /home/isms/workspace/isms && just build-web"
+
 # ── Go ───────────────────────────────────────────────────────────────────────
 
 # Go build of the ISMS binary
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -566,10 +566,11 @@ func (s *Server) handleReviewDiff(c echo.Context) error {
 	if from == "" {
 		if to != "" {
 			// Per-event: the proposal vs its parent. A root commit has no parent —
-			// leave the baseline empty (the whole document is genuinely new) rather
-			// than relying on a silently-swallowed resolve failure.
-			if _, perr := st.RefHash(to + "^"); perr == nil {
-				from = to + "^"
+			// leave the baseline empty (the whole document is genuinely new).
+			// Use ParentHash (direct commit object lookup) rather than "sha^" revision
+			// notation, which can silently fail on some bare-repo configurations.
+			if parentSHA, perr := st.ParentHash(to); perr == nil {
+				from = parentSHA
 			}
 		} else if review.SentHead != "" && (hasBranch || review.CommitHash != "") {
 			from = review.SentHead
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -3580,6 +3580,30 @@ func (s *Server) handleConfirmDocumentReview(c echo.Context) error {
 	})
 }
 
+// --- Open-review probe (read-only) ---
+
+// handleGetOpenReview returns the open (or changes_requested) review for a
+// document without firing any side effects.  HTTP 200 with {review_id} if one
+// exists, HTTP 404 otherwise.
+func (s *Server) handleGetOpenReview(c echo.Context) error {
+	orgID := getOrgID(c)
+	docID := c.Param("docId")
+	ctx := c.Request().Context()
+
+	review, err := s.db.GetOpenReviewForDocument(ctx, orgID, docID)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, fmt.Sprintf("checking open review: %v", err))
+	}
+	if review == nil {
+		return echo.NewHTTPError(http.StatusNotFound, "no open review for this document")
+	}
+	return c.JSON(http.StatusOK, map[string]interface{}{
+		"review_id": review.ID,
+		"status":    review.Status,
+		"version":   review.Version,
+	})
+}
+
 // --- Review Send (compound operation) ---
 
 func (s *Server) handleReviewSend(c echo.Context) error {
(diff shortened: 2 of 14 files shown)


unidoc-ahall:
The `test_e2e_mobile` failure in `just test` was **stale assets, not a code/test bug**: isms-test serves `web/dist` live from disk, and `just test` never rebuilt it (`test-restart` only reloads Go), so the test ran against the pre-`overflow-x-auto` bundle. After a fresh `just build-web` the mobile test is 3/3 green.

Fixed the workflow gap in the same PR: `just test` and `just test-e2e` now rebuild `web/dist` first, so no e2e can silently run against a stale UI again. CI was always fine (it builds web before the integration job).

unidoc-ahall:
Fixed the CI failure (`objectives: page overflows horizontally by 37px`).

**Root cause:** `Objectives.vue` was the only register view missing the outer `<div class="overflow-y-auto">` wrapper the other five have. That wrapper's `overflow-y: auto` forces `overflow-x` to compute to `auto` (CSS resolves a visible/auto pair to auto/auto), which contains horizontal overflow inside the wrapper instead of leaking it to the page. Added the same wrapper to Objectives.

**Proven locally:** a 900px-wide child leaks 542px to the page without the wrapper, 0px with it; all six registers pass at 390×844.

Also made `test_e2e_mobile` depend on the `tokens` fixture so it self-provisions and runs standalone (it previously only worked inside the full suite — running it alone 401'd on login).

unidoc-alip (changes requested):
Verified end-to-end against a live server at 390×844 (all six registers, seeded with long-title rows, plus the review split view) — the core claims hold: filter bars wrap, tables scroll horizontally inside their card (confirmed columns are reachable after a simulated swipe), and the review diff stacks vertically with no page overflow, even against a pathological unbroken 140-char string. No must-fix issues found. A few should-fix items below: the review-split panes don't shrink their `max-h-[80vh]` cap when stacked (up to ~160vh of scroll on a phone), the PR description's "Admin/Audit/Dashboard use multi-card layouts" justification for skipping those views doesn't hold for Audit's Programmes/Findings tables (still `overflow-hidden`, still single wide tables), the new test setup in `test_suggestions.py` triggers a real email dispatch + notification write as a side effect of merely probing for an existing review, and the new `ParentHash` duplicates parent-resolution logic already inlined in `DiffFiles`.

---

### 1. [should-fix] `tests/test_suggestions.py` line 34

This "probe" POST doesn't just check for an existing review — `handleReviewSend`'s open-review branch (`internal/isms/api/api_collab.go:3639-3663`) writes a notification and sends a real email via `mailer.SendReviewRequestBranded` to every entry in `reviewers` before this code closes and recreates the review a few lines below. On a persistent stack where a prior run left the review open (the exact scenario this test's docstring is written for), every re-run now costs 2 extra HTTP round trips, a DB notification write, and a live SMTP send purely to check existence — and can flake under SMTP throttling.

Consider a read-only existence check (e.g. exposing `GetOpenReviewForDocument` via a GET) instead of a POST that fires production side effects, or at minimum close any existing open review unconditionally before a single `POST`.

Suggested fix:

```python
existing = requests.get(f"{api_url}/documents/iso27001-6-1/reviews/open", headers=admin_headers)
if existing.status_code == 200:
    requests.put(f"{api_url}/reviews/{existing.json()['review_id']}/status",
                 headers=admin_headers, json={"status": "closed"})
r = requests.post(f"{api_url}/documents/iso27001-6-1/reviews",
                  headers=admin_headers,
                  json={"reviewers": [READER_EMAIL], "message": "Suggestion test"})
assert r.status_code == 201, f"Create fresh review failed: {r.text}"
TestSuggestionWorkflow.review_id = r.json()["review_id"]
```

This requires adding a lightweight `GET /documents/:docId/reviews/open` endpoint backed by the existing `GetOpenReviewForDocument` — no new side effects, and the test always ends up with exactly one 201-returning POST.

---

### 2. [should-fix] `web/src/components/SideBySideReview.vue` line 25

Both panes keep `max-h-[80vh]` even after the container becomes `flex-col` on mobile (line 22), so stacked panes can total up to ~160vh of independently-scrolling content on a phone instead of splitting the available height. Confirmed visually at 390×844: the "Previous" pane alone consumes nearly the full viewport before "Current" is reachable.

Consider a mobile-specific cap so both panes fit closer to one viewport when stacked.

Suggested fix:

```diff
-      <div ref="leftPane" data-pane="previous" @scroll="syncScroll('left')" class="flex-1 min-w-0 border-b md:border-b-0 md:border-r border-slate-800 overflow-y-auto max-h-[80vh]">
+      <div ref="leftPane" data-pane="previous" @scroll="syncScroll('left')" class="flex-1 min-w-0 border-b md:border-b-0 md:border-r border-slate-800 overflow-y-auto max-h-[40vh] md:max-h-[80vh]">
```
```diff
-      <div ref="rightPane" data-pane="current" @scroll="syncScroll('right')" class="flex-1 min-w-0 overflow-y-auto max-h-[80vh]">
+      <div ref="rightPane" data-pane="current" @scroll="syncScroll('right')" class="flex-1 min-w-0 overflow-y-auto max-h-[40vh] md:max-h-[80vh]">
```

---

### 3. [should-fix] `web/src/views/Audit.vue` line 203

The PR description justifies skipping Admin/Audit/Dashboard because they "use multi-card layouts... rather than a single register table," but Audit.vue's Programmes table (line 203-204) and Findings table (line 298-299) are wide `<table>` elements in the same `bg-slate-900 border border-slate-800 rounded-xl overflow-hidden` wrapper this PR fixed elsewhere — they'll still clip columns on a 390px viewport with no way to reach them, the exact bug this PR fixes for the other six registers.

If genuinely deferred, worth correcting the PR description so "left for later" doesn't read as "these views don't have the bug" — they do, for at least these two tables.

Suggested fix (apply at both line 203 and line 298):

```diff
-        <div v-else class="bg-slate-900 border border-slate-800 rounded-xl overflow-hidden">
+        <div v-else class="bg-slate-900 border border-slate-800 rounded-xl overflow-x-auto">
           <table class="w-full">
```

---

### 4. [should-fix] `internal/isms/store/git.go` line 462

`ParentHash` duplicates the parent-commit resolution already inlined in `DiffFiles` (same file, ~line 664-676: `toCommit.ParentHashes` + `s.repo.CommitObject(parents[0])`). Two independent implementations of "resolve a commit's first parent" now exist with different error-wrapping and different root-commit handling — one errors on a root commit, the other silently treats it as "diff against empty." A future fix applied to one is easily missed in the other.

Consider having `DiffFiles` call the new `ParentHash` (or extracting both into one shared helper) instead of leaving the logic duplicated.

Suggested fix:

```go
// firstParentCommit resolves the commit at hash and returns its first parent commit,
// or (nil, nil) if hash is a root commit with no parents.
func (s *Store) firstParentCommit(hash plumbing.Hash) (*object.Commit, error) {
    commit, err := s.repo.CommitObject(hash)
    if err != nil {
        return nil, fmt.Errorf("commit %q not found: %w", hash, err)
    }
    if len(commit.ParentHashes) == 0 {
        return nil, nil // root commit
    }
    return s.repo.CommitObject(commit.ParentHashes[0])
}
```

`ParentHash` and `DiffFiles`'s inline block both call this and interpret the `nil, nil` root-commit case in their own existing way, collapsing the duplicated resolution logic into one place.

---

### 5. [nit] `web/src/views/Assets.vue` line 217

The same two-class fix (`flex` → `flex flex-wrap` on the filter bar, `overflow-hidden` → `overflow-x-auto` on the table wrapper) is hand-applied identically across six views (Assets, Legal, Risks, Suppliers, Systems, Objectives — the last one twice, for its Programs and Objectives tables). A shared `RegisterTable`/filter-bar component would apply this to every current and future register automatically instead of relying on each new register remembering the same patch (see finding #3 — Audit.vue is exactly the kind of view that got missed).

Not blocking — consistent with the codebase's existing per-view style — but worth a follow-up given a sibling view (Audit.vue) already has the same gap. Suggested direction: extract a shared `web/src/components/RegisterTable.vue` wrapping `<slot name="filters">` in `flex flex-wrap` and `<slot name="table">` in `overflow-x-auto`, then migrate all six views (and fix Audit.vue's two tables per finding #3) to use it.

---

### 6. [nit] `tests/test_suggestions.py` line 39

The conditional close-and-recreate (`if existing.status_code == 200: ... else: r = existing`) is harder to read than closing any pre-existing open review unconditionally and then always issuing one `POST` that's guaranteed to return 201. The current form needs a reader to trace both branches to see that `r` always ends up meaning "the review to use for the rest of the test."

Subsumed by finding #1's fix — once the existence check is a GET, the test collapses to a single unconditional POST with no branch needed to figure out what `r` means.

unidoc-ahall (commented):
All four should-fixes addressed:

1. **GET `/documents/:docId/reviews/open`** — new read-only endpoint backed by the existing `GetOpenReviewForDocument` DB call (200 + `{review_id,status,version}` if found, 404 otherwise). `test_suggestions.py` now uses this for the idempotency check instead of the probe POST.

2. **SideBySideReview.vue** — `max-h-[40vh] md:max-h-[80vh]` on both panes.

3. **Audit.vue** — `overflow-x-auto` on Programmes and Findings table wrappers.

4. **git.go** — `firstParentCommit(hash)` private helper extracted; both `ParentHash` and `DiffFiles` call it.

Bundled `devenv/Justfile` `build-web` recipe in the same commit (runs under `-u isms` to avoid root-owned `web/dist` on the host).

unidoc-alip (approved):
looks good to me

Issue #116, opened 2026-07-03
Deduplicate renderMention (copy-pasted in CommentsPanel.vue and EntitySidebar.vue)


unidoc-ahall:
`renderMention(body)` — escape HTML, then wrap `@handle`/`@email` in a styled span — exists byte-identical in `web/src/components/CommentsPanel.vue` and `web/src/components/EntitySidebar.vue`.

Same drift class as the triple truncate-with-ellipsis flagged on #110: two copies of an escaping-sensitive function will eventually diverge, and this one is XSS-relevant (the escape-first step must never be lost in either copy).

**Fix:** move it to a composable (e.g. `web/src/composables/useRenderMention.js`) and import it in both components. Also a natural home for the #4 slice-2 frontend work (@autocomplete), which will want the same handle-matching in one place.

Found in a codebase review; not user-reported.

unidoc-ahall:
Resolved already in #117 .

Issue #114, opened 2026-07-03
OIDC session cleanup goroutine uses the request context and is cancelled mid-flight


unidoc-ahall:
`internal/isms/api/api_oidc.go` (handleOIDCAuthorize) fires the expired-session cleanup as:

```go
go s.db.DeleteExpiredOIDCSessions(ctx)
```

`ctx` is the request context. The handler returns a 302 immediately after, Echo cancels the request context, and the background DELETE is cancelled almost every time — so expired OIDC sessions are effectively never cleaned up and the `oidc_sessions` table grows without bound on an OIDC-heavy deployment.

**Fix:** detach the cleanup from the request lifetime — `go s.db.DeleteExpiredOIDCSessions(context.WithoutCancel(ctx))`. One line.

The other goroutines in the API (search-index rebuild, server start) don't use a request context and are fine.

Found in a codebase review; not user-reported.

unidoc-ahall:
Resolved already in #117 .

PR #117, opened 2026-07-03
fix: OIDC cleanup context, renderMention dedup, e2e flakiness (#114 #116)

unidoc-ahall:
## Summary

- **#114** — `DeleteExpiredOIDCSessions` goroutine was using the request context and getting cancelled when the handler returned. Fixed with `context.Background()`.
- **#116** — `renderMention` was copy-pasted in `CommentsPanel.vue` and `EntitySidebar.vue`. Extracted to `composables/useMention.js`.
- **test fix** — `test_suggestion_accept_ui` was flaky under load (10 s timeout too tight). Added `networkidle` wait + 20 s timeout.

## Test plan
- [ ] `just test` green (full suite incl. e2e)
- [ ] OIDC login flow works end-to-end
- [ ] Mentions render correctly in entity sidebar and comments panel


unidoc-alip (changes requested):
Small, well-scoped fix covering three independent issues (#114 OIDC cleanup context, #116 renderMention dedup, and an e2e flake). Verified locally end-to-end: the detached goroutine now survives request completion, and mention rendering (including HTML-escaping of injected `<script>`/`<b>` tags) works correctly in the browser through the shared composable. A few non-blocking points below worth a look before merge.

---

### 1. [should-fix] `internal/isms/api/api_oidc.go` line 124

`context.Background()` has no deadline, so if the DB is slow/overloaded this fire-and-forget `DELETE` can now hang indefinitely instead of being bounded by the (already-broken) request context. Every `/auth/oidc/authorize` call spawns one of these goroutines — under sustained DB slowness they'll pile up unbounded.

Confirmed by reproduction: holding all connections in a pool and firing 50 of these goroutines leaves all 50 permanently blocked in `Exec`, since nothing bounds the wait. Linked issue #114 suggested `context.WithoutCancel(ctx)` instead, which also has no deadline but at least preserves request-scoped values (trace/logger) — this PR's `context.Background()` diverges from that and drops those values too.

Suggested fix:
```go
// internal/isms/api/api_oidc.go:124
go func() {
    cleanupCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
    defer cancel()
    _ = s.db.DeleteExpiredOIDCSessions(cleanupCtx)
}()
```
This bounds the worst case to 5s per goroutine regardless of DB health, while still surviving request-context cancellation (the original bug in #114). If preserving request-scoped values matters, wrap with `context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)` instead of `context.Background()`.

---

### 2. [should-fix] `web/src/components/EntitySidebar.vue` line 72

`EntitySidebar.vue` doesn't appear to be imported/rendered anywhere in `web/src/` (confirmed via `git log` — it's only ever touched in the initial scaffold commit and this PR). The dedup here is correct but has no live surface to exercise; worth either wiring the component up or removing it as dead code in a follow-up, so "copy-pasted in two places" doesn't silently become "copy-pasted in one live place and one dead file."

Suggested fix — two options, either resolves the ambiguity:
1. **Remove it** — delete `web/src/components/EntitySidebar.vue` if it was scaffolded speculatively and superseded by `CommentsPanel.vue` handling the same tabs.
2. **Wire it up** — import and mount it from its intended parent, then add a test that exercises its comment-rendering path so the #116 fix has real coverage there too.

Recommend (1) unless there's a concrete near-term consumer planned, to avoid maintaining a second identical mention-rendering call site with no test coverage.

---

### 3. [nit] `tests/test_e2e_browser.py` line 438

`wait_for_load_state("networkidle")` is a reasonable pragmatic fix for the immediate flake, but Playwright's own docs discourage `networkidle` for general-purpose waiting since apps with any background polling/analytics never go fully idle. Worth double-checking this doesn't reintroduce flakiness (or slow the suite) if any such background activity gets added near this flow later; an explicit wait on the response/toast element would be more robust long-term.

Suggested fix:
```python
# tests/test_e2e_browser.py:438 — instead of:
page.wait_for_load_state("networkidle")
page.locator("text=Accepted").first.wait_for(state="visible", timeout=20000)

# prefer relying solely on the element-level wait, which already tolerates
# slow server round-trips without depending on global network quiescence:
page.locator("text=Accepted").first.wait_for(state="visible", timeout=20000)
```
No change needed urgently since nothing currently polls on this page — safe to leave as-is and revisit if/when background polling (websocket, presence heartbeat, notification polling) is added near this flow.

unidoc-ahall:
Thanks — all three addressed:

1. **OIDC cleanup timeout** — bounded the detached cleanup with `context.WithTimeout(context.Background(), 5*time.Second)`, so it still survives request completion (#114) but a slow/overloaded DB can't leave these goroutines blocked in `Exec` indefinitely. Good catch on the pile-up.
2. **EntitySidebar.vue** — removed it. Verified it's imported nowhere in `web/src` (and referenced in no test/doc), so it was dead code; deleting it beats maintaining a second identical mention call site with no coverage.
3. **networkidle** — dropped it; relying on the 20s element-level `wait_for` alone, which already tolerates a slow round-trip without depending on global network quiescence.

unidoc-alip (approved):
looks good to me

Issue #115, opened 2026-07-03
Changelog/audit-trail write errors are silently swallowed in ~54 handlers


unidoc-ahall:
Across the API (`api_incidents.go`, `api_corrective.go`, and others — ~54 sites) changelog writes are best-effort with the error discarded:

```go
_ = s.db.LogChange(ctx, orgID, &db.ChangelogEntry{...})
```

Meanwhile the suggestions path correctly propagates `LogChangeTx` errors inside its transaction. For a product whose pitch includes a complete audit trail, silently dropping a governance entry is the worst failure mode: the entity mutation succeeds, the trail has a hole, and nobody knows.

**Fix (minimal):** keep best-effort semantics but log every failure to stderr with org/entity context, so a systemic failure (DB pressure, constraint bug) is visible instead of invisible. A small helper, e.g. `s.logChangeBestEffort(...)`, avoids touching 54 call sites' control flow while centralising the logging.

Whether any of these sites should instead fail the request (transactional, like suggestions) can be decided case by case later — this issue is only about making the failures observable.

Found in a codebase review; not user-reported.

unidoc-ahall:
Resolved already in #117 .

PR #118, opened 2026-07-03
fix: log changelog write errors instead of swallowing them (#115)

unidoc-ahall:
## Summary

54 callsites used `_ = s.db.LogChange(...)` / `_ = s.db.LogChanges(...)`, silently discarding any error from audit-trail writes.

Adds `logChange` and `logChanges` helpers on `Server` that log errors with `log.Printf` and return — changelog writes are best-effort and must not block mutations, but errors should be visible in server logs.

All 54 callsites replaced across 8 files.

## Test plan
- [ ] `just test` green
- [ ] `go build ./...` clean

### DIFF
--- a/internal/isms/api/api_audit.go
+++ b/internal/isms/api/api_audit.go
@@ -288,7 +288,7 @@ func (s *Server) handleCreateAudit(c echo.Context) error {
 	if err != nil {
 		return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
 	}
-	_ = s.db.LogChange(ctx, orgID, &db.ChangelogEntry{
+	s.logChange(ctx, orgID, &db.ChangelogEntry{
 		EntityType: "audit",
 		EntityID:   int64(out.ID),
 		Action:     "create",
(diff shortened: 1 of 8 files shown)


unidoc-alip (changes requested):
Good fix for the silent-error-swallowing problem: all 54 callsites converted cleanly, argument lists verified identical to the originals, `go build` is clean, and live testing confirms changelog writes are correctly best-effort (the request still succeeds even when the changelog write fails). One bug in the new helper itself undermines the PR's own goal, though — see below.

---

### 1. [must-fix] `internal/isms/api/api_collab.go` line 66

`logChange`'s `log.Printf` uses `%s` for `entry.EntityID`, but `entry.EntityID` is `int64` (see `db.ChangelogEntry` in `internal/isms/db/changelog.go:15`), not a string. `go vet ./internal/isms/api/...` flags it directly: `format %s has arg entry.EntityID of wrong type int64`.

Confirmed live: forced a real changelog write failure locally (dropped the target table) and the emitted log line read:

```
changelog write error (org 1, entity risk %!s(int64=2)): ERROR: relation "entity_changelog" does not exist (SQLSTATE 42P01)
```

instead of `entity risk 2`. This defeats the PR's stated purpose — the entity ID, the single most useful piece of debugging context, is corrupted on every logged occurrence, across all ~30 call sites now routed through this helper.

Fix:

```go
log.Printf("changelog write error (org %d, entity %s %d): %v", orgID, entry.EntityType, entry.EntityID, err)
```

(`logChanges`'s format string is unaffected — it already uses `%d` correctly for its entry count.)

---

unidoc-ahall:
Fixed — `logChange` now uses `%d` for `entry.EntityID` (int64):

```go
log.Printf("changelog write error (org %d, entity %s %d): %v", orgID, entry.EntityType, entry.EntityID, err)
```

Good catch — that corrupted the entity ID on every logged failure, defeating the whole point. `go vet ./internal/...` is clean now and `go test ./internal/...` passes locally (the go-unit job should go green on the amended commit). `logChanges` was already correct.

unidoc-alip (approved):
looks good to me

PR #120, opened 2026-07-04
fix: self-hosted apex org resolution + operator diagnostics (#119)

### LINKED ISSUE #119: Self-hosted operability: apex org resolution, single-org auto-detect, auth/email diagnostics
Self-hosted deployments whose apex is not `isms.sh` (e.g. `isms.stsplatform.com`) had a broken login/org experience. This fixes it and adds the operator diagnostics that made it debuggable.

**Apex org resolution**
- The SPA classified its own host at module-import time (`router.js`, `App.vue`) before `/config` called `setApexHost()`, so a self-hosted apex was misread as tenant `isms` → `/login` 404 on `/auth/oidc/providers?org=isms`. Fix: the server injects its apex host into `index.html` as a `<meta name="isms-apex">` tag (a meta tag, not inline `<script>`, because CSP is `script-src 'self'`); `useCurrentOrg.js` seeds `_apexHosts` from it synchronously.
- Single-org deployments still prompted for an org. `GET /config` now resolves the apex to the sole org when exactly one exists → the login page auto-selects it. Multi-tenant apexes unaffected.

**Operator diagnostics (CLI)**
- `user test-auth` — run login's credential checks read-only against this box's `DATABASE_URL`.
- `user reset-otp` — clear a user's OTP to re-enroll (needed when an OTP secret was encrypted with a different `ISMS_SECRET`, e.g. a DB copied to a new install; symptom was "invalid credentials" with no OTP prompt).
- `test-email --org <slug>` — send with the org's branded From, mirroring real tenant mail.

Config is unchanged.

unidoc-ahall:
Fixes #119

## Apex org resolution
- **Boot-race:** server injects its apex host into `index.html` as `<meta name="isms-apex">`; `useCurrentOrg.js` seeds `_apexHosts` synchronously at module load (before `/config`). Meta tag, not inline `<script>`, because CSP is `script-src 'self'`. Fixes a self-hosted apex being misread as tenant `isms`.
- **Single-org:** `GET /config` resolves the apex to the sole org when exactly one exists → login auto-selects it. Multi-tenant unaffected.

## Operator diagnostics (CLI)
- `user test-auth` — read-only login credential check against this box's DB.
- `user reset-otp` — clear OTP to re-enroll (ISMS_SECRET mismatch on a copied DB).
- `test-email --org` — branded-From preview + send.

No config change.

## Test plan
- [ ] `just build-web && just test-restart` then `just test tests/test_routing.py -v`
- [ ] On isms.stsplatform.com: apex auto-selects the single org, no `org=isms` 404

### DIFF
--- a/cmd/isms/test_email.go
+++ b/cmd/isms/test_email.go
@@ -11,12 +11,18 @@ import (
 // testEmailCmd sends a test email using the current SMTP_* env vars, so an
 // operator can verify mail delivery without going through the signup flow.
 func testEmailCmd() *cobra.Command {
-	return &cobra.Command{
+	var orgSlug string
+	cmd := &cobra.Command{
 		Use:   "test-email <recipient>",
 		Short: "Send a test email to verify SMTP configuration",
 		Long: `Send a test email to the given recipient using the SMTP_* environment
 variables. Useful for confirming Postmark / SendGrid / SES credentials are correct
 before relying on signup verification or review-notification emails.
+
+With --org <slug>, the email is sent the way a tenant's real mail is: the org's
+name becomes the From display name (e.g. "STS" <noreply@host>) while the envelope
+sender stays SMTP_FROM, so SPF/DKIM alignment is preserved. Without --org, the
+raw SMTP_FROM is used as-is. --org needs DATABASE_URL set.
 
 Required env vars:
   SMTP_HOST       (e.g. smtp.postmarkapp.com)
@@ -43,6 +49,28 @@ Required env vars:
 				return fmt.Errorf("mailer not configured — set SMTP_HOST and SMTP_FROM at minimum")
 			}
 
+			// Resolve org branding when --org is given. SendBranded/PreviewFrom
+			// only consume Branding.Name (the From display name) — the generic
+			// test body isn't a colored template — so we set just the name.
+			var branding mail.Branding
+			if orgSlug != "" {
+				d, err := connectDB()
+				if err != nil {
+					return err
+				}
+				defer d.Close()
+				ctx := cmd.Context()
+				org, err := d.GetOrganizationBySlug(ctx, orgSlug)
+				if err != nil || org == nil {
+					return fmt.Errorf("organization %q not found", orgSlug)
+				}
+				branding.Name = org.Name
+				header, envelope := mail.PreviewFrom(cfg.From, branding.Name)
+				fmt.Printf("Org context: %s (%s)\n", org.Name, org.Slug)
+				fmt.Printf("From header: %s\n", header)
+				fmt.Printf("Envelope:    %s\n", envelope)
+			}
+
 			subject := "ISMS test email"
 			body := `<div style="font-family: sans-serif; max-width: 500px; margin: 0 auto;">
 <h2>SMTP test successful</h2>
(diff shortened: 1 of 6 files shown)


unidoc-alip (commented):
Solid fix for issue #119 — the boot-race root cause (SPA classifying its own host before `/config` resolves) is correctly diagnosed and fixed with a CSP-safe `<meta>` injection, the single-org auto-detect is narrowly scoped, and the new `test-auth`/`reset-otp`/`test-email --org` diagnostics mirror the real login/mail code paths well. Two should-fix items below, both verified with a PoC: the apex host is interpolated into the `<meta>` tag using Go's `%q` (string escaping, not HTML-attribute escaping), and `test-email --org` fetches `branding_color` but it's dead code since `SendBranded` only consumes the brand name. Neither blocks merge; both are one-line fixes.

---

### 1. [should-fix] `internal/isms/api/server.go` line 853

`%q` here is Go string-escaping, not HTML-attribute escaping. Verified with a PoC: `fmt.Sprintf("content=%q", ...)` on a value containing `"` produces `content="evil.com\"><script>alert(1)</script>"` — HTML doesn't treat `\` as an escape character, so a browser closes the attribute at the first literal `"` right after the backslash, then `>` closes the `<meta>` tag early and `<script>alert(1)</script>` becomes live markup in `<head>`.

Exploitability today is low — `s.apexHost` comes from `ISMS_DOMAIN`/`extractHostname(ISMS_BASE_URL)`, both operator-set env vars at boot, not per-request input — but the escaping primitive is simply wrong for this context and would silently become exploitable if apex derivation ever incorporates less-trusted input.

Suggested fix:
```go
import "html"
...
inject := fmt.Sprintf(`<meta name="isms-apex" content="%s"></head>`, html.EscapeString(s.apexHost))
```

---

### 2. [should-fix] `cmd/isms/test_email.go` line 46

`branding.Color` is populated from the org's `branding_color` setting here, but `mail.SendBranded`/`mail.PreviewFrom` (via `resolveFrom`) only ever read `Branding.Name` — confirmed by tracing the call chain: `resolveFrom(configFrom, fromName string)` takes just a name, `PreviewFrom` forwards to it, and `SendBranded` calls `sendAs(to, subject, body, b.name())`. `Color` never crosses these signatures. So this diagnostic does a DB round-trip for `branding_color` and prints/sends nothing that reflects it — `--org` doesn't preview the color-branded look that `SendVerificationBranded`/`SendPasswordResetBranded`/etc. actually use (those do call `b.color()`).

Suggested fix: either drop the unused `GetOrgSetting(ctx, org.ID, "branding_color")` block since it has no effect on this command's output, or, for a faithful preview, render the diagnostic body with the same colored HTML template the other `*Branded` sends use (via `b.color()`) so `--org` actually demonstrates the branded color, not just the From name.

---

### 3. [nit] `cmd/isms/user.go` line 121

`reset-otp` clears 2FA for an account with no confirmation prompt. Checked sibling commands in the same file (`user create`, `user set-password`) — none of them prompt for confirmation either, so this is consistent with the file's existing convention rather than a deviation. Raising only as an optional hardening suggestion: since this disables a security control (2FA) rather than just resetting a credential, a `--yes` flag or interactive y/N prompt would be cheap insurance against a fat-fingered `--email`.

Suggested fix (optional, not required for merge given existing convention):
```go
if !yes {
    fmt.Printf("Clear OTP for %s? [y/N] ", user.Email)
    var resp string
    fmt.Scanln(&resp)
    if resp != "y" && resp != "Y" {
        return fmt.Errorf("aborted")
    }
}
```

unidoc-ahall:
Thanks — all three addressed:

1. **meta escaping** — switched from `%q` to `html.EscapeString(s.apexHost)` with an explicit `content="%s"`, so a stray quote can't break out of the attribute. Right primitive for the HTML-attribute context, even though `apexHost` is operator-set today.
2. **dead branding_color** — dropped the `GetOrgSetting(..., "branding_color")` fetch in `test-email --org`; `SendBranded`/`PreviewFrom` only consume `Branding.Name`, so it did a DB round-trip for nothing. The command previews the From name (which is what `SendBranded` actually applies to this generic body).
3. **reset-otp confirm** — added a `[y/N]` prompt before clearing 2FA, skippable with `--yes` for scripts. It disables a security control, so worth the guard.

`just test-go` clean.

unidoc-alip (approved):
looks good to me

PR #121, opened 2026-07-06
feat(cli): change-management commands + CLI usage guide (#55, #57)

### LINKED ISSUE #55: CLI: add change management commands + map remaining module coverage
## Summary

The CLI has no change-management commands, and several other modules have no CLI surface. To make the CLI genuinely useful as an API mirror, add change management and map the remaining coverage gaps.

## Scope

- Add `isms change` commands mirroring the API: list / show / create / update / status (`/changes`, `/changes/:id`, `/changes/:id/status`).
- Produce a short **module coverage list**: which entities/API areas the CLI does NOT cover (tasks, suggestions / entity-comments / references, admin plane, entity show/stats, …), so we can fill per use case in follow-ups.

## Acceptance criteria

- `isms change list/show/create/update/status` work against the API (RBAC etc. enforced server-side, like the other CLI commands).
- A documented list of remaining CLI coverage gaps.

## Notes

Part of completing the CLI-as-API-mirror (CLI/TUI scope). The CLI already talks to the API correctly — this fills missing surface. From the CLI gap analysis.

### LINKED ISSUE #57: Write a CLI usage guide (philosophy + clone→sync workflow + command reference)
## Summary

There is no dedicated CLI usage guide. CLI mentions are scattered across operator-guide / auditor-guide / README config tables. Write a `docs/cli-guide.md` that documents both the **philosophy** and the **how-to**.

## Philosophy (decided — CLI/TUI scope)

- **CLI = a thin, complete client over the HTTP API** — the same model as `gh` for GitHub. Use it directly for scripting/automation, or let an AI agent drive it (contributor access). We complete the API mirror incrementally, per real use case (not all at once).
- **TUI = a read-only reader, for now** — browse and read documents, and optionally view registers (risks, etc.) read-only. No editing. Nothing more is planned for now.
- Anything beyond that (mutation at scale, the long tail) is the web UI or an AI agent's job.

## How-to (the guide)

- The core workflow: `clone` → edit locally → `sync` (auth/env setup).
- A command reference grouped by entity.
- Shell completion (#54).

## Acceptance criteria

- `docs/cli-guide.md` exists, opens with the philosophy above, then the clone→sync workflow + command reference.
- Documents only what actually works (don't document half-wired commands).

## Notes

Best written/finished alongside the CLI coverage work (#55) so docs match reality. This replaces the need for a separate CLI/TUI scope discussion — the philosophy is settled and lives here.

unidoc-ahall:
Closes #55
Closes #57

## #55 — CLI change management
`isms change list|show|create|status`, wrapping the existing change-request API. Adds `CreateChange`/`UpdateChangeStatus` client methods (`ListChanges`/`GetChange` already existed) and registers the command. Coverage map for the remaining modules is in the guide.

## #57 — CLI usage guide
`docs/cli-guide.md`: leads with *why* the CLI exists (git-backed docs, bulk/raw-markdown via `clone`, managers/automation; a `sync` writes a **draft** — governance is never bypassed), the clone→edit→sync workflow, and a grouped command reference.

## Test plan
- [ ] `just test-go` green
- [ ] smoke: `isms change create --title …`, `isms change list`, `isms change status <id> approved` against the stack

### DIFF
--- a/cmd/isms/change.go
+++ b/cmd/isms/change.go
@@ -1,0 +1,224 @@
+package main
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"isms.sh/internal/isms/client"
+	"isms.sh/internal/isms/db"
+)
+
+func changeCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "change",
+		Short: "Manage change requests",
+	}
+	cmd.AddCommand(changeListCmd(), changeShowCmd(), changeCreateCmd(), changeUpdateCmd(), changeStatusCmd())
+	return cmd
+}
+
+// resolveChangeID accepts either a numeric database id or a per-org identifier
+// (e.g. CR-1) and returns the numeric id. Identifiers are resolved via the list,
+// since identifier and id genuinely diverge (identifier is a per-org sequence).
+func resolveChangeID(c *client.Client, arg string) (int, error) {
+	if id, err := strconv.Atoi(arg); err == nil {
+		return id, nil
+	}
+	changes, _, err := c.ListChanges("")
+	if err != nil {
+		return 0, err
+	}
+	for _, ch := range changes {
+		if strings.EqualFold(ch.Identifier, arg) {
+			return ch.ID, nil
+		}
+	}
+	return 0, fmt.Errorf("no change request with id or identifier %q", arg)
+}
+
+func changeListCmd() *cobra.Command {
+	var status string
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List change requests",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := requireAPI()
+			changes, total, err := c.ListChanges(status)
+			if err != nil {
+				return err
+			}
+			fmt.Printf("%-5s %-8s %-32s %-9s %-8s %-12s %-14s\n",
+				"ID", "REF", "TITLE", "PRIORITY", "RISK", "STATUS", "ASSIGNED")
+			fmt.Println(repeat("-", 96))
+			for _, ch := range changes {
+				fmt.Printf("%-5d %-8s %-32s %-9s %-8s %-12s %-14s\n",
+					ch.ID, ch.Identifier, truncate(ch.Title, 32),
+					truncate(ch.Priority, 9), truncate(ch.RiskLevel, 8),
+					truncate(ch.Status, 12), truncate(ch.AssignedTo, 14))
+			}
+			if total > len(changes) {
+				fmt.Printf("\n%d of %d change requests (raise the server page size to see all)\n", len(changes), total)
+			} else {
+				fmt.Printf("\n%d change requests\n", len(changes))
+			}
+			return nil
+		},
+	}
+	cmd.Flags().StringVar(&status, "status", "", "Filter by status ("+strings.Join(db.ChangeStatuses, ", ")+")")
+	return cmd
+}
+
+func changeShowCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "show <id|ref>",
+		Short: "Show a change request in detail (numeric id or CR- reference)",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := requireAPI()
+			id, err := resolveChangeID(c, args[0])
+			if err != nil {
+				return err
+			}
+			ch, err := c.GetChange(id)
+			if err != nil {
+				return err
+			}
+			fmt.Printf("%s  %s\n", ch.Identifier, ch.Title)
+			fmt.Println(repeat("-", 60))
+			fmt.Printf("Status:        %s\n", ch.Status)
+			fmt.Printf("Priority:      %s\n", ch.Priority)
+			fmt.Printf("Risk level:    %s\n", ch.RiskLevel)
+			fmt.Printf("Category:      %s\n", ch.Category)
+			fmt.Printf("Requested by:  %s\n", ch.RequestedBy)
+			if ch.AssignedTo != "" {
+				fmt.Printf("Assigned to:   %s\n", ch.AssignedTo)
+			}
+			if ch.ApprovedBy != "" {
+				fmt.Printf("Approved by:   %s\n", ch.ApprovedBy)
+			}
+			if ch.Description != "" {
+				fmt.Printf("\nDescription:\n%s\n", ch.Description)
+			}
+			if ch.Justification != "" {
+				fmt.Printf("\nJustification:\n%s\n", ch.Justification)
+			}
+			if ch.RollbackPlan != "" {
+				fmt.Printf("\nRollback plan:\n%s\n", ch.RollbackPlan)
+			}
+			if ch.Notes != "" {
+				fmt.Printf("\nNotes:\n%s\n", ch.Notes)
+			}
+			return nil
+		},
+	}
+}
+
+func changeCreateCmd() *cobra.Command {
+	var cr db.ChangeRequest
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create a change request (starts as 'proposed'; transition with 'change status')",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if cr.Title == "" {
+				return fmt.Errorf("--title is required")
+			}
+			c := requireAPI()
+			out, err := c.CreateChange(&cr)
+			if err != nil {
+				return err
+			}
+			fmt.Printf("Created %s (%s) — status %s\n", out.Identifier, out.Title, out.Status)
+			return nil
+		},
+	}
+	// No --status: a change is born "proposed" and transitioned via
+	// `change status`, so approved_by / the follow-up task / the changelog
+	// transition are never skipped.
+	cmd.Flags().StringVar(&cr.Title, "title", "", "Change title (required)")
+	cmd.Flags().StringVar(&cr.Description, "desc", "", "What is changing and why")
+	cmd.Flags().StringVar(&cr.Justification, "justification", "", "Business justification")
+	cmd.Flags().StringVar(&cr.Priority, "priority", "medium", "Priority ("+strings.Join(db.ChangePriorities, ", ")+")")
+	cmd.Flags().StringVar(&cr.Category, "category", "", "Category ("+strings.Join(db.ChangeCategories, ", ")+")")
+	cmd.Flags().StringVar(&cr.RiskLevel, "risk", "", "Risk level ("+strings.Join(db.ChangeRiskLevels, ", ")+")")
+	cmd.Flags().StringVar(&cr.RollbackPlan, "rollback", "", "Rollback plan")
+	cmd.Flags().StringVar(&cr.Notes, "notes", "", "Notes")
+	cmd.Flags().StringVar(&cr.AssignedTo, "assigned-to", "", "Assignee email")
+	return cmd
+}
+
+func changeUpdateCmd() *cobra.Command {
+	var title, desc, justification, priority, category, risk, rollback, notes, assignedTo string
+	cmd := &cobra.Command{
+		Use:   "update <id|ref>",
+		Short: "Update change request fields (only the flags you pass are changed)",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := requireAPI()
+			id, err := resolveChangeID(c, args[0])
+			if err != nil {
+				return err
+			}
+			// Send only changed fields — the server contract is nil = leave alone.
+			fields := map[string]interface{}{}
+			f := cmd.Flags()
+			for flag, val := range map[string]*string{
+				"title": &title, "desc": &desc, "justification": &justification,
+				"priority": &priority, "category": &category, "risk": &risk,
+				"rollback": &rollback, "notes": &notes, "assigned-to": &assignedTo,
+			} {
+				if f.Changed(flag) {
+					fields[changeFieldJSON[flag]] = *val
+				}
+			}
+			if len(fields) == 0 {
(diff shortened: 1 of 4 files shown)


unidoc-alip (changes requested):
The `isms change` commands are a clean, thin wrapper over the existing API, and RBAC, validation, and identity-stamping are all correctly left to the server (security review came back clean). However, the shipped surface has contract problems — every finding below was reproduced against a live local server (fresh DB, server built from this branch at b077433, two test orgs): the status vocabulary in the flag help is wrong (`draft`/`submitted` don't exist — the real enum is `proposed, approved, rejected, in_progress, implemented, closed`), the list output gives no way to obtain the numeric id that `show`/`status` require, and `docs/cli-guide.md`'s central governance claim ("a sync writes a draft; nothing goes live until approved") is contradicted by the code — a synced edit that flipped a document to `approved` and injected new body content was served immediately with zero rows in the `reviews` table. Issue #55's acceptance criteria (`update` subcommand, a list of coverage gaps) are also not fully met.

---

### 1. [must-fix] `docs/cli-guide.md` line 17

The "Governance is never bypassed… a `sync`/push writes a **draft**… Nothing you push goes live until it's approved" claim doesn't match the code. `isms sync` pushes to `refs/heads/main` (`cmd/isms/sync.go:262-269`; sync requires being on main), the server serves documents from that same ref (`store.LoadDocument` reads the bare repo's HEAD tree), and `api_git.go:125` invalidates the store cache right after a push so readers see the content immediately. The receive-pack hook only enforces manager/admin push rights, protection of `review/*` refs, fast-forward, and duplicate-id/path validation — there is no draft gate and nothing rewrites frontmatter `status`.

Verified live (fresh org, iso27001 scaffold, manager token):

```
$ isms clone && cd acme-isms
# edit 4.01-external-and-internal-issues.md offline:
#   status: "draft" → status: "approved", plus a new body line
$ git commit -am "edit approved doc offline" && isms sync
Sync complete.
$ isms document cat iso27001-4-1 | tail -1
INJECTED-WITHOUT-REVIEW: this line bypassed the review workflow entirely.
$ curl .../documents/iso27001-4-1/body → status: approved
$ psql: SELECT count(*) FROM reviews;  → 0
```

The pushed content — including the status flip to `approved` — was live instantly, with zero reviews in the database. `GET /documents/needs-review` merely flags it after the fact.

Suggested fix: since #57's acceptance criteria say "documents only what actually works", rewrite this paragraph (and the same claim in the clone→edit→sync section and the `isms sync` table row) to state the real behavior: pushes land on `main` and are live to readers immediately — which is why push is restricted to manager/admin — and documents changed since approval are surfaced via the needs-review view; frontmatter `status` is not enforced by the push path. Alternatively (larger scope, separate issue): add a server-side gate in `api_git.go` post-receive that rejects pushes modifying approved documents outside a review ref.

---

### 2. [must-fix] `cmd/isms/change.go` line 120

Two problems with `--status` on create, both reproduced live.

First, the help text is wrong — `draft` is not a valid change status and the default is `proposed`:

```
$ isms change create --title "Rotate TLS certs" --status draft
Error: API error 400: {"message":"invalid status: \"draft\"
  (allowed: [proposed approved rejected in_progress implemented closed])"}
$ isms change create --title "Rotate TLS certs"
Created CR-1 (Rotate TLS certs) — status proposed
```

Second, allowing an initial status lets a change be born approved, skipping everything `handleUpdateChangeStatus` does:

```
$ isms change create --title "Firewall rule change" --status approved
Created CR-2 (Firewall rule change) — status approved
```

DB after transitioning CR-1 properly (`isms change status 1 approved`) vs born-approved CR-2:

```
 identifier |  status  |  approved_by      tasks: "Implement CR-1: Rotate TLS certs" (only)
 CR-1       | approved | mgr@test.local    changelog: proposed → approved (CR-1 only)
 CR-2       | approved | (NULL)            CR-2: no task, no changelog transition, no approver
```

Suggested fix: drop the `--status` flag from `create` entirely — changes are created as `proposed` and transitioned with `isms change status <id> <status>`, which preserves the follow-up task, `approved_by`, and changelog invariants. If the flag must stay, fix the help to `Initial status: proposed, approved, rejected, in_progress, implemented, closed (default proposed)` and consider having the server route non-`proposed` initial statuses through the same side effects.

---

### 3. [must-fix] `cmd/isms/change.go` line 36

The "ID" column prints `ch.Identifier` (e.g. `CR-1`), but `change show <id>` and `change status <id>` require the numeric database id — and the two genuinely diverge, because `Identifier` is a per-org sequence while `id` is a global serial. Reproduced live with two orgs:

```
$ isms change show CR-1
Error: id must be a number
# org "beta"'s first change:
$ isms change list          → CR-1  Beta's first change
# but its numeric id is 3 (orgs share the serial):
$ isms change show 1        → Error: API error 404: change request not found
$ isms change show 3        → CR-1  Beta's first change   (works)
```

The numeric id appears nowhere in any CLI output, so `show`/`status` are unusable from `list` alone (`isms incident list` prints `inc.ID` for exactly this reason).

Suggested fix: add the numeric id as its own column, e.g.

```go
fmt.Printf("%-5s %-8s %-34s …\n", "ID", "REF", "TITLE", …)
fmt.Printf("%-5d %-8s %-34s …\n", ch.ID, ch.Identifier, truncate(ch.Title, 34), …)
```

or make `show`/`status` accept `CR-…` refs (resolve via `ListChanges` and match `Identifier` when `Atoi` fails).

---

### 4. [should-fix] `cmd/isms/change.go` line 44

Same enum problem as create: `draft` and `submitted` are not change statuses, and the real values `proposed`/`in_progress` are missing from the help. The list endpoint doesn't validate the filter, so the documented value silently returns nothing. Reproduced live:

```
$ isms change list --status draft
ID           TITLE     …
---------------------------------
0 change requests        # while 2 proposed/approved changes exist
```

(`--risk` on create has the same class of drift: it omits `critical`, which `db.ChangeRiskLevels` accepts.)

Suggested fix: derive the help strings from the exported enums so they can't drift:

```go
"Filter by status ("+strings.Join(db.ChangeStatuses, ", ")+")"
```

Same for `--risk` with `db.ChangeRiskLevels`.

---

### 5. [should-fix] `cmd/isms/change.go` line 16

Issue #55's acceptance criteria list `isms change list/show/create/update/status`, and the server already exposes the endpoint, but there's no `update` subcommand. Verified live:

```
$ isms change update 2 --title "New title"
Error: unknown flag: --title      # falls through to parent help
$ curl -X PUT …/v1/changes/2 -d '{"title":"Firewall rule change (edited via raw API)"}'
{"id":2, "identifier":"CR-2", "title":"Firewall rule change (edited via raw API)", …}   # 200 OK
```

So a typo'd title or wrong priority can't be fixed from the CLI (the status endpoint only transitions status), but the API supports it.

Suggested fix: add `changeUpdateCmd()` mirroring `isms corrective update` (cmd/isms/corrective.go): flags for title/description/justification/priority/category/risk/rollback/notes/assigned-to, sending only changed fields to `PUT /v1/changes/:id` via a new `client.UpdateChange`. The server's `changeUpdateRequest` uses nil-means-leave-alone pointers, so use `cmd.Flags().Changed(...)` to build the payload.

---

### 6. [should-fix] `internal/isms/client/client.go` line 595

`ChangeCreate` hand-copies `db.ChangeRequest`'s fields/JSON tags and has already drifted at birth: the server's `changeCreateRequest` (api_collab.go:2360) also accepts `planned_at` and `references`, which this struct omits. Verified live:

```
$ curl -X POST …/v1/changes -d '{"title":"Change with planned date","planned_at":1783468800}'
{'identifier': 'CR-3', 'title': 'Change with planned date', 'planned_at': 1783468800}   # 200 OK
$ isms change create --help | grep -c planned
0                                            # unreachable from the CLI
```

Every other create method in this file (`CreateIncident`, `CreateCorrectiveAction`, `CreateSystem`, `CreateLegal`, …) passes `*db.X` directly and lets the server bind what it knows.

Suggested fix: delete `ChangeCreate` and use the file's own convention:

```go
func (c *Client) CreateChange(cr *db.ChangeRequest) (*db.ChangeRequest, error) {
    data, err := c.post("/v1/changes", cr)
    …
}
```

`db.ChangeRequest` already carries `planned_at` with the right JSON tag, so a `--planned <date>` flag on create becomes possible for free.

---

### 7. [should-fix] `cmd/isms/change.go` line 40

`GET /v1/changes` is paginated with a default limit of 50 (api_collab.go:2415) and returns `{data, total, page, page_size}`, but `ListChanges` unwraps only `data`. Reproduced live with 58 changes in the org:

```
$ isms change list | tail -1
50 change requests            # DB: SELECT count(*) → 58
```

The newest 8 changes are silently invisible and the footer misstates the count as the page size.

Suggested fix: have `ListChanges` read the wrapper's `total` (and either request a high `limit` or loop pages):

```go
var wrapper struct {
    Data  []db.ChangeRequest `json:"data"`
    Total int                `json:"total"`
}
```

then print `fmt.Printf("\n%d of %d change requests\n", len(changes), total)` when truncated.

---

### 8. [should-fix] `cmd/isms/change.go` line 139

`handleUpdateChangeStatus` returns only `{"status": "..."}`, not a full change request, so `ch.Identifier` is always empty here. Reproduced live:

```
$ isms change status 1 approved
 → approved                      # literally nothing before the arrow
```

Suggested fix: print what the user already gave you, and stop pretending the client gets a full entity back:

```go
// client: func (c *Client) UpdateChangeStatus(id int, status string) (string, error)
//         unmarshal into struct{ Status string `json:"status"` }
fmt.Printf("Change #%d → %s\n", id, newStatus)
```

(or have the server return the updated change request like `handleUpdateChange` does — it already fetches `updated` before replying).

---

### 9. [should-fix] `docs/cli-guide.md` line 88

`isms review` has no `merge` subcommand — the command set is send, list, show, approve, assign, close (cmd/isms/review.go:18). Verified live: `isms review merge 1` falls through to the parent help instead of merging. Merging is only available via the web UI or the MCP `merge_review` tool, so a CLI-only approval pipeline can't actually be completed. #57's acceptance criteria say to document only what works.

Suggested fix: change the row to `Send for review, approve, close` and add a note that merging an approved review happens in the web UI (or via the MCP `merge_review` tool) — or add a `review merge` subcommand in a follow-up and keep the text.

---

### 10. [should-fix] `docs/cli-guide.md` line 123

Issue #55 asked for "a documented list of remaining CLI coverage **gaps**", but this section lists what is covered (redundant with the tables above) and asks the reader to discover gaps themselves. The gaps are real and enumerable — verified live:

```
$ isms task list        → Error: unknown command "task" for "isms"
$ isms suggestion list  → Error: unknown command "suggestion" for "isms"
```

Suggested fix: replace the closing sentence with the explicit gap list so follow-ups can be scoped, e.g.: "Not yet scriptable from the CLI: tasks, suggestions, entity comments & references, approval policies, notifications, evidence upload/download, entity stats, and change-request `update` (see #55). Server-side admin is CLI-only via `isms server …`." Adjust to match reality at merge time.

---

### 11. [nit] `cmd/isms/change.go` line 33

Sibling commands (incident.go, corrective.go, audit.go, legal.go, review.go) use `strings.Repeat` for these separators rather than the hand-rolled `repeat` helper (which is O(n²) string concat). Also, the columns above sum to 94 chars (12+34+9+8+12+14 + 5 spaces), so the 92-dash rule comes up 2 short — visible in every listing.

Suggested fix: `fmt.Println(strings.Repeat("-", 94))` (and `strings.Repeat("-", 60)` in `show`).

---

### 12. [nit] `cmd/isms/change.go` line 37

`truncate` slices by byte index (`s[:n-3]`), so multibyte titles get cut mid-rune and the byte-based `%-34s` padding misaligns the row. Reproduced live:

```
CR-60        情報セキュリティポリ�...                     medium    low   …
CR-59        Änderung der Zugriffsrichtlini...  medium    low   …
```

(note the replacement char and the shifted columns).

Suggested fix (minimum, rune-safe):

```go
func truncate(s string, n int) string {
    r := []rune(s)
    if len(r) <= n { return s }
    return string(r[:n-3]) + "..."
}
```

Pre-existing helper, but this PR is its first use on free-text titles in a fixed-width table; column alignment for wide (CJK) runes would need a display-width-aware pad, which can stay out of scope.

---

### 13. [nit] `cmd/isms/change.go` line 112

Sibling register commands spell this flag `--description` (e.g. `isms corrective`); verified live that the sibling spelling fails here (`Error: unknown flag: --description`), so scripts must special-case this one command.

Suggested fix: rename to `--description` (optionally keep `--desc` as an alias).

unidoc-ahall:
Thanks — thorough review, all eight addressed:

1. **Governance claim (must)** — you're right, the doc was wrong: `sync` is live-immediate, not a draft gate. Rewrote the 'Why a CLI' paragraph, the clone→sync section, and the `sync` table row to state the real behavior (pushes land on `main`, live to readers immediately, manager/admin-only; needs-review surfaces changes after the fact; `status` not enforced by the push path). **Flagged to ahall as a product decision** — accept live-push as the model, or add a post-receive gate for approved docs (separate issue).
2. **--status on create (must)** — dropped the flag entirely. Changes are born `proposed` and transitioned via `change status`, so approved_by / follow-up task / changelog transition can't be skipped.
3. **numeric id (must)** — `list` now prints ID + REF columns; `show`/`status` accept a numeric id OR a `CR-` ref (resolved via the list).
4. **enum help drift (should)** — status/priority/risk/category help is now derived from `db.ChangeStatuses`/etc.
5. **update subcommand (should)** — added `change update` (new `client.UpdateChange`, sends only `cmd.Flags().Changed` fields).
6. **ChangeCreate drift (should)** — deleted; `CreateChange` takes `*db.ChangeRequest` (planned_at/references bind for free).
7. **pagination (should)** — `ListChanges` reads `{data,total}`; list prints 'N of TOTAL' when truncated.
8. **status output (should)** — prints `Change #<id> → <status>`.

`just test-go` green. Coverage-gap table added to the guide for #55's acceptance.

unidoc-alip (approved):
looks good to me

PR #122, opened 2026-07-06
feat(review): @-mention members in entity comments + change requests (#4)

### LINKED ISSUE #4: Tag members in review comments, rollback plan, and change requests (with notifications)

unidoc-ahall:
Closes #4

Slice 2 (slice 1 = review comments, merged in #110). Mentions + notifications now work across the app, not just reviews.

## What
- `notifyReviewMentions` generalized to `notifyMentions(…, title, link)`; review path unchanged.
- `handleCreateEntityComment` notifies @-mentions on **any** entity comment (change requests + all registers), linked via `entityLink(type,id)` (mirrors `router.js`).
- Change-request **description + rollback plan** notify @-mentions **on create**.

## Scope note
Field mentions (description/rollback) fire on **create only** — re-scanning a mutable field on every edit would re-notify the same people. Ongoing collaboration uses comments, which are append-only and notify once. If edit-time field mentions are wanted, that's a separate follow-up (needs new-vs-old diffing).

## Test plan
- [ ] `just test-go` green
- [ ] `just test tests/test_review_mentions.py` (entity-comment mention notification)

### DIFF
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -94,10 +94,11 @@ func docLink(docID string) string {
 // email local-part (maria@… → @maria) or their name with spaces removed.
 var mentionRe = regexp.MustCompile(`@([a-zA-Z0-9._-]+)`)
 
-// notifyReviewMentions creates an in-app notification for each org member
-// @-mentioned in a review comment (#4), excluding the comment's author. Matching
-// is case-insensitive against the member's email local-part or name-slug.
-func (s *Server) notifyReviewMentions(ctx context.Context, orgID, reviewID int, actor, body string) {
+// notifyMentions creates an in-app notification for each org member @-mentioned
+// in body (excluding the actor), with the given title and link. Matching is
+// case-insensitive against the member's email local-part or name-slug. Generic
+// across surfaces — review comments, entity comments, change fields (#4).
+func (s *Server) notifyMentions(ctx context.Context, orgID int, actor, body, title, link string) {
 	matches := mentionRe.FindAllStringSubmatchIndex(body, -1)
 	if len(matches) == 0 {
 		return
@@ -124,8 +125,6 @@ func (s *Server) notifyReviewMentions(ctx context.Context, orgID, reviewID int, 
 		}
 	}
 	snippet := truncateStr(body, 200)
-	title := fmt.Sprintf("%s mentioned you in a review comment", actor)
-	link := fmt.Sprintf("/reviews/%d", reviewID)
 	notified := map[string]bool{}
 	for _, m := range matches {
 		// Skip "@scope/pkg" tokens (e.g. @babel/core): a '/' right after the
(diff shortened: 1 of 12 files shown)


unidoc-alip (changes requested):
Solid slice 2: the `notifyMentions` generalization is clean, the review path is preserved verbatim, and the create-only scope for change-request fields is well reasoned and documented. One blocking issue, reproduced live on this branch: entity-comment mention links are built from the entity **identifier** (what the web sends as `entity_id`, e.g. `CR-1`), but every SPA detail route resolves `parseInt(route.params.id)` — so the deep link lands on the register list without opening the entity, across all main registers. The new test asserts this broken format, cementing it. Also `entityLink` misses commentable types (`document`, `task`), producing `link: null` notifications (also reproduced live). All three findings below were verified against a live server built from commit a63d33f.

---

### 1. [must-fix] `internal/isms/api/api_entity_comments.go` line 73

Entity-comment mention links are dead deep-links for every main register. The web sends the entity **identifier** as `entity_id` (`CommentsPanel :entityId="selectedChange.identifier"` — same in Risks/Suppliers/Assets/Systems/Incidents/CorrectiveActions/Legal/Tasks), so this builds e.g. `/changes/CR-1`. But every SPA route opener does `parseInt(route.params.id)` (`openChangeFromRoute`, `openRiskFromRoute`, …), so `parseInt("CR-1")` → `NaN`, the lookup fails silently, and the notification lands on the register list without opening the entity.

Reproduced live (local server on this branch):

```
created change request: id=1 identifier=CR-1
POST /entity-comments {entity_type: change_request, entity_id: "CR-1",
                       body: "@testreviewer please review the rollback step"}
→ notification link: '/changes/CR-1'
→ Changes.vue openChangeFromRoute: parseInt('CR-1') = NaN
  → changes.find(c => c.id === NaN) never matches → entity never opens
contrast (create-path in this same PR): link='/changes/2' (numeric cr.ID → resolves fine)
```

Suggested fix (web-side — one pattern fixes all 8 registers at once, and also covers the `task` case in finding 2): make each `openXFromRoute` fall back to an identifier match when `parseInt` yields `NaN`. E.g. in `Changes.vue`:

```js
async function openChangeFromRoute(id) {
  const numId = parseInt(id)
  let cr = changes.value.find(c => c.id === numId || c.identifier === id)
  if (!cr && !isNaN(numId)) {
    try { cr = await api.fetchJSON(`/api/v1/changes/${numId}`) } catch { return }
  }
  ...
```

The list is already loaded before `openXFromRoute` runs in every view's `onMounted`, so the in-memory identifier match is sufficient; skip the fetch fallback for non-numeric ids since `GET /changes/CR-1` would 4xx anyway.

Alternative (server-side): resolve identifier → numeric ID in `handleCreateEntityComment` before building the link — but that needs a per-entity-type lookup and touches more code for the same result.

Either way, `test_entity_comment_mention_notifies_the_member` currently asserts the broken format (`/changes/{identifier}`); if you take the web-side fix the backend format stays and the test stands, but add an e2e or at least a comment noting the router-side fallback is load-bearing.

---

### 2. [should-fix] `internal/isms/api/api_collab.go` line 159

Two commentable surfaces with real routes are missing from the `entityLink` map, so their mention notifications get an empty link:

- `document` — Documents.vue mounts `CommentsPanel entityType="document"`; the existing `docLink()` helper (same file, ~line 80) already produces the right `/documents?doc=<id>` route.
- `task` — Tasks.vue mounts `CommentsPanel entityType="task"` and `/tasks/:id` exists in router.js.

Reproduced live: mentions in comments on `task` and `document` entities deliver the notification but with `link: null`:

```
{'title': '… mentioned you in a comment', 'body': '@testreviewer poc2 task',     'link': None}
{'title': '… mentioned you in a comment', 'body': '@testreviewer poc2 document', 'link': None}
```

Suggested fix:

```go
func entityLink(entityType, entityID string) string {
    if entityID == "" {
        return ""
    }
    if entityType == "document" {
        return docLink(entityID) // documents resolve by doc_id, not a path param
    }
    prefix := map[string]string{
        "risk":              "/risks/",
        "supplier":          "/suppliers/",
        "asset":             "/assets/",
        "system":            "/systems/",
        "legal_requirement": "/legal/",
        "incident":          "/incidents/",
        "change_request":    "/changes/",
        "corrective_action": "/corrective-actions/",
        "task":              "/tasks/",
    }[entityType]
    // objective / audit* use tab-nested routes with no flat deep-link — deliberately unlinked.
    if prefix == "" {
        return ""
    }
    return prefix + entityID
}
```

Note: the `task` link inherits the identifier-vs-id issue from finding 1 (Tasks.vue passes `identifier`, `openTaskFromRoute` parseInts) — the web-side fallback suggested there covers it.

---

### 3. [nit] `internal/isms/api/api_collab.go` line 2515

Justification and Notes are also free-text fields where a mention is plausible ("per @maria's risk assessment…"), and issue #4 asks for tagging "anywhere a member should be pulled in". Verified live: a change request created with `justification: "per @testreviewer's risk assessment"` and `notes: "cc @testreviewer"` produces no notification.

Suggested fix — either scan all four free-text fields:

```go
s.notifyMentions(ctx, orgID, cr.RequestedBy,
    strings.Join([]string{cr.Description, cr.Justification, cr.Notes, cr.RollbackPlan}, "\n"),
    ...)
```

or, if description+rollback is the intended scope, say so in the comment above the call so the field choice reads as deliberate rather than accidental.

unidoc-ahall:
Thanks — sharp review, all three fixed:

1. **Dead identifier deep-links (must-fix)** — took the web-side fix: `openXFromRoute` in all 9 register views now matches `x.id === numId || x.identifier === id`, so an identifier link (`/changes/CR-1`) opens the entity. Numeric links (the create-path) still resolve. Backend link format unchanged; the test now documents that the router fallback is load-bearing.
2. **Missing types (should-fix)** — `entityLink` now handles `document` (via the existing `docLink` → `/documents?doc=…`) and `task` (`/tasks/`). objective/audit stay unlinked (tab-nested routes), noted in a comment.
3. **Free-text fields (nit)** — change-create now scans description + justification + notes + rollback plan.

`just test-go` green; `just build-web` clean.

unidoc-alip (approved):
looks good to me

Issue #26, opened 2026-06-07
Epic: Foundational integrity — one enforced write path per entity


unidoc-ahall:
## Summary

Today business rules (open-CA guards, lifecycle timestamps, status RBAC,
notifications) live in the HTTP handler layer (`api_*.go`). The suggestion-apply
path (`api_suggestions.go`, runs in `WithOrgTx`) calls the DB-layer `*Tx`
functions directly and never executes those handler rules. Two divergent write
paths → the whole integrity-gap cluster, candidate 3, and the status-parity bugs
all share this root.

Establish **one validated, transaction-aware mutation path per entity** that
every writer — HTTP handlers, suggestion-apply, CLI, MCP — calls. Rules can't
diverge because there's one place that enforces them.

## Why now / 0.7.0

#23 (contributors propose-only) makes suggestion-apply the **primary write path
for everyone except managers**. Shipping #23 on the current divergent path would
route the main input path through the *less* enforced code. So this is the
foundation #23 stands on — it must land first or together, in 0.7.0.

## Approach — incremental, not big-bang

1. Establish the unified path on **one entity** (incident or CA), with full
   tests, proving the pattern (handler + apply both call the same enforced
   function inside the transaction).
2. Migrate the remaining entities to the pattern one at a time.

Do NOT convert all entities at once.

## Absorbs (these are symptoms of the divergent path, fixed by the unified one)

- suggestion-apply bypasses the incident open-CA guard + lifecycle timestamps
- suggestion-apply for CAs skips resolved_at/resolved_by_id
- CA created via suggestion defaults to the wrong starting state

## Acceptance criteria

- [ ] One enforced mutation path per entity; HTTP handler and suggestion-apply
      both go through it.
- [ ] The three absorbed bypasses are closed by construction, with tests.
- [ ] Tests assert identical behavior (rules, timestamps, RBAC) across the HTTP
      and suggestion-apply paths for the migrated entities.
- [ ] Pattern documented so new entities follow it.

unidoc-ahall:
Epic complete — one enforced write path per entity, shared by the HTTP and
  suggestion-apply handlers with parity tests. Slices landed: incidents + corrective actions (#123),
  asset/supplier/legal/system/objective create consistency (#140), audit_finding status update (#142), change_request status (#144).
  Closing.

PR #123, opened 2026-07-06
feat(integrity): one enforced write path — incident + corrective action (#26)

unidoc-ahall:
Advances #26 — establishes the enforced-write-path pattern and closes all three named integrity bypasses. Epic stays open for any remaining entities (none with known divergence today); close it if the named scope is considered complete.

## Problem
Business rules lived only in HTTP handlers; suggestion-apply called `db.*Tx` directly and skipped them. Since #23 makes suggestion-apply the primary write path for contributors/agents, the main input path was the least enforced.

## Fix — one enforced Tx function per entity, called by both writers
- **incident** — `enforceIncidentWriteTx`: open-CA guard on resolve/close + lifecycle timestamps.
- **corrective action** — `enforceCorrectiveActionWriteTx`: open-task guard on resolve + `resolved_at`/`resolved_by_id`; `applyCorrectiveActionDefaults` shares create defaults across both paths.
- Handlers wrap them in `WithOrgTx`, map guard violations → 409; suggestion-apply calls the same functions inside the apply tx.
- DB: `CountOpenCAsByIncidentTx`, `SetIncidentLifecycleTx`, `CountOpenTasksByCATx`, `SetCorrectiveActionResolvedTx`.

## Test plan
- [ ] `just test-go` green
- [ ] `just test tests/test_incident_writepath.py tests/test_corrective_writepath.py` — guard + timestamp + create-default parity across HTTP and apply paths

Pattern for the rest: `docs/write-path-pattern.md`.

### DIFF
--- a/docs/write-path-pattern.md
+++ b/docs/write-path-pattern.md
@@ -1,0 +1,76 @@
+# One enforced write path per entity (#26)
+
+## The problem
+
+Business rules (RBAC, guards, lifecycle timestamps, notifications) historically
+lived only in the HTTP handler layer (`api_*.go`). But entities have **more than
+one writer**:
+
+- HTTP handlers (`handleUpdateX`)
+- suggestion-apply (`api_suggestions.go`, runs inside `WithOrgTx`, calls the
+  DB-layer `*Tx` functions directly)
+- CLI / MCP (via the HTTP API)
+
+When suggestion-apply called `db.UpdateXTx` directly it skipped every rule that
+lived in the handler. Since #23 makes suggestion-apply the primary write path for
+contributors and agents, the *main* input path was the *least* enforced one.
+That divergence is the root of a whole cluster of integrity bugs.
+
+## The pattern
+
+**One enforced, transaction-aware mutation function per entity. Every writer
+calls it inside a transaction.** Rules can't diverge because there's exactly one
+place that enforces them.
+
+```
+enforceXWriteTx(ctx, tx, orgID, entity, prevState) error
+  ├─ guards / RBAC-independent business rules   (e.g. open-CA guard)
+  ├─ the row write                              (db.UpdateXTx)
+  └─ derived state                              (e.g. lifecycle timestamps)
+```
+
+- Lives in the `api` package (it composes DB `*Tx` calls + business rules), named
+  `enforceXWriteTx`.
+- Takes a `pgx.Tx` — the caller owns the transaction (`WithOrgTx`), so the whole
+  mutation is atomic.
+- Returns typed errors for rule violations (e.g. `openCAsLinkedError`) so the HTTP
+  handler can map them to the right status (409) while suggestion-apply surfaces
+  them as the apply failure.
+
+## Migrated entities
+
+- **incident** — `enforceIncidentWriteTx` (open-CA guard + lifecycle timestamps).
+- **corrective_action** — `enforceCorrectiveActionWriteTx` (open-task guard +
+  `resolved_at`/`resolved_by_id`) and `applyCorrectiveActionDefaults` (shared
+  create defaults, so apply and HTTP seed the same starting state).
+
+Together these close the three integrity bypasses #26 names. Remaining entities
+follow the same recipe, one at a time.
+
+## Reference implementation — incident
+
+- `enforceIncidentWriteTx` — `internal/isms/api/api_incidents_writepath.go`
+- DB primitives — `internal/isms/db/suggestions_tx.go`:
+  `CountOpenCAsByIncidentTx`, `SetIncidentLifecycleTx`, `UpdateIncidentTx`
+- Callers:
+  - `handleUpdateIncident` (`api_incidents.go`) wraps it in `WithOrgTx`, maps
+    `openCAsLinkedError` → HTTP 409.
+  - `applyIncidentUpdate` (`api_suggestions.go`) calls it inside the existing
+    apply transaction.
+- Parity test — `tests/test_incident_writepath.py`: the open-CA guard and the
+  `resolved_at` timestamp now fire identically on the HTTP and apply paths.
+
+## Adding the next entity
+
+1. Write `enforceXWriteTx(ctx, tx, orgID, x, prev)` capturing the rules currently
+   inlined in `handleUpdateX` (guards, derived timestamps, status transitions).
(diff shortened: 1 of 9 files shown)


unidoc-alip (changes requested):
Solid first slice of #26: incident and corrective-action updates now flow through one enforced, transaction-aware function shared by the HTTP handler and suggestion-apply, closing the three named bypasses with parity tests to lock it in. Compiles clean and the guard-before-write ordering is correct on both entities. One should-fix below is a regression in an edge-case safeguard the refactor dropped; verified live against a throwaway DB, not just by reading the code.

---

### 1. [should-fix] `internal/isms/api/api_corrective_writepath.go` line 250

`enforceCorrectiveActionWriteTx` dropped the explicit empty-identifier guard that the code it replaced had:

```go
if existing.Identifier == "" {
    return echo.NewHTTPError(http.StatusInternalServerError, "corrective action has no identifier")
}
```

`CountOpenTasksByCATx` matches via `title LIKE '%'+identifier+'%'`. With an empty identifier that becomes `LIKE '%%'`, which matches **every** non-done/cancelled `ca_followup` task in the org — not zero. So a corrective action with a corrupt/empty identifier either gets incorrectly blocked from resolving (if any such tasks exist org-wide) or, if none exist, silently resolves with the guard effectively disabled — exactly the failure mode the original comment ("skipping would silently disable enforcement") was written to prevent.

**Verified live** against a throwaway Postgres DB (schema from `migrations/20260327000000_initial_schema.sql`) — not just by reading the code. Seeded a corrective_action with `identifier = ''` and one open `ca_followup` task with zero relation to it, then ran `CountOpenTasksByCATx`'s exact query with `caIdentifier = ''`:

```sql
SELECT count(*) FROM tasks
WHERE organization_id = <org id> AND task_type = 'ca_followup'
  AND status NOT IN ('done','cancelled') AND deleted_at IS NULL
  AND (title LIKE '%' || '' || '%' OR COALESCE(description,'') LIKE '%' || '' || '%');
-- => 1   (matches a task with zero relation to the CA)
```

A control run with a real, non-matching identifier (`'CA-42'`) against the same task correctly returns `0`. This confirms the empty-identifier case degrades to `LIKE '%%'` and matches everything, reproducing both failure modes described above.

Suggested fix — restore the check inside `enforceCorrectiveActionWriteTx`, gated the same way the original code gated it (only on the `resolving` transition, before calling `CountOpenTasksByCATx`):

```go
func enforceCorrectiveActionWriteTx(ctx context.Context, tx pgx.Tx, orgID int, ca *db.CorrectiveAction, prevStatus, actor string) error {
    resolving := ca.Status == "resolved" && prevStatus != "resolved"
    if resolving {
        // An empty identifier is corrupt data — the open-task LIKE query would
        // otherwise degrade to '%%' and match every open task in the org, either
        // false-blocking the resolve or silently disabling the guard (#26).
        if ca.Identifier == "" {
            return fmt.Errorf("corrective action %d has no identifier", ca.ID)
        }
        n, err := db.CountOpenTasksByCATx(ctx, tx, orgID, ca.Identifier)
        if err != nil {
            return err
        }
        if n > 0 {
            return openTasksLinkedError{identifier: ca.Identifier, n: n}
        }
    }
    if err := db.UpdateCorrectiveActionTx(ctx, tx, orgID, ca); err != nil {
        return err
    }
    if resolving {
        return db.SetCorrectiveActionResolvedTx(ctx, tx, orgID, ca.ID, actor)
    }
    return nil
}
```

No handler changes needed: this returns a plain `error`, so both callers' existing fallback path already does the right thing — `handleUpdateCorrectiveAction`'s `errors.As(err, &ote)` check will miss (it's not an `openTasksLinkedError`) and fall through to `pgxHTTPError(err)`, which maps an unrecognized error to 500 — matching the original handler's behavior for this exact case. Suggestion-apply surfaces it as the apply failure message, same as any other guard error.

---

### 2. [nit] `internal/isms/db/suggestions_tx.go` line 499

`SetIncidentLifecycleTx` now runs unconditionally on every incident write (even when status didn't change), adding a second `UPDATE incidents` statement per request. `enforceIncidentWriteTx` (`internal/isms/api/api_incidents_writepath.go:385-399`) calls it at the end of every invocation regardless of `inc.Status != prevStatus` — contrast with the open-CA guard three lines above it, which is correctly gated on that comparison. The `COALESCE(contained_at, now())`-style clauses make repeated calls idempotent, so this isn't a correctness bug, just avoidable DB round-trip work on every plain field edit (title, notes, etc.) that doesn't touch status. Worth a look since this pattern is explicitly meant to be replicated for other entities per `docs/write-path-pattern.md`.

Suggested fix — gate the lifecycle call the same way the guard is already gated:

```go
func enforceIncidentWriteTx(ctx context.Context, tx pgx.Tx, orgID int, inc *db.Incident, prevStatus string) error {
    statusChanged := inc.Status != prevStatus
    if statusChanged && (inc.Status == "resolved" || inc.Status == "closed") {
        n, err := db.CountOpenCAsByIncidentTx(ctx, tx, orgID, inc.Identifier)
        if err != nil {
            return err
        }
        if n > 0 {
            return openCAsLinkedError{verb: statusVerb(inc.Status), n: n}
        }
    }
    if err := db.UpdateIncidentTx(ctx, tx, orgID, inc); err != nil {
        return err
    }
    if !statusChanged {
        return nil
    }
    return db.SetIncidentLifecycleTx(ctx, tx, orgID, inc.ID, inc.Status)
}
```

unidoc-ahall:
Both addressed:

1. **CA empty-identifier guard (should-fix)** — restored inside `enforceCorrectiveActionWriteTx`, gated on the `resolving` transition before `CountOpenTasksByCATx`, so a corrupt/empty identifier errors out instead of degrading the `LIKE` to `'%%'` and matching every open task. Returns a plain `error`, so both callers' existing fallback (handler → `pgxHTTPError` → 500; apply → surfaced) behaves as before. Good catch — that was a real safeguard my refactor dropped.
2. **Incident lifecycle write (nit)** — now gated on `statusChanged`, so `SetIncidentLifecycleTx` only runs on a status transition, not on every plain field edit. The pattern doc's example follows this shape now too.

`just test-go` green.

unidoc-alip (approved):
looks good to me

PR #125, opened 2026-07-06
fix(cli): honor ISMS_ROOT / --root for clone, sync, git (#124)

### LINKED ISSUE #124: CLI ignores ISMS_ROOT / --root — clone, sync, git don't use the configured local-repo root
`ISMS_ROOT` (and the `--root` flag) are effectively dead: the only reader is `getRoot()` in cmd/isms/main.go, and `getRoot()` is never called. So:
- `isms clone` with no directory arg defaults to `{org-slug}-isms` and ignores `ISMS_ROOT`.
- `isms sync` / `isms git` locate the repo via `gitRoot()` (walk up from cwd), also ignoring `ISMS_ROOT`/`--root`.

Setting `ISMS_ROOT=/path/to/clone` does nothing — surprising, since it looks like the canonical 'where is my clone' setting.

**Fix (Option A — wire it up):** make one resolver, `resolveRepoRoot()` (precedence: `--root` → `ISMS_ROOT` → walk-up-to-.git, error otherwise), the single source of truth. `clone` uses the configured root as the target dir when no arg is given (printing which root it used; still fails if the dir exists); `sync`/`git` use it so the repo is found from anywhere, not just inside it.

NB: `isms tui` reads documents from the server API, not a local clone — a separate question of whether it should read `ISMS_ROOT` for offline/local-edit browsing (own issue).

unidoc-ahall:
Closes #124

`ISMS_ROOT`/`--root` were dead config (only `getRoot()` read them, and it was never called). Now `resolveRepoRoot()` (`--root` → `ISMS_ROOT` → walk-up-to-.git) is the one resolver:
- `clone` uses the configured root as the target when no dir arg (prints it; still fails if it exists).
- `sync`/`git` find the repo from anywhere via the same resolver.

## Test plan
- [ ] `ISMS_ROOT=/tmp/x isms clone` → clones into /tmp/x, prints 'Using configured root'
- [ ] `ISMS_ROOT=/tmp/x isms sync` from a different cwd → operates on /tmp/x
- [ ] no ISMS_ROOT, inside a clone → sync/git still work (walk-up)
- [ ] `just test-go` green

### DIFF
--- a/cmd/isms/clone.go
+++ b/cmd/isms/clone.go
@@ -20,7 +20,8 @@ func cloneCmd() *cobra.Command {
 		Short: "Clone the ISMS repository from the server",
 		Long: `Clone the organization's ISMS git repository from the server.
 
-Directory defaults to {org-slug}-isms. All config comes from your env file:
+Directory defaults to the configured root (--root or ISMS_ROOT), else
+{org-slug}-isms. All config comes from your env file:
   ISMS_BASE_URL, ISMS_API_TOKEN, ISMS_ORGANIZATION`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// Get base URL and token from env
@@ -50,13 +51,20 @@ Directory defaults to {org-slug}-isms. All config comes from your env file:
 				return fmt.Errorf("token is not scoped to an organization")
 			}
 
-			// Target directory
+			// Target directory. Precedence: explicit --dir/arg, then the
+			// configured root (--root or ISMS_ROOT) so clone and the other local
+			// commands agree on where the repo lives, then {slug}-isms.
 			target := dir
 			if target == "" && len(args) > 0 {
 				target = args[0]
 			}
 			if target == "" {
-				target = info.OrganizationSlug + "-isms"
+				if r := configuredRoot(); r != "" {
+					target = r
+					fmt.Printf("Using configured root %s (from --root/ISMS_ROOT)\n", target)
+				} else {
+					target = info.OrganizationSlug + "-isms"
+				}
 			}
 
 			if _, err := os.Stat(target); err == nil {
--- a/cmd/isms/git.go
+++ b/cmd/isms/git.go
@@ -18,16 +18,16 @@ func gitCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:                "git [-- git-args...]",
 		Short:              "Run a git command inside the ISMS repo (auth handled)",
-		Long:               "Thin passthrough to the system `git` CLI. Runs in the ISMS repo root and uses the credential helper that `isms sync` configured for the remote, so the user does not have to manage tokens.",
+		Long:               "Thin passthrough to the system `git` CLI. Runs in the ISMS repo root and uses the credential helper that `isms sync` configured for the remote, so the user does not have to manage tokens.\n\nNote: --root is NOT supported here (args are passed straight through to the system git CLI) — use ISMS_ROOT to point at a clone from another directory.",
 		DisableFlagParsing: true,
 		Args:               cobra.MinimumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if _, err := exec.LookPath("git"); err != nil {
 				return fmt.Errorf("git CLI not found on PATH — install it")
 			}
-			root, err := gitRoot()
+			root, err := resolveRepoRoot()
 			if err != nil {
-				return fmt.Errorf("not in an ISMS git repository — run `isms clone` first")
+				return fmt.Errorf("no ISMS repo found — set ISMS_ROOT, or cd into your clone (or run `isms clone`)")
 			}
 			// Refresh credential helper for the current API token. Idempotent —
 			// catches older repos that were cloned before the helper code worked.
--- a/cmd/isms/main.go
+++ b/cmd/isms/main.go
@@ -113,20 +113,25 @@ func isServerCommand(cmd *cobra.Command) bool {
 	return false
 }
 
-func getRoot() string {
+// configuredRoot returns the explicitly configured repo root — the --root flag,
+// else ISMS_ROOT — or "" if neither is set. The single place that encodes that
+// precedence, shared by resolveRepoRoot and clone so they can't drift.
+func configuredRoot() string {
 	if ismsRoot != "" {
 		return ismsRoot
 	}
-	if v := os.Getenv("ISMS_ROOT"); v != "" {
-		return v
+	return os.Getenv("ISMS_ROOT")
+}
+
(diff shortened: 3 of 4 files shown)


unidoc-alip (changes requested):
`resolveRepoRoot()` is a clean single resolver and `sync` now uses it correctly. I verified two real gaps by building the binary and running it, plus one doc nit. Neither should-fix blocks merge — `ISMS_ROOT` works as a functioning alternative for both — but both are confirmed, concrete issues worth fixing before or shortly after merge. No security issues found (env vars/flags are trusted local input and the git exec uses argv, not a shell).

---

### 1. [should-fix] `cmd/isms/git.go` line 19

`gitCmd()` sets `DisableFlagParsing: true`. Verified by building the binary: `--root` **never** works with `isms git`, in any position — not just "after the subcommand" as it might first appear. Cobra forwards the raw, unparsed arg list straight to the underlying `git` binary once it dispatches to a `DisableFlagParsing` command, so both `isms git --root /foo status` and `isms --root /foo git status` fail identically with `unknown option: --root` from the system `git` process. Worse, the error this PR adds at git.go:30 ("no ISMS repo found — set ISMS_ROOT, **pass --root**, or cd into your clone") tells the user to do something that cannot work for this command. `ISMS_ROOT` (env var, not a flag) works fine since it's read directly via `os.Getenv`, unaffected by cobra's flag parsing.

PoC (built from this branch):
```
$ go build -o /tmp/isms-test ./cmd/isms
$ cd /tmp/isms-poc/fakerepo && git init -q .

# --root AFTER the subcommand
$ /tmp/isms-test git --root /tmp/isms-poc/fakerepo status
unknown option: --root
Error: exit status 129

# --root BEFORE the subcommand — same failure
$ /tmp/isms-test --root /tmp/isms-poc/fakerepo git status
unknown option: --root
Error: exit status 129

# --root=value form — same failure
$ /tmp/isms-test --root=/tmp/isms-poc/fakerepo git status
unknown option: --root=/tmp/isms-poc/fakerepo
Error: exit status 129

# Baseline: isms git works fine with no --root at all, from inside a real repo
$ /tmp/isms-test git status
On branch master
nothing to commit ...
exit: 0

# Workaround: ISMS_ROOT env var (not a flag) works from ANY cwd
$ cd /tmp && ISMS_ROOT=/tmp/isms-poc/fakerepo /tmp/isms-test git status
On branch master
nothing to commit ...
exit: 0
```
Even `isms git --help` reproduces the issue — it prints system `git`'s help, not cobra's, because `--help` is also swallowed by the raw passthrough.

Minimum fix: stop telling users `--root` works with `isms git`. Change the error message at git.go:30 to only mention the working option:
```go
return fmt.Errorf("no ISMS repo found — set ISMS_ROOT, or cd into your clone (or run `isms clone`)")
```
and add a note to the command's `Long` help text: "Note: --root is not supported for `isms git` (it's a raw passthrough to the system git CLI) — use ISMS_ROOT instead."

Deeper fix (follow-up, not required for this PR): manually scan the raw `args` in `gitCmd()`'s `RunE` for a leading `--root`/`--root=value` before forwarding the rest to `exec.Command("git", ...)`, so the flag can be supported without re-enabling cobra's flag parsing (which would otherwise misinterpret native git flags like `-C`).

---

### 2. [should-fix] `cmd/isms/clone.go` line 61

This re-implements the `--root` → `ISMS_ROOT` precedence that `resolveRepoRoot()` (main.go) already encodes, just without the walk-up-to-`.git` fallback (which is correct to skip here — `clone` shouldn't target an *existing* enclosing repo). Confirmed by diffing the two implementations — both independently encode "`ismsRoot` var, else `ISMS_ROOT` env var, else fall through":

`main.go` (`resolveRepoRoot`, lines 121-129):
```go
if ismsRoot != "" {
    return ismsRoot, nil
}
if v := os.Getenv("ISMS_ROOT"); v != "" {
    return v, nil
}
return gitRoot()
```

`clone.go` (lines 60-71):
```go
if ismsRoot != "" {
    target = ismsRoot
} else if root := os.Getenv("ISMS_ROOT"); root != "" {
    target = root
}
if target != "" {
    fmt.Printf("Using configured root %s (from --root/ISMS_ROOT)\n", target)
} else {
    target = info.OrganizationSlug + "-isms"
}
```

Concrete drift scenario: if a future change adds a third source (e.g. a `~/.ismsrc` config file) to `resolveRepoRoot()`, `clone.go`'s inline copy would silently keep ignoring it — `isms clone` would default to `{slug}-isms` while `isms sync`/`isms git` would honor the new config file, reintroducing exactly the kind of "clone and sync disagree on where the repo lives" bug this PR set out to fix.

Suggested fix — extract the shared, non-walking-up part into its own helper and have both call sites use it:
```go
// configuredRoot returns the explicitly configured root (--root or
// ISMS_ROOT), or "" if neither is set.
func configuredRoot() string {
    if ismsRoot != "" {
        return ismsRoot
    }
    return os.Getenv("ISMS_ROOT")
}

func resolveRepoRoot() (string, error) {
    if r := configuredRoot(); r != "" {
        return r, nil
    }
    return gitRoot()
}
```
```go
// clone.go
if target == "" {
    if r := configuredRoot(); r != "" {
        target = r
        fmt.Printf("Using configured root %s (from --root/ISMS_ROOT)\n", target)
    } else {
        target = info.OrganizationSlug + "-isms"
    }
}
```

---

### 3. [nit] `cmd/isms/clone.go` line 23

The `Long` help text still says "Directory defaults to {org-slug}-isms" without mentioning that `--root`/`ISMS_ROOT` now take precedence over that default. Confirmed via `isms clone --help` — the text is unchanged from before this PR and doesn't reflect the new precedence, unlike the inline code comment added at clone.go:53-55.

```go
Long: `Clone the organization's ISMS git repository from the server.

Directory defaults to the configured root (--root or ISMS_ROOT), else
{org-slug}-isms. All config comes from your env file:
  ISMS_BASE_URL, ISMS_API_TOKEN, ISMS_ORGANIZATION`,
```

unidoc-ahall:
All three addressed:

1. **`--root` unusable with `isms git` (should-fix)** — right, `DisableFlagParsing: true` forwards args straight to system git, so `--root` can never work there. Dropped 'pass --root' from the error message and added a note in the command's Long help that `--root` isn't supported for `isms git` — use `ISMS_ROOT`. (The raw-arg-scan deeper fix can be a follow-up.)
2. **Duplicated precedence (should-fix)** — extracted `configuredRoot()` (--root → ISMS_ROOT, no walk-up); both `resolveRepoRoot()` and `clone` call it, so a future third source can't make them drift.
3. **clone help (nit)** — Long text now documents that `--root`/`ISMS_ROOT` take precedence over the `{slug}-isms` default.

`just test-go` + `go build ./cmd/isms` clean.

unidoc-alip (approved):
looks good to me

PR #130, opened 2026-07-06
feat(tui): read the local clone offline instead of the API (#126)

### LINKED ISSUE #126: TUI reads documents from the API — should read the local clone (offline, pre-sync)
isms tui currently reads documents over the API (client.ListAllDocuments / GetDocumentBody). But the TUI belongs to the clone→edit→sync workflow: it should read
  the **local clone** so it works offline and reflects your local, un-synced edits.

  **Change:** point the TUI at resolveRepoRoot() (--root / ISMS_ROOT / walk-up-to-.git) and read documents/**/*.md from the working tree instead of the API. Folder browser +
  render stay the same; only the data source changes.

  **No-clone behavior:** if no clone is found, fail with a clear "clone first" message (offline-first; API fallback can be added later if a real need surfaces).

  Depends on the ISMS_ROOT wiring (resolveRepoRoot).

unidoc-ahall:
Closes #126

The TUI now reads the **local clone** (offline), not the server API — it belongs to the clone→edit→sync workflow, so it reflects your un-synced local edits and needs no token/network.

- `tui.Run(root)` / `New(root)` open `store.New(root)` and list docs via `ListDocFolders` + `LoadDocumentsFromDir`, caching parsed bodies (no re-read on preview/open).
- The command resolves the clone with `resolveRepoRoot()` (--root / ISMS_ROOT / walk-up) and errors with 'run isms clone first' if none — offline-first, no API fallback.
- Drops the client dependency + dead `stripFrontmatter`.

## Test plan
- [ ] `just test-go` green (incl. new TUI load test)
- [ ] `isms tui` inside a clone → browse folders, read a doc rendered
- [ ] `ISMS_ROOT=/path/to/clone isms tui` from elsewhere → works
- [ ] `isms tui` with no clone → clear 'run isms clone first' error

### DIFF
--- a/cmd/isms/tui.go
+++ b/cmd/isms/tui.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 	"isms.sh/internal/isms/tui"
 )
@@ -8,10 +10,14 @@ import (
 func tuiCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "tui",
-		Short: "Interactive terminal UI",
+		Short: "Interactive terminal reader for the local clone (offline)",
+		Long:  "Browse and read the documents in your local ISMS clone, offline. Reads the clone at --root / ISMS_ROOT, or walks up from the current directory — run `isms clone` first if you don't have one.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			c := requireAPI()
-			return tui.Run(c)
+			root, err := resolveRepoRoot()
+			if err != nil {
+				return fmt.Errorf("no ISMS repo found — set ISMS_ROOT or cd into your clone (run `isms clone` first)")
+			}
+			return tui.Run(root)
 		},
 	}
 }
--- a/internal/isms/tui/tui.go
+++ b/internal/isms/tui/tui.go
@@ -12,7 +12,7 @@ import (
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/charmbracelet/glamour"
 	"github.com/charmbracelet/lipgloss"
-	"isms.sh/internal/isms/client"
+	"isms.sh/internal/isms/store"
 )
 
 // mode controls the top-level layout.
@@ -24,17 +24,21 @@ const (
 )
 
 // item is a row in the document list — either a folder header or a document.
+// Document rows cache the parsed body/version so the reader is fully offline —
+// no re-read when previewing or opening.
 type item struct {
 	docID    string // empty for folder rows
 	title    string
 	path     string // folder path for folder rows
 	status   string
+	version  string
+	body     string // markdown body (frontmatter stripped), empty for folder rows
 	isFolder bool
 }
 
 // Model is the bubbletea model for the document TUI.
 type Model struct {
-	client   *client.Client
+	store    *store.Store
 	width    int
 	height   int
 	ready    bool
@@ -56,14 +60,14 @@ const (
 	helpHeight    = 1
 )
 
-// New creates a new document-focused TUI model.
-func New(c *client.Client) Model {
+// New creates a document-focused TUI model that reads the local clone at root.
+func New(root string) Model {
 	r, _ := glamour.NewTermRenderer(
 		glamour.WithAutoStyle(),
 		glamour.WithWordWrap(80),
 	)
 	m := Model{
-		client:   c,
+		store:    store.New(root),
 		renderer: r,
 	}
 	m.loadDocuments()
@@ -182,7 +186,7 @@ func (m Model) handleKey(msg tea.KeyMsg) (tea.Model, tea.Cmd) {
 		if m.mode == modeList {
 			vis := m.visibleItems()
 			if m.cursor < len(vis) && !vis[m.cursor].isFolder {
-				m.openDocument(vis[m.cursor].docID)
+				m.openDocument(vis[m.cursor])
 			}
 		}
 		return m, nil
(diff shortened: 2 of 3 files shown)


unidoc-alip (changes requested):
Solid rework — the TUI correctly moves off the API client and onto `store.New(root)` / `ListDocFolders` / `LoadDocumentsFromDir`, caching bodies so preview/read are truly offline. Build and the new load test pass. Two things worth fixing before merge: folder headers now show the raw directory name instead of the `.title`-file display name the API path used to honor (the PR description says "folder browser + render stay the same"), and a single malformed doc in a folder silently zeroes out that whole folder in the list with no error surfaced. Both are verified below with a reproducing test and a working fix.

---

### 1. [must-fix] `internal/isms/tui/tui.go` line 425

Folder headers now always render the raw directory name (`strings.ToUpper(folder)`) instead of the `.title`-file display name the previous API-backed code used (`f.Title`, falling back to `f.Name` only when no title was set). Any org with a custom folder display name (e.g. dir `iso27001` with `.title` = "ISO 27001:2022") silently loses that label in the TUI — contradicting the PR's own stated goal that "Folder browser + render stay the same; only the data source changes." `store.ReadFile` is already exported and used elsewhere for reading `.title` (see `api/server.go`'s `readDirTitle`) — the TUI can read `{root}/documents/{folder}/.title` the same way.

**Verified.** A temp test creating `documents/iso27001/.title` = "ISO 27001:2022" plus one valid doc, then calling `New(root)`, shows the folder header renders `"ISO27001"` (raw dir name) instead of the expected `"ISO 27001:2022"`:

```
=== RUN   TestPoC_FolderTitleIgnoresDotTitleFile
    folder header title = "ISO27001" (expected something derived from .title = "ISO 27001:2022", got raw dir name)
--- PASS: TestPoC_FolderTitleIgnoresDotTitleFile (0.01s)
```

**Suggested fix** — read `.title` the same way the API path does, falling back to the folder name:

```go
func (m *Model) loadDocuments() {
	m.items = nil
	var loadErrs []string
	for _, folder := range m.store.ListDocFolders() {
		docs, err := m.store.LoadDocumentsFromDir(folder)
		if err != nil {
			loadErrs = append(loadErrs, fmt.Sprintf("%s: %v", folder, err))
			continue
		}
		if len(docs) == 0 {
			continue
		}
		m.items = append(m.items, item{title: strings.ToUpper(m.folderLabel(folder)), isFolder: true, path: folder})
		sort.SliceStable(docs, func(i, j int) bool {
			return docs[i].Frontmatter.DocumentID < docs[j].Frontmatter.DocumentID
		})
		for _, d := range docs {
			m.items = append(m.items, item{
				docID:   d.Frontmatter.DocumentID,
				title:   d.Frontmatter.Title,
				status:  d.Frontmatter.Status,
				version: d.Frontmatter.Version,
				body:    d.Body,
			})
		}
	}
	if len(m.items) == 0 && len(loadErrs) == 0 {
		m.loadErr = "No documents found in the local clone (looked under documents/)."
	} else if len(loadErrs) > 0 {
		m.loadErr = "Failed to load: " + strings.Join(loadErrs, "; ")
	}
}

// folderLabel returns the folder's display name — the content of a .title
// file under it, same convention the API-backed folder tree honored via
// readDirTitle — falling back to the raw folder name when no .title is set.
func (m *Model) folderLabel(folder string) string {
	data, err := m.store.ReadFile(filepath.Join(m.store.DocsRoot(), folder, ".title"))
	if err != nil || len(data) == 0 {
		return folder
	}
	return strings.TrimSpace(string(data))
}
```

(needs `"path/filepath"` added to the import block)

Re-running the PoC test with this fix applied confirms it resolves the issue — the header now renders `"ISO 27001:2022"`. Full suite (`go test ./internal/isms/...`) still passes.

---

### 2. [should-fix] `internal/isms/tui/tui.go` line 422

`if err != nil || len(docs) == 0 { continue }` silently drops an entire folder from the list when `LoadDocumentsFromDir` errors — e.g. one file with malformed frontmatter causes `loadMarkdownRecursive` to bail out and discard every doc already collected for that folder, including valid ones. The user just sees the folder missing, with no indication anything went wrong (the pre-PR code surfaced load errors via `m.loadErr`).

**Verified.** A temp test with `documents/iso27001/4-1.md` (valid) plus `documents/iso27001/4-2.md` (no frontmatter delimiters) shows the whole folder — including the valid doc — disappears, and `m.loadErr` reports only the generic "no documents found" message:

```
=== RUN   TestPoC_OneBadDocHidesWholeFolder
    items after load: []
    loadErr = "No documents found in the local clone (looked under documents/)."
--- PASS: TestPoC_OneBadDocHidesWholeFolder (0.00s)
```

**Suggested fix** is the same `loadDocuments` diff as finding 1 above — it accumulates per-folder errors into `loadErrs` and surfaces them via `m.loadErr` instead of swallowing them. Re-running the PoC test with the fix applied shows the real error surfacing:

```
loadErr = "Failed to load: iso27001: loading .../documents/iso27001/4-2.md: parsing frontmatter: no frontmatter found"
```

Note this fix still drops the valid sibling doc (`4-1.md`) along with the bad one — `store.loadMarkdownRecursive` aborts the whole recursive walk on the first parse error. Surfacing the real error is a clear improvement over the status quo; making one bad doc not blot out its valid siblings would need a further change in `store.loadMarkdownRecursive` to collect-and-continue. Worth a follow-up, not blocking this PR.

---

### 3. [nit] `internal/isms/tui/tui.go` line 57

The comment `// markdown body (frontmatter stripped), empty for folder rows` is slightly stale — `store.LoadDocument` already parses frontmatter out via `parseFrontmatter` before `tui.go` ever sees the body, so nothing is being "stripped" in this code path anymore (that was the whole point of this PR removing `stripFrontmatter`). Suggest:

```go
body     string // markdown body (frontmatter already parsed out by store), empty for folder rows
```

unidoc-alip (approved):
looks good to me

PR #133, opened 2026-07-06
feat: self-service email change (#128) + change-request access-request type

### LINKED ISSUE #128: Users can't change their email — need self-service (API endpoint + Settings UI + CLI)
There is no way to change a user email anywhere — no API endpoint, no UI, no CLI. Fix all three:

  - API endpoint for changing the authenticated user email (e.g. PUT /users/me/email).
  - Settings UI: change-email field for the signed-in user.
  - CLI: `isms server user set-email` for admin/ops.
  - Security: email is the login identity — require re-auth (password/OTP) and verify the new address (verification link before it takes effect), same rigor as password
  change.

  Note: user links resolve by id (assignee_id = SELECT id FROM users WHERE email=…) so they survive a rename; point-in-time text fields (reporter, created_by,
  changelog/activity actor) keep the old email as audit history — confirm acceptable in review.

### LINKED ISSUE #131: Change requests: support an access-request type
Change requests should be able to represent an **access request** as well as a normal change, so a customer's 'please grant X access' flows through the same review/approval trail instead of a bespoke feature.

Scope (kept deliberately minimal — graduates the access-request discussion):
- `change_requests.type` — `change` (default) or `access_request`.
- Same lifecycle and approval flow; no access-request-only structured fields — the target system, grantee and duration live in the existing **notes**.
- Surfaced in the API (create + read), the CLI (`--type`), and the Changes create form.

Ships in 0.7.0 alongside #128 (they share the single 0.7.0 migration).

### LINKED ISSUE #132: ISMS_RATE_LIMIT is documented and configured but never read by the code
`ISMS_RATE_LIMIT` is documented in the README and operator guide, set in the dev/test compose (both services), and `tests/conftest.py` relies on it to disable rate limiting for the suite — but **no Go code has ever read it** (`git log -S ISMS_RATE_LIMIT -- '*.go'` is empty back to the initial commit).

Impact: the auth limiter (`handleLogin` per-email, passkey login, per-IP middleware) runs regardless of the setting. A full pytest run accumulates >5 failed logins on an account, trips the per-email cap, and then every subsequent login returns 429 — the whole suite collapses and it looks like everything is broken. A documented switch that silently does nothing is the grave part.

Fix: wire `ISMS_RATE_LIMIT` — `0` disables auth rate limiting, `N>0` sets the per-IP cap, unset keeps the default (20). Guarded by unit tests so it can't silently rot again.

unidoc-ahall:
Closes #128
Closes #131
Closes #132

Two 0.7.0 features that share the **single** release migration `20260707000000_v0.7.0.sql`, plus a grave documented-but-unimplemented bug fix.

## 1. Self-service email change — verify-before-swap (#128)
- `PUT /auth/email` re-authenticates (current password + OTP if enabled) and stashes the new address in `users.pending_email`. **Your current email stays active.**
- A confirmation link is mailed to the **new** address; `POST /auth/verify-email-change` (unauthenticated — the token is the proof) swaps `email = pending_email` and marks it verified. A typo can never lock you out of your sign-in identity.
- The swap is atomic and re-checks uniqueness — a race just yields *already in use*. Existing sessions carry the old email and re-login.
- Settings gains a **Change email** section (shows the pending state); `/verify-email-change` confirms it.
- Admin escape hatch: `isms server user set-email --email … --new-email …` swaps directly.

## 2. Change requests carry a type
- `change_requests.type` = `change` (default) or `access_request` — same approval flow, access-request specifics live in the notes (no extra structured fields).
- API (create + read), CLI (`--type`, shown in `change show`), and the Changes create form.

## 3. Fix: ISMS_RATE_LIMIT was never wired (grave)
- Documented in README + operator guide, set in the dev/test compose, relied on by `conftest.py` — but **no Go code ever read it** (empty in `git log -S … -- '*.go'` since the initial commit). The limiter ran regardless, so a full test run tripped the per-email cap → 429 cascade → the suite looked broken.
- Now wired in `rate_limit.go`: `0` = off, `N>0` = per-IP cap, unset = default (20); applied to `handleLogin`, passkey login, and the per-IP middleware. Guarded by `rate_limit_test.go` (incl. a nil-DB canary proving the bypass short-circuits before touching the DB).
- `test_security.py::test_brute_force_protection` now `skipif ISMS_RATE_LIMIT==0` — it was passing *by accident* on the old bug (429 while the switch was ignored). It can't run against a server with the limiter off, and turning it on would trip the per-IP cap during the suite; the limiter is unit-tested instead.

## Migration + embed hygiene
- One file for 0.7.0 (`…_v0.7.0.sql`): `users.pending_email`, `change_requests.type`, and a widened `email_verifications.purpose` CHECK. The one-migration-per-minor rule is now documented in `docs/releasing.md`. All `IF NOT EXISTS` — safe against the persistent dev DB.
- **Embedded migrations are no longer in git.** `cmd/isms/migrations/*.sql` is gitignored; only a `.keep` is tracked so `//go:embed all:migrations` resolves on a fresh clone (verified: compiles with only `.keep`). Real migrations live in `migrations/` and are `cp`'d into the embed dir at compile time — `release.yml` already did; this adds the same `cp` to the `tests.yml` CI job and the `isms-test` compose command. The `initial_schema.sql` mistakenly committed into the embed dir at the initial commit is untracked.

## Test plan
- [ ] `just test-restart` (recompiles via go run → applies the embedded 0.7.0 migration + honors ISMS_RATE_LIMIT=0)
- [ ] `just test` green — incl. `tests/test_change_email.py` + `TestChangeType`
- [ ] Settings → Change email → link to new inbox → confirm → re-login with new email
- [ ] Create a change with type = Access request in the UI; `isms change create --type access_request …`
- [ ] `isms server user set-email` swaps an account directly


unidoc-alip (changes requested):
Solid, well-tested feature (self-service email change + change-request type + the `ISMS_RATE_LIMIT` fix), but the email-change request path can leave a user's account permanently stuck in "pending change" with no recovery if the mailer is unset or a send fails — `SetPendingEmail` and the verification token are committed before the mail is sent, and there's no cancel endpoint. A few smaller correctness/consistency gaps and some duplicated re-auth/token logic are worth a look too. All findings below were re-verified against the current code, with a proof-of-concept and a suggested fix for each.

---

### 1. [must-fix] `internal/isms/api/api_auth.go` line 540

`handleRequestEmailChange` commits `pending_email` (via `SetPendingEmail`, line 540) and creates the verification token (`CreateEmailChangeToken`, line 555) *before* checking whether the mailer is configured (line 559) or whether the send itself succeeds (line 565-567). If the mailer is `nil`/disabled, or `SendEmailChangeVerificationBranded` fails, the handler returns a 500 but the DB is left with `pending_email` set and a live, undelivered token. There's no endpoint wired to `ClearPendingEmail` (it exists in `db/users.go:231` but isn't exposed via the API — only `PUT /auth/email` and `POST /auth/verify-email-change` are registered), so on a self-hosted deployment without SMTP configured, every attempt to change email leaves the account permanently showing the "pending — check your inbox" banner with no way to clear it short of an admin touching the DB directly.

Reproduction: with no mailer configured, `PUT /auth/email {"new_email":"new@example.com","current_password":"<pw>"}` returns 500 "email delivery is not configured on this server", but `SELECT pending_email FROM users WHERE id = <id>` now shows `new@example.com` and a live `email_verifications` row exists. Every retry hits the same 500 and leaves the account in the same stuck state — there's no code path back to a clean state.

Suggested fix — check mailer availability before committing any state:
```go
if s.mailer == nil || !s.mailer.Enabled() {
    return echo.NewHTTPError(http.StatusInternalServerError,
        "email delivery is not configured on this server — contact the administrator")
}
// ... then SetPendingEmail + CreateEmailChangeToken + send
```
That covers the common self-hosted case. For the remaining case — `SetPendingEmail`/`CreateEmailChangeToken` succeed but the send itself fails (network blip, provider outage) — either send the mail first and only persist state after a successful send, or on send failure call `s.db.ClearPendingEmail(ctx, user.ID)` and `s.db.InvalidateEmailVerifications(ctx, user.ID, "email_change")` before returning the 500. Also wire `ClearPendingEmail` to a real endpoint (e.g. `DELETE /auth/email`) so a user/admin can cancel a pending change without DB access.

---

### 2. [should-fix] `internal/isms/db/users.go` line 191

`SetPendingEmail` has no DB-level uniqueness guard: the migration only adds `pending_email TEXT` with no unique index (`migrations/20260707000000_v0.7.0.sql:8`). The `isUniqueViolation` branch here can never fire from this statement — it's dead code copied from the pattern used by `SetEmail`/`SwapPendingEmail`, which *do* hit the real unique constraint on `email`. Two users can concurrently request the same new address; both get `"verification_sent"` and both receive a confirmation link to the same inbox. The pre-check at `api_auth.go:536` (`GetUserByEmail(ctx, newEmail)`) only queries the active `email` column, not `pending_email`, so it doesn't see the other user's in-flight request either.

Reproduction: user A requests a change to `shared@example.com` (succeeds, pending_email set, token mailed). Before A verifies, user B requests a change to the same `shared@example.com` — the pre-check passes (nobody's *active* email is that address yet), `SetPendingEmail` succeeds for B too, and both get 200 with a token mailed to the same inbox. Whoever verifies first wins via `SwapPendingEmail`'s real constraint on `email`; the second gets a confusing 409 at verification time instead of at request time, and is left with a stale `pending_email` with no way to clear it (see #1).

Suggested fix — add a partial unique index so the DB rejects the second claim at request time:
```sql
CREATE UNIQUE INDEX IF NOT EXISTS users_pending_email_unique
  ON users (pending_email) WHERE pending_email IS NOT NULL;
```
With that in place, `SetPendingEmail`'s existing `isUniqueViolation` branch stops being dead code and the race above surfaces a 409 immediately instead of at verification time.

---

### 3. [should-fix] `.github/workflows/tests.yml` line 12

This PR fixes `tests.yml` missing the `cp -f migrations/*.sql cmd/isms/migrations/` step, but the same line is now duplicated across five places (`Justfile:16`, `Justfile:116`, `release.yml:49`, `devenv/compose.yml:138`, `tests.yml` here) with no single source of truth. `MigrateFS` (`internal/isms/db/postgres.go:193-237`) reads `.sql` files from the embedded dir and simply does nothing — no error, no log — if that dir only has a `.keep` placeholder. That's exactly the failure mode this PR exists to fix, just reachable through any future entrypoint that forgets the same `cp` line.

Suggested fix — two independent changes:
1. Make the failure loud instead of silent: treat `len(files) == 0` in `MigrateFS` as an error (or at minimum log a warning), so a skipped `cp` step fails fast at startup instead of surfacing as a confusing "relation does not exist" error later.
2. Consolidate the `cp` line into one place every build path calls — e.g. a `just migrations-sync` target that `Justfile`, both workflow files, and `devenv/compose.yml` all invoke, or a `go:generate` directive next to the `//go:embed` in `cmd/isms`.

---

### 4. [should-fix] `cmd/isms/user.go` line 253

`isms server user set-email` only checks that `--new-email` is non-empty — it doesn't validate the address format the way the self-service API path does (`netmail.ParseAddress` + exact-match check in `handleRequestEmailChange`, `api_auth.go:501-503`). An admin typo or malformed value (e.g. a display-name-embedded address) is written straight to `email` with only lowercasing applied via `SetEmail`.

Reproduction: `isms server user set-email --email admin@example.com --new-email "Display Name <notreal@@bad"` writes the malformed string straight into `users.email` with no error. Subsequent logins or mail sends to that account will fail unpredictably since `email` is no longer a valid address.

Suggested fix — reuse the same validation as the API handler:
```go
lower := strings.ToLower(strings.TrimSpace(newEmail))
if addr, err := netmail.ParseAddress(lower); err != nil || addr.Address != lower {
    return fmt.Errorf("--new-email is not a valid email address")
}
```

---

### 5. [nit] `internal/isms/api/api_auth.go` line 599

`UseEmailVerification` (marking the token consumed) only runs after `SwapPendingEmail` succeeds (line 608). If the swap fails with `ErrEmailTaken` or "no pending change" (lines 601-606), the function returns early and the token is never marked used — it stays valid and replayable until its natural 2h expiry instead of being invalidated on first (failed) use.

Suggested fix — mark the token used unconditionally before attempting the swap:
```go
s.db.UseEmailVerification(ctx, verification.ID)
newEmail, err := s.db.SwapPendingEmail(ctx, verification.UserID)
if err != nil { ... }
```

---

### 6. [nit] `internal/isms/api/api_auth.go` line 521

The password + OTP re-authentication block here duplicates the same `bcrypt.CompareHashAndPassword`/`verifyTOTP` pattern already used elsewhere in this file (3 call sites total), and the token-generation snippet a few lines below (`rand.Read` + `sha256.Sum256`, line 552-554) is the 6th copy of that exact sequence in `api_auth.go` (signup verify, password reset, resend-verification, etc.). Worth extracting shared helpers so future changes to re-auth or token strength don't require hunting down every call site.

Suggested fix:
```go
func reauthenticate(user *model.User, password, otp string) error {
    if user.HasPassword() {
        if password == "" { return errNoPassword }
        if err := bcrypt.CompareHashAndPassword([]byte(*user.PasswordHash), []byte(password)); err != nil {
            return errBadPassword
        }
    }
    if user.HasOTP() {
        if otp == "" { return errNoOTP }
        if !verifyTOTP(*user.OTPSecret, otp) { return errBadOTP }
    }
    return nil
}

func newHashedToken() (raw, hash string) {
    b := make([]byte, 32)
    rand.Read(b)
    raw = hex.EncodeToString(b)
    h := sha256.Sum256([]byte(raw))
    return raw, hex.EncodeToString(h[:])
}
```
Replace the 3 re-auth call sites and 6 token-generation call sites with calls to these.

---

### 7. [nit] `internal/isms/api/api_auth.go` line 536

The `GetUserByEmail(ctx, newEmail)` pre-check does a full DB round-trip — including decrypting the fetched user's OTP secret, which is immediately discarded — purely to return a nicer 409. The comment itself says it's "best-effort," and combined with #2, this check doesn't actually prevent the race it's checking for.

Suggested fix — once the partial unique index from #2 is in place, drop this pre-check entirely and rely on `SetPendingEmail`'s `ErrEmailTaken` branch (line 541-543) for the 409, saving a round-trip and an unnecessary OTP decrypt.

unidoc-ahall:
Thanks @unidoc-alip — all seven addressed in the latest push.

**1. [must-fix] stuck 'pending' state.** The mailer-availability check now runs *before* any state is written; `SetPendingEmail`/`CreateEmailChangeToken` happen only after it. A send failure rolls back (`ClearPendingEmail` + `InvalidateEmailVerifications`). Added `DELETE /auth/email` (+ a Cancel button in Settings) so a user/admin can clear a stuck pending change without DB access.

**2. [should-fix] no uniqueness on pending_email.** Added a partial unique index `users_pending_email_unique ON users (pending_email) WHERE pending_email IS NOT NULL`. Two concurrent requests to the same address now 409 at request time via `SetPendingEmail`'s `ErrEmailTaken` branch (no longer dead code).

**3. [should-fix] duplicated cp + silent MigrateFS.** One source of truth now: `scripts/sync-migrations.sh` (+ `just migrations-sync`); the Justfile, both workflows and the isms-test compose command all call it. `MigrateFS` now returns a clear error when the embed dir has no `.sql` (a skipped sync) instead of applying nothing silently.

**4. [should-fix] set-email validation.** `isms server user set-email` now validates with `net/mail` (trim + lowercase + exact-match), same as the API handler.

**5. [nit] replayable token.** `verify-email-change` consumes the token *before* the swap, so a failed swap can't leave it valid until expiry.

**6. [nit] duplication.** Extracted `reauthenticate()` and `newHashedToken()`; adopted the token helper at the signup / forgot-password / invite / resend-invite sites.

**7. [nit] wasteful pre-check.** Replaced the `GetUserByEmail` pre-check with a cheap `EmailExists()` (no row hydration, no OTP decrypt). Kept it (rather than dropping it) so an active-email collision still 409s up front — the pending_email index only covers pending-vs-pending.

Note: the unique index lands in the already-existing 0.7.0 migration, so verifying it locally needs a `just test-reset` (fresh DB), not just a restart. `just test-go` + the change-email/CA suites are green.

unidoc-alip (approved):
looks good to me

PR #135, opened 2026-07-06
test(ca): fix persistent-stack pagination flake in CA write-path test

### LINKED ISSUE #134: Flaky test: CA write-path test scans only the first page of a paginated list
`tests/test_corrective_writepath.py::test_apply_create_uses_handler_default_status` looks up the CA it just created by scanning `GET /corrective-actions` (first page only). The test stack is persistent, so once enough corrective actions accumulate (169+) the new one falls outside the page and the lookup returns None — a false failure. `GET /corrective-actions/:id` only accepts a numeric id, and the apply response returns the identifier (CA-N), so the test can't fetch it directly.

Fix: search by the CA's unique title (`?q=`), the persistent-stack-safe pattern the other suites use. See also the standing note that integration tests must be idempotent against the shared stack.

unidoc-ahall:
Closes #134

`test_apply_create_uses_handler_default_status` found the CA it created by scanning the **first page** of `GET /corrective-actions`. On the persistent test stack (169+ CAs) the new CA falls outside that page → lookup returns None → false failure. `GET /corrective-actions/:id` only accepts a numeric id and apply returns the identifier, so this searches by the CA's **unique title** (`?q=`) instead — the same persistent-stack-safe pattern used elsewhere.

## Test plan
- [ ] `just test tests/test_corrective_writepath.py` green, run twice (idempotent)

### DIFF
--- a/tests/test_corrective_writepath.py
+++ b/tests/test_corrective_writepath.py
@@ -10,8 +10,12 @@ import uuid
 import requests
 
 
-def _ca_by_identifier(api_url, headers, identifier):
-    r = requests.get(f"{api_url}/corrective-actions", headers=headers)
+def _ca_by_identifier(api_url, headers, identifier, q=""):
+    # Search by the (unique) title rather than scanning the first page — the test
+    # stack is persistent, so a plain list is paginated and a freshly-created CA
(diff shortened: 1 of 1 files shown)


unidoc-alip (approved):
looks good to me

PR #136, opened 2026-07-06
feat(tui): collapsible folder tree with titles + stable scroll (#53)

### LINKED ISSUE #53: TUI: working read-only document reader (folder browser, read, render)
## Problem

The TUI doesn't work as it should — navigation is broken and the two-pane layout (tree left, content right) doesn't behave. It needs to be taken through and made into a genuinely usable **read-only document reader** to start with.

## Scope (v1 — read-only documents)

- **Folder/document browser** in the left pane (tree of documents).
- **Read files**: selecting a document fetches and shows its content in the right pane (lazy load on select).
- **Render markdown nicely** for review (clean, readable).
- **Working keyboard navigation** (move, open, back, scroll).
- Read-only — no register editing. (Registers in the left nav are a later scope-add.)

## Known issue (from analysis)

- `renderPreview()` does a blocking HTTP call inside `View()` on every keypress (`internal/isms/tui/tui.go:267,376`) → freezes on a non-localhost API. Lift the fetch into a `tea.Cmd`, cache the last body, debounce on cursor movement.

## Acceptance criteria

- Browse the document tree, open a document, content renders cleanly in the right pane.
- Navigation is smooth against a remote API (no per-keypress freeze).

## Notes

The TUI stays a focused reader (decided in the CLI/TUI scope discussion). This makes that reader actually usable. From ahall's hands-on testing.

unidoc-ahall:
Closes #53

The local-clone reader (#126/#130) rendered docs but the **browser** was unusable — everything **flattened under the template folder** (subfolders like clauses/annex-a disappeared), always open, the sidebar jumped while navigating, and rows showed the raw id/filename instead of the title. This rebuilds the left pane into a real file tree:

- **Mirrors the real `documents/<template>/<subfolder>/` layout** — recursive, with `.title` display names at every level, the same structure the web viewer shows.
- **The tree always stays visible.** The selected document renders live in the right pane; `enter`/`→` moves focus into the document to scroll it, `esc`/`←`/`h` returns to the tree. (The old full-screen reader that hid the sidebar is gone — the reported bug.)
- **Collapsible folders** (`▸`/`▾`), **closed by default**; `enter`/`→` expand, `←`/`h` collapse. A lone top folder opens so the reader isn't a single dead row.
- **Titles, not ids**: document rows show the frontmatter title (fallback to id).
- **Stable scroll**: a `listOffset` that follows the cursor only when it would leave the pane (no more jumping).
- Filter surfaces matching docs with their ancestor folders, across nesting.
- **Vim page keys** in the reader: `ctrl+f`/`ctrl+b` (page), `ctrl+d`/`ctrl+u` (half-page).
- **`e` to edit**: opens the selected doc in `$EDITOR` (a working-tree edit in the local clone, consistent with clone→edit→sync); re-read on return. No-op with a hint if `$EDITOR` is unset.

## Test plan
- [ ] `just test-go` green (nested tree, collapse/expand, title display, .title labels, filter, doc path, no-editor no-op, bad-doc error)
- [ ] `isms tui` in a clone → subfolders nest under the template; folders start closed; enter/→ opens, ←/h closes; no sidebar jump on j/k
- [ ] selecting a document keeps the tree visible; enter scrolls it, esc back to tree
- [ ] `ctrl+f`/`ctrl+b` page through a long doc; `ctrl+d`/`ctrl+u` half-page
- [ ] `e` opens the doc in $EDITOR; on save+quit the reader shows the edit
- [ ] rows show titles; folder with a .title shows the pretty name
- [ ] `/` filter finds a doc; enter reads it; esc back

### DIFF
--- a/internal/isms/tui/tui.go
+++ b/internal/isms/tui/tui.go
@@ -1,10 +1,15 @@
 // Package tui provides an interactive terminal UI for the ISMS, focused on
-// reading documents. List view on the left, rendered document on the right.
-// Press enter to open a document in full-screen reader mode; q/esc to go back.
+// reading documents. A collapsible folder tree on the left mirroring the real
+// documents/ directory (template → subfolders → docs, closed by default), and the
+// selected document rendered in a scrollable pane on the right. The tree stays
+// visible while reading: enter/→ moves focus into the document to scroll it,
+// esc/←/h returns focus to the tree.
 package tui
 
 import (
 	"fmt"
+	"os"
+	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
@@ -16,49 +21,71 @@ import (
 	"isms.sh/internal/isms/store"
 )
 
-// mode controls the top-level layout.
-type mode int
-
-const (
-	modeList   mode = iota // two-pane: list on left, preview on right
-	modeReader             // full-screen rendered document
-)
-
-// item is a row in the document list — either a folder header or a document.
-// Document rows cache the parsed body/version so the reader is fully offline —
-// no re-read when previewing or opening.
-type item struct {
-	docID    string // empty for folder rows
-	title    string
-	path     string // folder path for folder rows
+// treeItem is one row of the flattened document tree (pre-order DFS). Folders
+// carry an expand flag and depth; a folder's descendants are the following items
+// with a greater depth. Documents carry cached metadata + body (offline read).
+type treeItem struct {
+	isFolder bool
+	label    string // folder .title/name, or document title
+	depth    int
+	expanded bool // folders only
+	docID    string
 	status   string
 	version  string
-	body     string // markdown body (frontmatter already parsed out by store), empty for folder rows
+	body     string // markdown body (frontmatter already parsed out by store)
+	path     string // absolute file path (for opening in $EDITOR)
+}
+
+// display is the row's shown name — the label, falling back to the document id
+// when a doc has no title (matches how the web viewer labels documents).
+func (it treeItem) display() string {
+	if !it.isFolder && strings.TrimSpace(it.label) == "" {
+		return it.docID
+	}
+	return it.label
+}
+
+// tnode is the nested tree built while reading the clone; flattened into
+// []treeItem for display.
+type tnode struct {
 	isFolder bool
+	label    string
+	docID    string
+	status   string
+	version  string
+	body     string
+	path     string
+	children []tnode
+}
+
+// editDoneMsg is delivered after $EDITOR exits (idx = the item edited).
+type editDoneMsg struct {
+	idx int
+	err error
 }
 
 // Model is the bubbletea model for the document TUI.
 type Model struct {
-	store    *store.Store
-	width    int
-	height   int
-	ready    bool
-	mode     mode
-	items    []item
-	cursor   int
-	filter   string
-	filterOn bool
-	viewport viewport.Model
-	docBody  string // raw markdown of the currently-open document
-	docTitle string
-	rendered string
-	renderer *glamour.TermRenderer
-	loadErr  string
+	store      *store.Store
+	width      int
+	height     int
+	ready      bool
+	reading    bool // focus is on the document pane (scrolling) vs the tree
+	items      []treeItem
+	cursor     int    // index into visibleRows()
+	listOffset int    // first visible row (stable scroll)
+	loadedDoc  string // docID currently rendered in the viewport
+	filter     string
+	filterOn   bool
+	viewport   viewport.Model
+	renderer   *glamour.TermRenderer
+	loadErr    string
 }
 
 const (
 	listPaneWidth = 38
 	helpHeight    = 1
+	readerHeader  = 2 // title + meta lines above the scrollable body
 )
 
 // New creates a document-focused TUI model that reads the local clone at root.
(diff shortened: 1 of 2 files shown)


unidoc-alip (changes requested):
Solid rewrite that fixes the reported #53 bugs (flattened tree, always-open sidebar, jumpy scroll, id-not-title rows) — the recursive tree, stable-scroll `listOffset`, and reading/tree focus split are all sound and well covered by tests. All four behavioral findings below were reproduced with an executable PoC (temporary test added, run, then deleted — no artifacts committed); the three cleanup nits are structural observations confirmed by direct inspection.

---

### 1. [should-fix] `internal/isms/tui/tui.go` line 335

Clearing an active filter with `esc` (tree focus, not filter-typing) clears `m.filter` and clamps the cursor into the full unfiltered row list, but doesn't reset it to 0 or try to keep it on the same item. Since row indices differ between the filtered and unfiltered views, the cursor after `esc` can land on an unrelated folder/document, and `syncPreview()` loads that unrelated item into the reader pane with no indication anything moved.

**Confirmed by execution:** filtering to a folder+doc pair and sitting on the "Leadership" doc (`docID: iso27001-5-1`), then pressing `esc`, left the cursor pointing at the unrelated top-level `policies` folder:
```
cursor after esc = 1, points at {isFolder:true label:policies depth:0 ... docID: ...}
```

**Suggested fix** — reset cursor/offset here too, matching the `filterOn`-mode esc/backspace/typing handlers just above (lines 262-286), which already do this:
```go
case "esc":
    if m.filter != "" {
        m.filter = ""
        m.cursor, m.listOffset = 0, 0   // was: just clampCursor(len(m.visibleRows()))
        m.syncPreview()
    }
    return m, nil
```

---

### 2. [should-fix] `internal/isms/tui/tui.go` line 279

`len(key) == 1` measures UTF-8 bytes, not runes. Any non-ASCII character typed while filtering (e.g. `é`, `ü`, non-Latin scripts) is 2+ bytes and silently fails this check — the keystroke is dropped with no feedback, so users can't filter by titles containing accented/non-ASCII characters.

**Confirmed by execution:** typing `É` while filtering left `m.filter` empty instead of `"É"`.

**Suggested fix** — compare rune count, not byte length:
```go
default:
    if r := []rune(key); len(r) == 1 {
        m.filter += string(r[0])
        m.cursor, m.listOffset = 0, 0
        m.syncPreview()
        return m, nil
    }
    return m, nil
```
Once multi-byte runes can enter the filter, backspace at line 273 also needs to remove one rune, not one byte:
```go
case "backspace":
    if r := []rune(m.filter); len(r) > 0 {
        m.filter = string(r[:len(r)-1])
        m.cursor, m.listOffset = 0, 0
        m.syncPreview()
    }
    return m, nil
```

---

### 3. [should-fix] `internal/isms/tui/tui.go` line 677

`buildFolder` drops any folder with zero children (`return n, len(n.children) > 0`), regardless of whether it has a `.title`. The web API's equivalent tree-builder (`internal/isms/api/server.go:1057` and `:1090`) keeps a folder if it has a title even when empty: `len(sub.Files) > 0 || len(sub.SubFolders) > 0 || sub.Title != ""`. So an empty template folder that only carries a `.title` placeholder (e.g. before docs are scaffolded into it) shows up in the web viewer's tree but silently vanishes from the TUI tree — a real cross-surface divergence, caused by the TUI reimplementing the same "walk documents/, skip dotfiles, resolve .title, recurse" logic independently instead of sharing it with `server.go`.

**Confirmed by execution:** an empty `annex-a/` folder with only a `.title` file ("Annex A") did not appear in the TUI tree at all, while `server.go`'s handler would keep it via its `sub.Title != ""` clause.

**Suggested fix** — match the API's inclusion rule in `buildFolder`:
```go
return n, len(n.children) > 0 || n.label != filepath.Base(dir)
```
(`n.label != filepath.Base(dir)` signals "has an explicit `.title`", mirroring `sub.Title != ""` — `folderLabel` falls back to the raw directory name when there's no `.title` file.)

Longer-term, consider extracting a shared tree-builder in `internal/isms/store` that both the API and the TUI call, so the inclusion rule, `.title` resolution, and sort order can't drift apart again — reasonable as a follow-up rather than blocking this PR.

---

### 4. [nit] `internal/isms/tui/tui.go` line 529

`countDocs` counts all descendant documents in the full tree, ignoring the active filter. While filtered to a term matching 1 of 5 docs in a folder, if the cursor lands on that ancestor folder, the placeholder still reports "5 document(s)" instead of the 1 actually reachable in the filtered view.

**Confirmed by execution:** filtering a 3-doc `clauses` folder down to 1 match left `countDocs` still reporting 3.

**Suggested fix** — count through the filter-aware `visibleRows()` instead of raw depth-walking:
```go
func (m Model) countDocs(fi int) int {
    if fi < 0 {
        return 0
    }
    d := m.items[fi].depth
    rows := m.visibleRows()
    inSubtree, n := false, 0
    for _, r := range rows {
        if r == fi {
            inSubtree = true
            continue
        }
        if inSubtree {
            if m.items[r].depth <= d {
                break
            }
            if !m.items[r].isFolder {
                n++
            }
        }
    }
    return n
}
```
Simpler alternative given this is a nit: just label the placeholder conditionally (e.g. "1 of 3 document(s) match") when `m.filter != ""`, without touching `countDocs`.

---

### 5. [nit] `internal/isms/tui/tui.go` line 390

The up/down/g/G handlers (and the three filter-typing branches at lines 262-286) each repeat the same three-line pattern: mutate cursor, `m.clampCursor(len(m.visibleRows()))`, `m.syncPreview()` — 7 occurrences total.

**Suggested fix** — factor into a helper:
```go
func (m *Model) setCursor(n int) {
    m.cursor = n
    m.clampCursor(len(m.visibleRows()))
    m.syncPreview()
}
```
Then `up`/`down`/`g`/`G` become one-liners (`m.setCursor(m.cursor - 1)`, etc.), and finding #1's fix collapses to `m.setCursor(0)`.

---

### 6. [nit] `internal/isms/tui/tui.go` line 30

`tnode` (nested, built during load) and `treeItem` (flattened, used for display) duplicate 7 of 9 fields. `buildFolder` builds a `[]tnode` tree, then a separate `flatten()` closure walks it into `[]treeItem`.

**Suggested fix** — have `buildFolder` append directly into `m.items` with a depth parameter, dropping the intermediate `tnode` type and `flatten` closure entirely. This is a larger refactor than the other findings — reasonable to defer to a follow-up since the current code works correctly.

---

### 7. [nit] `internal/isms/tui/tui.go` line 419

`visIndexOf` is a plain linear search equivalent to the stdlib `slices.Index` (available given this module's Go version). Its only call site always finds the target first, so the not-found/return-0 fallback is currently dead code — worth a comment if kept, since a future refactor could hit that path and silently teleport the cursor to the top row with no signal.

**Suggested fix**:
```go
import "slices"
...
m.cursor = slices.Index(m.visibleRows(), j)
```

unidoc-ahall:
Thanks @unidoc-alip — all addressed in the latest push (findings 1–5 and 7; #6 deferred as you suggested).

**1. esc-clear-filter cursor** — clearing an active filter now `setCursor(0)` instead of clamping into the full list, so the cursor no longer lands on an unrelated item.
**2. rune vs byte** — filter typing and backspace both operate on runes now, so accented/non-ASCII titles are searchable.
**3. .title-only folders** — `buildFolder` now returns `len(children) > 0 || label != filepath.Base(dir)`, mirroring the web API's `sub.Title != ""` rule; a placeholder folder no longer diverges between the two surfaces. (Agree a shared `store` tree-builder is the real long-term fix — noted as a follow-up.)
**4. countDocs** — filter-aware now: reachable matches while filtering, full subtree otherwise (so a collapsed folder still shows its total to entice expand).
**5. setCursor()** — extracted; nav + filter handlers and the finding-1 fix all use it.
**7. slices.Index** — replaces `visIndexOf`; the dead not-found fallback is gone.

**6 (tnode/treeItem merge)** — deferred to a follow-up as you noted; the current split works and I kept this PR's diff contained.

`just test-go` green, incl. new regression tests for 1–4.

unidoc-alip (approved):
looks good to me

PR #138, opened 2026-07-07
feat(change): editable, DB-constrained change-request type

### LINKED ISSUE #137: Change requests: make type editable + DB-constrained (from access-request discussion)
Graduated from the access-request discussion (#129).

`change_requests.type` ('change' default | 'access_request') shipped create-only in #131. Make it **editable** on existing records (misclassification is fixable) and constrain it properly at the DB level.

Decision:
- Keep `type TEXT` with a DB **CHECK** ('change', 'access_request') — defense-in-depth matching every other status/category column; extend later by adding a value to the CHECK + `db.ChangeTypes`. No lookup table, no access-request-only structured fields (specifics live in notes).
- Editable via the API (`PUT /changes/:id`), suggestion-apply, the Changes detail/edit UI (+ shown in the read view), and the CLI (`change update --type`). `validateEnum` guards all write paths as the first line; the CHECK is the DB backstop.

Closes the gap where the type could be set at create time but not viewed/changed afterwards.

unidoc-ahall:
Closes #137
From the access-request discussion #129 (decision recorded there).

The change-request `type` shipped create-only in #131 — you could set it when creating but not view or change it afterwards. This makes it **editable** and constrains it at the DB level.

- **Migration:** a `CHECK` on `change_requests.type` ('change','access_request') — defense-in-depth matching every other status/category column. Extend by adding a value here **and** in `db.ChangeTypes` (stays `TEXT` + Go enum; no lookup table, no access-request-only fields — specifics live in notes).
- **Editable everywhere:** `PUT /changes/:id` (validateEnum first-line guard), suggestion-apply, the Changes detail/edit UI (select + shown in the read view), CLI `change update --type`.

## Test plan
- [ ] `just test-reset` (fresh DB → migration incl. the new CHECK) then `just test tests/test_changes.py` — incl. edit-type + reject-invalid-type
- [ ] UI: open a change → Overview → Edit → change Type → save; type shows in the read view
- [ ] `isms change update <ref> --type access_request`
- [ ] DB rejects an invalid type (CHECK) even if a caller bypasses validateEnum

### DIFF
--- a/cmd/isms/change.go
+++ b/cmd/isms/change.go
@@ -151,7 +151,7 @@ func changeCreateCmd() *cobra.Command {
 }
 
 func changeUpdateCmd() *cobra.Command {
-	var title, desc, justification, priority, category, risk, rollback, notes, assignedTo string
+	var changeType, title, desc, justification, priority, category, risk, rollback, notes, assignedTo string
 	cmd := &cobra.Command{
 		Use:   "update <id|ref>",
 		Short: "Update change request fields (only the flags you pass are changed)",
@@ -166,7 +166,7 @@ func changeUpdateCmd() *cobra.Command {
 			fields := map[string]interface{}{}
 			f := cmd.Flags()
 			for flag, val := range map[string]*string{
-				"title": &title, "desc": &desc, "justification": &justification,
+				"type": &changeType, "title": &title, "desc": &desc, "justification": &justification,
 				"priority": &priority, "category": &category, "risk": &risk,
 				"rollback": &rollback, "notes": &notes, "assigned-to": &assignedTo,
 			} {
@@ -185,6 +185,7 @@ func changeUpdateCmd() *cobra.Command {
 			return nil
 		},
 	}
+	cmd.Flags().StringVar(&changeType, "type", "", "Type ("+strings.Join(db.ChangeTypes, ", ")+")")
 	cmd.Flags().StringVar(&title, "title", "", "Change title")
 	cmd.Flags().StringVar(&desc, "desc", "", "Description")
 	cmd.Flags().StringVar(&justification, "justification", "", "Business justification")
@@ -199,6 +200,7 @@ func changeUpdateCmd() *cobra.Command {
 
 // changeFieldJSON maps update flag names to the API's JSON field names.
 var changeFieldJSON = map[string]string{
+	"type":  "type",
 	"title": "title", "desc": "description", "justification": "justification",
 	"priority": "priority", "category": "category", "risk": "risk_level",
 	"rollback": "rollback_plan", "notes": "notes", "assigned-to": "assigned_to",
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -2378,6 +2378,7 @@ type changeCreateRequest struct {
 // metadata (approved_at, approved_by, implemented_at) is cleared correctly on
 // reverse transitions — never inline-set via UpdateChangeRequest.
 type changeUpdateRequest struct {
+	Type          *string    `json:"type"`
 	Title         *string    `json:"title"`
 	Description   *string    `json:"description"`
 	Justification *string    `json:"justification"`
@@ -2559,6 +2560,11 @@ func (s *Server) handleUpdateChange(c echo.Context) error {
 	if err := c.Bind(&req); err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
 	}
+	if req.Type != nil {
+		if err := validateEnum("type", *req.Type, db.ChangeTypes); err != nil {
+			return err
+		}
+	}
 	if req.Priority != nil {
 		if err := validateEnum("priority", *req.Priority, db.ChangePriorities); err != nil {
 			return err
@@ -2597,6 +2603,9 @@ func (s *Server) handleUpdateChange(c echo.Context) error {
 		}
 	}
 	cr := *old
+	if req.Type != nil {
+		cr.Type = *req.Type
+	}
 	if req.Title != nil {
 		cr.Title = *req.Title
 	}
--- a/internal/isms/api/api_suggestions.go
+++ b/internal/isms/api/api_suggestions.go
@@ -1074,6 +1074,14 @@ func applyChangeUpdate(ctx context.Context, tx pgx.Tx, s *Server, orgID int, sg 
 	if err != nil {
 		return "", fmt.Errorf("change request %s not found: %w", sg.EntityID, err)
 	}
+	if v, ok := payload.Fields["type"]; ok {
+		if sv, ok := v.(string); ok {
+			if err := validateEnum("type", sv, db.ChangeTypes); err != nil {
+				return "", err
+			}
+			cr.Type = sv
+		}
+	}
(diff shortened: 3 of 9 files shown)


unidoc-alip (changes requested):
Clean implementation of #137 — the load-then-overlay update pattern keeps partial updates safe, `validateEnum` guards both write paths, the changelog picks up type changes via `ToChangeMap`, and CHECK violations already map to HTTP 400. Security review found no issues: authorization is admin/manager-gated on every write path, all SQL is parameterized, and `type` gates no authorization logic anywhere. Two things are worth addressing before merge, both verified with live proofs of concept against a real Postgres (throwaway DB, migrate-twice run, and a Go reproduction — details inline below): the CHECK never reaches databases that already ran the v0.7.0 migration, and the empty-value defaulting in the update functions silently reclassifies records in exactly the scenario the CHECK is meant to catch.

---

### 1. [should-fix] `migrations/20260707000000_v0.7.0.sql` line 28

This amends a migration that already shipped on master in #133 (commit 30144c2). The runner records applied migrations by filename and skips them (`internal/isms/db/postgres.go:163-166`), so any database that migrated on master since #133 — dev, staging, self-hosted deploys tracking master — will never get this CHECK. The stated defense-in-depth only exists on fresh databases, and the test plan's `just test-reset` cannot detect that.

Verified with a live PoC:

```
# Step 1: migrate a fresh DB with master's migrations (= env that deployed #133)
  Applying 20260327000000_initial_schema.sql...
  Applying 20260707000000_v0.7.0.sql...        # master's version — no CHECK
# Step 2: migrate the same DB with this PR's migrations
  No pending migrations.                        # amended file skipped by filename
# Step 3:
SELECT conname FROM pg_constraint WHERE conname='change_requests_type_check';
(0 rows)                                        # CHECK silently missing
```

Suggested fix — move the constraint to a new migration file and revert the statement additions here (the comment update can stay):

```sql
-- migrations/20260708000000_change_type_check.sql
-- CHECK backstop for change_requests.type (defense-in-depth, matching every
-- other status/category column); extend by adding a value here + db.ChangeTypes.
ALTER TABLE change_requests DROP CONSTRAINT IF EXISTS change_requests_type_check;
ALTER TABLE change_requests ADD CONSTRAINT change_requests_type_check
    CHECK (type IN ('change', 'access_request'));
```

The `DROP CONSTRAINT IF EXISTS` guard makes it safe everywhere, and existing rows all satisfy it (`type` is `NOT NULL DEFAULT 'change'`, and every write path has been enum-validated since #133).

---

### 2. [should-fix] `internal/isms/db/changes.go` line 146

The empty→`'change'` defaulting here (and in `UpdateChangeRequestTx`, `suggestions_tx.go:466`) undermines the backstop this PR adds: a caller that builds a `ChangeRequest` without loading the row first silently reclassifies an `access_request` back to `change` instead of failing loudly on the CHECK. Both current callers load via `GetChangeRequest` first, so the branch is unreachable today — which is exactly when it's cheapest to remove.

Verified with a live PoC (Go test against a real DB, row seeded as `access_request`, then updated by a caller that set every field except `Type`):

```
before: type="access_request" title="Grant DB access to contractor"
after:  type="change" notes="updated notes only" (no error returned)
```

The fix was also validated end-to-end: with the 3-line default removed and the CHECK in place, the identical buggy caller now fails loudly and the data is preserved:

```
err=ERROR: new row for relation "change_requests" violates check constraint
"change_requests_type_check" (SQLSTATE 23514)
-- row still: type = access_request
```

Suggested fix — delete the defaulting from both update functions (keep it in `CreateChangeRequest`, where it's correct since there is no prior state):

```diff
 func (d *DB) UpdateChangeRequest(ctx context.Context, orgID int, id int, cr *ChangeRequest) error {
-	if cr.Type == "" {
-		cr.Type = "change"
-	}
 	_, err := d.pool.Exec(ctx, `
```

(same in `UpdateChangeRequestTx`). Note the loud failure depends on the CHECK actually existing — i.e. finding 1 — so the two fixes are complementary layers.

---

### 3. [nit] `tests/test_changes.py` line 152

Coverage for the API edit path is good. The suggestion-apply path (`applyChangeUpdate`) has its own validation branch for `type` but no test — `tests/test_suggestions.py` contains no `change_request` coverage at all, and no other test exercises type-via-suggestion. It's the second new write path this PR adds and the one agents (MCP) use. Suggested test:

```python
def test_suggestion_apply_change_type(self, api_url, admin_headers):
    r = requests.post(f"{api_url}/changes", headers=admin_headers, json={
        "title": "Reclassify via suggestion", "description": "x"})
    cid = r.json()["id"]
    r = requests.post(f"{api_url}/suggestions", headers=admin_headers, json={
        "entity_type": "change_request", "suggestion_type": "update",
        "entity_id": str(cid),
        "payload": {"fields": {"type": "access_request"}},
        "rationale": "misclassified"})
    sid = r.json()["id"]
    r = requests.post(f"{api_url}/suggestions/{sid}/apply", headers=admin_headers)
    r = requests.get(f"{api_url}/changes/{cid}", headers=admin_headers)
    assert r.json()["type"] == "access_request"
```

(Adjust the payload shape to match the existing suggestion tests — `test_suggestion_apply_task.py` is the closest template.)

---

### 4. [nit] `web/src/views/Changes.vue` line 203

The type options are hardcoded twice in this file — the create form (lines 31–32) and this edit form (lines 204–205), plus the display ternary at line 271. Consistent with how priority/category are done, so fine as-is — but when a third type is added (the CHECK comment anticipates it), there will be four places to update (CHECK, `db.ChangeTypes`, create form, edit form). A shared constant in the component script would cut that down:

```js
const CHANGE_TYPES = [
  { value: 'change', label: 'Change' },
  { value: 'access_request', label: 'Access request' },
]
```

```html
<option v-for="t in CHANGE_TYPES" :key="t.value" :value="t.value">{{ t.label }}</option>
```

The read-view ternary could reuse the same list via a small `typeLabel()` helper.

unidoc-ahall:
Thanks @unidoc-alip — all four addressed.

**1. CHECK never reaching applied DBs** — moved it to a new migration `20260707120000_change_type_check.sql` (`DROP CONSTRAINT IF EXISTS` + `ADD`), reverted the statements from the v0.7.0 file. A new filename runs on every DB, including ones that already ran v0.7.0. Good catch — the amend-in-place would've been a silent no-op on master-tracking deploys.
**2. empty→'change' silently reclassifies** — removed the default from `UpdateChangeRequest` and `UpdateChangeRequestTx` (kept in `CreateChangeRequest`). With the CHECK present, a blank type on update now fails loudly instead of rewriting `access_request` → `change`.
**3. suggestion-apply test** — added `test_suggestion_apply_change_type` (the MCP/agent path); reclassify-via-suggestion now covered.
**4. hardcoded options** — extracted `CHANGE_TYPES` + `typeLabel()` in Changes.vue; create form, edit form and the read-view label share one list.

`just test-go` green; `test_changes.py` 20/20 against the stack (×2).

Note: findings 1+2 mean 0.7.0 now carries a second migration file. That bends the one-migration-per-release convention, but it's forced by v0.7.0 having already merged to master — an amendment can't reach DBs that ran it. Happy to note this refinement in docs/releasing.md.

unidoc-alip (approved):
looks good to me

PR #140, opened 2026-07-07
fix(#26): suggestion-apply create matches HTTP defaults (slice A)

### LINKED ISSUE #139: #26 slice A: suggestion-apply create must match HTTP defaults
Part of the #26 enforced-write-path epic. An entity created via suggestion-apply started in a different default state than one created over HTTP:
- **asset**: apply defaulted status `active`, HTTP `open` (headline bug)
- **supplier**: apply didn't default owner
- **legal**: apply didn't default status/owner — and `CreateLegalRequirementTx` used an identifier key `legal` that violates `identifier_sequences`' CHECK, so applying a legal suggestion **always failed**
- **system**: apply didn't default status; HTTP didn't default classification/criticality
- **objective**: HTTP didn't default status/owner; apply had draft+actor

Fix: one `applyXDefaults(x, actor)` per entity, called by both the HTTP create handler and applyXCreate; legal identifier allocation goes through the shared `nextIdentifierTx("legal_requirement")` like HTTP.

unidoc-ahall:
Part of the #26 enforced-write-path epic (slice A: create consistency). Closes #139.

Suggestion-apply created entities in a **different default state** than HTTP. Now one `applyXDefaults(x, actor)` per entity is called by **both** the HTTP create handler and `applyXCreate`:

| entity | was | now |
|---|---|---|
| asset | apply `active` vs HTTP `open` | both `open` |
| supplier | apply had no owner | both owner→actor |
| legal | apply no status/owner **+ apply always failed** (bad identifier key) | both `open`+owner; shared `nextIdentifierTx('legal_requirement')` |
| system | apply no status; HTTP no classification/criticality | union both ways |
| objective | HTTP no status/owner; apply had draft+actor | both `draft`+owner |

## Test plan
- [ ] `just test-restart` then `just test tests/test_create_writepath.py` (green ×2 — already verified locally)
- [ ] create asset via a suggestion → status `open` (not `active`)
- [ ] apply a legal suggestion → succeeds (was always failing) with a `LEGAL-N` id

### DIFF
--- a/internal/isms/api/api_legal.go
+++ b/internal/isms/api/api_legal.go
@@ -124,18 +124,7 @@ func (s *Server) handleCreateLegal(c echo.Context) error {
 		Completion:        req.Completion,
 	}
 
-	if lr.Jurisdiction == "" {
-		lr.Jurisdiction = "EU"
-	}
-	if lr.Category == "" {
-		lr.Category = "privacy"
-	}
-	if lr.Owner == "" {
-		lr.Owner = getUserEmail(c)
-	}
-	if lr.Status == "" {
-		lr.Status = "open"
-	}
+	applyLegalDefaults(&lr, getUserEmail(c))
 	if err := validateEnum("status", lr.Status, db.LegalStatuses); err != nil {
 		return err
 	}
--- a/internal/isms/api/api_legal.go
+++ b/internal/isms/api/api_legal.go
@@ -125,13 +125,7 @@ func (s *Server) handleCreateLegal(c echo.Context) error {
 	}
 
 	applyLegalDefaults(&lr, getUserEmail(c))
-	if err := validateEnum("status", lr.Status, db.LegalStatuses); err != nil {
-		return err
-	}
-	if err := validateEnum("treatment", lr.Treatment, db.LegalTreatments); err != nil {
-		return err
-	}
-	if err := validateEnum("category", lr.Category, db.LegalCategories); err != nil {
+	if err := validateLegalCreate(&lr); err != nil {
 		return err
 	}
 
--- a/internal/isms/api/api_objectives.go
+++ b/internal/isms/api/api_objectives.go
@@ -309,12 +309,6 @@ func (s *Server) handleCreateObjective(c echo.Context) error {
 	if req.Title == "" {
 		return echo.NewHTTPError(http.StatusBadRequest, "title is required")
 	}
-	if err := validateEnum("status", req.Status, db.ObjectiveStatuses); err != nil {
-		return err
-	}
-	if err := validateEnum("target_operator", req.TargetOperator, db.ObjectiveTargetOperators); err != nil {
-		return err
-	}
 	// Cross-org parent guard: confirm the program belongs to this org.
 	if _, err := s.db.GetProgram(ctx, orgID, req.ProgramID); err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "program not found in this organization")
@@ -337,6 +331,9 @@ func (s *Server) handleCreateObjective(c echo.Context) error {
 		Notes:             req.Notes,
 	}
 	applyObjectiveDefaults(&o, getUserEmail(c))
+	if err := validateObjectiveCreate(&o); err != nil {
+		return err
+	}
 
 	if err := s.db.CreateObjective(ctx, orgID, &o); err != nil {
 		return pgxHTTPError(err)
--- a/internal/isms/api/api_objectives.go
+++ b/internal/isms/api/api_objectives.go
@@ -336,6 +336,7 @@ func (s *Server) handleCreateObjective(c echo.Context) error {
 		StartedAt:         req.StartedAt,
 		Notes:             req.Notes,
 	}
+	applyObjectiveDefaults(&o, getUserEmail(c))
 
 	if err := s.db.CreateObjective(ctx, orgID, &o); err != nil {
 		return pgxHTTPError(err)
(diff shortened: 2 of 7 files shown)


unidoc-alip (changes requested):
Solid slice A of the #26 epic — reviewed and then verified against a live local stack running this branch (fbc6a52): Postgres + server, the PR's four tests re-run green. Confirmed in code and at runtime that the old `'legal'` sequence key violates `identifier_sequences`' CHECK (only `legal_requirement` is allowed), so applying a legal suggestion always failed before, and the new allocator produces the same `LEGAL-N` format as HTTP. The fix is actually stronger than the description claims: the old apply-side asset default `active` violates `assets_status_check` (allowed: `draft/open/archived`), so asset apply-create was *completely broken* before this PR too — reproduced live against pre-PR code (500 `assets_status_check`), alongside the legal failure and the supplier NULL owner. Every new default is a valid member of its enum list, each helper is a faithful union of the two paths' previous behavior (empty-only, never overwriting input), and `go build`/`go vet` pass.

One cross-org concern was investigated and refuted, so for completeness: `applyObjectiveCreate` inserts `payload.ProgramID` without the HTTP handler's explicit org guard, but `CreateObjectiveTx` (`internal/isms/db/suggestions_tx.go:616`) resolves the program key for DisplayID generation with `SELECT key FROM programs WHERE id = $1 AND organization_id = $2`, which enforces tenant isolation on the apply path anyway — verified by PoC (org B applying a suggestion with org A's `program_id` fails, nothing persists). Only the error shape differs, which is finding 1's theme.

---

### 1. [should-fix] `internal/isms/api/api_writepath_defaults.go` line 5

Defaults are now shared, but enum **validation** still diverges: the HTTP create handlers call `validateEnum` after these defaults, while the `applyXCreate` handlers never validate. Verified with a live PoC on this branch — same invalid payload, two different outcomes:

- `POST /assets` with `asset_type: "laptop"` → **400** `invalid asset_type: "laptop" (allowed: [...])`
- suggestion-apply with the same payload → **500** `apply failed: apply: ERROR: new row for relation "assets" violates check constraint "assets_asset_type_check" (SQLSTATE 23514)`

No data-integrity risk — the DB CHECK constraints reject the row and the `WithOrgTx` transaction rolls back cleanly (nothing persists, the suggestion stays retryable). But a manager applying an agent-generated suggestion gets a raw SQL error instead of an actionable message, and an agent can't distinguish "your payload is invalid" from "the server is broken".

Suggested fix (either works; the first is more in the spirit of #26):

1. Give each helper a validation counterpart called by both paths — one enforced create path for defaults *and* validation:

   ```go
   // called by both the HTTP handler and applyAssetCreate, after applyAssetDefaults
   func validateAssetCreate(a *db.Asset) error {
       if err := validateEnum("status", a.Status, db.AssetStatuses); err != nil {
           return err
       }
       return validateEnum("asset_type", a.AssetType, db.AssetTypes)
   }
   ```

2. Minimally: route `applyXCreate` errors through `pgxHTTPError` (middleware.go:640 already maps CHECK violations to 400s) instead of the blanket `echo.NewHTTPError(500, "apply failed: ...")` at `api_suggestions.go:419`.

If validation is deliberately a later slice of #26, a TODO here documenting that would be enough for now.

---

### 2. [nit] `tests/test_create_writepath.py` line 1

Objective is in the PR's parity table — and it's the one entity where the **HTTP** path's behavior changed (status/owner now defaulted) — but it has no test here. Suggested addition, verified passing against this branch alongside the PR's four tests (5/5; note objectives match on `display_id`, not `identifier`, and `applied_entity_id` for objectives is the DisplayID):

```python
def test_objective_apply_matches_http(api_url, admin_headers):
    program_id = _ensure_program(api_url, admin_headers)  # first program, or create one

    # HTTP path: minimal payload, no status/owner
    r = requests.post(f"{api_url}/objectives", headers=admin_headers,
                      json={"title": f"wp-obj-http-{uuid.uuid4().hex[:8]}", "program_id": program_id})
    assert r.status_code in (200, 201), r.text
    http_obj = r.json()
    assert http_obj["status"] == "draft"
    assert http_obj.get("owner") == ADMIN_EMAIL

    # Apply path: same minimal payload via suggestion
    ident, name = _apply_create(api_url, admin_headers, "objective", "title",
                                {"program_id": program_id})
    r = requests.get(f"{api_url}/objectives", headers=admin_headers,
                     params={"q": name, "limit": 100})
    items = r.json().get("data", [])
    obj = next((x for x in items if x.get("display_id") == ident), None)
    assert obj is not None, f"objective {ident} not found"
    assert obj["status"] == http_obj["status"] == "draft"
    assert obj.get("owner") == http_obj.get("owner") == ADMIN_EMAIL
```

unidoc-ahall:
Thanks @unidoc-alip — both addressed, and good catch that asset apply-create was fully broken pre-PR too.

**1. Validation divergence** — went with your first option (the #26 spirit). Added `validateXCreate()` helpers (asset/supplier/legal/system/objective) called by **both** the HTTP handler and `applyXCreate`, right after `applyXDefaults` — a faithful mirror of the inline `validateEnum` calls. The apply dispatch now preserves an actionable status: an `*echo.HTTPError` (validation 400 with the allowed list) is returned as-is, DB constraint violations map to 400 via `pgxHTTPError`, only genuine failures stay 500. So the same invalid payload is now 400 on both paths with the same message.
**2. Objective test** — added `test_objective_apply_matches_http` (matching on `display_id`, as you noted) plus `test_apply_invalid_enum_is_400_not_500`.

`just test-go` green; `test_create_writepath.py` 6/6 against the stack (×2).

unidoc-alip (approved):
looks good to me

PR #142, opened 2026-07-07
fix(#26): audit-finding status transition via suggestion-apply (slice B)

### LINKED ISSUE #141: #26 slice B: audit-finding status transition via suggestion-apply
Part of the #26 enforced-write-path epic. Applying an audit-finding suggestion that changes status (e.g. to `closed`) failed: the apply path used a field-level update restricted to title/description, so `status` returned 'field not updatable' and the closure metadata (`closed_at`/`closed_by`) was never stamped. HTTP uses `SetAuditFindingStatus`.

Fix: a tx variant `SetAuditFindingStatusTx` (same closure logic as HTTP), and `applyAuditFindingUpdate` routes a `status` field through it (validated against AuditFindingStatuses).

unidoc-ahall:
Part of the #26 enforced-write-path epic (slice B: update consistency). Closes #141.

**Stacked on Slice A** (base = `feat/26-create-consistency`). Merge the Slice A PR first; this retargets to master afterwards.

Applying an audit-finding suggestion that changed `status` **errored** before — the apply path's field-level update rejected `status`, so closure metadata was never stamped. Now it routes through `SetAuditFindingStatusTx` (a tx variant of `SetAuditFindingStatus`), stamping `closed_at`/`closed_by` exactly like the HTTP handler.

Risk was checked and intentionally excluded: `accepted` is not a valid risk status (`RiskStatuses = draft/open/closed`), so the HTTP `status=='accepted'` block is vestigial — no real divergence.

## Test plan
- [ ] `just test-restart` then `just test tests/test_update_writepath.py` (green ×2 — already verified)
- [ ] apply an audit-finding suggestion with `{"status":"closed"}` → succeeds, `closed_at` set (errored before)

### DIFF
--- a/internal/isms/api/api_suggestions.go
+++ b/internal/isms/api/api_suggestions.go
@@ -1603,12 +1603,24 @@ func applyAuditFindingUpdate(ctx context.Context, tx pgx.Tx, s *Server, orgID in
 		return "", fmt.Errorf("invalid update payload: %w", err)
 	}
 	idInt, _ := parseEntityID(sg.EntityID)
-	// Update individual fields using existing field-level update
 	for field, val := range payload.Fields {
-		if sv, ok := val.(string); ok {
-			if err := db.UpdateAuditFindingFieldTx(ctx, tx, orgID, int(idInt), field, sv); err != nil {
-				return "", fmt.Errorf("updating field %s: %w", field, err)
+		sv, ok := val.(string)
+		if !ok {
+			continue
+		}
+		// Status transitions go through the shared closure-metadata path (same as
+		// the HTTP handler) — a plain field write would skip closed_at/closed_by.
+		if field == "status" {
+			if !db.AuditFindingStatuses[sv] {
+				return "", fmt.Errorf("invalid status: %s", sv)
 			}
+			if err := db.SetAuditFindingStatusTx(ctx, tx, orgID, int(idInt), sv, actor); err != nil {
+				return "", err
+			}
+			continue
+		}
+		if err := db.UpdateAuditFindingFieldTx(ctx, tx, orgID, int(idInt), field, sv); err != nil {
+			return "", fmt.Errorf("updating field %s: %w", field, err)
 		}
 	}
 	return sg.EntityID, nil
--- a/internal/isms/api/api_suggestions.go
+++ b/internal/isms/api/api_suggestions.go
@@ -1604,19 +1604,23 @@ func applyAuditFindingUpdate(ctx context.Context, tx pgx.Tx, s *Server, orgID in
 	}
 	idInt, _ := parseEntityID(sg.EntityID)
 	for field, val := range payload.Fields {
-		sv, ok := val.(string)
-		if !ok {
-			continue
-		}
 		// Status transitions go through the shared closure-metadata path (same as
 		// the HTTP handler) — a plain field write would skip closed_at/closed_by.
 		if field == "status" {
+			sv, ok := val.(string)
+			if !ok {
+				return "", fmt.Errorf("field status: expected a string, got %T", val)
+			}
 			if !db.AuditFindingStatuses[sv] {
 				return "", fmt.Errorf("invalid status: %s", sv)
 			}
 			if err := db.SetAuditFindingStatusTx(ctx, tx, orgID, int(idInt), sv, actor); err != nil {
 				return "", err
 			}
+			continue
+		}
+		sv, ok := val.(string)
+		if !ok {
 			continue
 		}
 		if err := db.UpdateAuditFindingFieldTx(ctx, tx, orgID, int(idInt), field, sv); err != nil {
--- a/internal/isms/db/audits.go
+++ b/internal/isms/db/audits.go
@@ -4,7 +4,15 @@ import (
 	"context"
 	"fmt"
 	"strings"
+
+	"github.com/jackc/pgx/v5/pgconn"
 )
+
+// pgExecer is satisfied by both *pgxpool.Pool and pgx.Tx, so a pool-based DB
+// method and its tx variant can share one query body (no duplicated SQL).
+type pgExecer interface {
+	Exec(ctx context.Context, sql string, args ...any) (pgconn.CommandTag, error)
+}
 
 // --- Status / type allowlists ---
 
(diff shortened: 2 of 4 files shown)


unidoc-alip (changes requested):
Slice B does exactly what it says: `applyAuditFindingUpdate` now routes `status` through the new `SetAuditFindingStatusTx`, which is a byte-for-byte mirror of the pool-based `SetAuditFindingStatus` — same validation against `AuditFindingStatuses`, same closure stamping/clearing, same org + soft-delete filters — so apply-close now matches the HTTP path, and the previously-erroring status apply works. RBAC parity holds too (apply requires admin/manager, same as the HTTP status endpoint). Security review found no issues: all new SQL is parameterized and the actor comes from the session. The risk `accepted` exclusion is correctly reasoned (`RiskStatuses` is draft/open/closed).

No blockers. Every finding below was verified against a fresh local stack built from this branch (live PoC where noted); the should-fix is worth doing while you're here.

Out-of-scope observation for the epic, not this PR: `handleUpdateAuditFinding` wraps its calls in `WithOrgTx` but then calls the pool-based `s.db.UpdateAuditFindingPartial`/`SetAuditFindingStatus` (ignoring the tx), so its "atomically per RLS" comment doesn't hold — a natural slice C candidate now that the Tx variant exists.

---

### 1. [should-fix] `internal/isms/db/suggestions_tx.go` line 755

Both branches of `SetAuditFindingStatusTx` ignore `RowsAffected`, so a status apply against a nonexistent or soft-deleted finding silently succeeds. On the HTTP path this can't happen — `handleUpdateAuditFindingStatus` 404s on `GetAuditFinding` before mutating. On the apply path there is no existence check at all, and the stale-detection guard is bypassed by `force=true` — which the stale response itself instructs users to send ("Re-submit with force=true to apply anyway").

Verified live: create finding → suggest `{"status":"closed"}` → `DELETE /audit-findings/:id` (200) → apply with `force=true` returned `200 {"status":"applied"}`; the finding stayed deleted (GET 404) and its changelog now reads `suggestion_applied, delete, create` — an applied-mutation record stamped after the delete, for a mutation that never touched the entity.

`UpdateAuditFindingPartial` in this package already handles this case, so there's local precedent — the same check here closes the gap:

```go
res, err := tx.Exec(ctx, `...`)
if err != nil {
    return err
}
if res.RowsAffected() == 0 {
    return fmt.Errorf("finding not found")
}
return nil
```

One caveat: the error reaches the client as 500 ("apply failed: apply: finding not found") since `handleApplyEntitySuggestion` maps all txErr to 500 — still strictly better than a false "applied". Mirroring the check into the pool-based `SetAuditFindingStatus` is optional (HTTP is protected by its `GetAuditFinding` pre-check).

---

### 2. [nit] `internal/isms/db/suggestions_tx.go` line 752

`SetAuditFindingStatusTx` duplicates the SQL of `SetAuditFindingStatus` (audits.go) verbatim — confirmed by diffing the two bodies with the receiver normalized: zero differences. A future change to closure metadata now has to be made in two places, which is the exact divergence class #26 exists to eliminate. Both `*pgxpool.Pool` and `pgx.Tx` expose a compatible `Exec`, so one unexported core keeps a single source of truth:

```go
type pgExecer interface {
    Exec(ctx context.Context, sql string, args ...any) (pgconn.CommandTag, error)
}

func setAuditFindingStatus(ctx context.Context, e pgExecer, orgID, id int, status, closedBy string) error { ... }
```

`(d *DB) SetAuditFindingStatus` passes `d.pool`, the Tx variant passes `tx`. Fine as a follow-up given this file's existing Tx-variant convention.

---

### 3. [nit] `internal/isms/api/api_suggestions.go` line 1578

A non-string field value is silently skipped and the suggestion still gets marked `applied`. Verified live: a suggestion with `{"status": 123}` applied with HTTP 200 while the finding stayed `open` with no closure metadata — a silent no-op recorded as applied. That was the old behavior too, but now that `status` has real semantics here, returning an error would be more honest:

```go
sv, ok := val.(string)
if !ok {
    return "", fmt.Errorf("field %s: expected string value", field)
}
```

If erroring on every non-string field feels too broad a behavior change, scope the check to `status` only.

---

### 4. [nit] `tests/test_update_writepath.py` line 55

Only the close direction is covered — the reopen branch of `SetAuditFindingStatusTx` (clearing `closed_at`/`closed_by`/`closed_by_user_id`) is new code with no test. The path was exercised live (close-then-reopen via suggestions correctly cleared the closure fields), so this test passes as-is:

```python
def test_audit_finding_apply_reopen_clears_closure(api_url, admin_headers):
    # (reuse the same programme/audit/finding setup as the close test)
    _suggest_update(api_url, admin_headers, "audit_finding", fid, {"status": "closed"})
    _suggest_update(api_url, admin_headers, "audit_finding", fid, {"status": "open"})
    got = requests.get(f"{api_url}/audit-findings/{fid}", headers=admin_headers).json()
    assert got["status"] == "open", got
    assert not got.get("closed_at"), "reopen must clear closed_at"
    assert not got.get("closed_by"), "reopen must clear closed_by"
```

unidoc-ahall:
Thanks @unidoc-alip — should-fix + all three nits done.

**1. RowsAffected** — `SetAuditFindingStatusTx` (now the shared core) errors when no row matched, so a status apply against a deleted/nonexistent finding no longer records a false 'applied'. As you noted it surfaces as 500 via the current dispatch — strictly better than a false success; the cleaner 4xx mapping is the same dispatch work as PR #140's finding 1.
**2. Dedup** — extracted one `setAuditFindingStatus(ctx, pgExecer, ...)` core (the `pgExecer` interface you suggested); `DB.SetAuditFindingStatus` passes the pool, `SetAuditFindingStatusTx` passes the tx. Single source for the closure logic + the not-found guard.
**3. Non-string status** — now errors (`field status: expected a string, got %T`) instead of a silent applied no-op; scoped to `status` so other fields' behaviour is unchanged.
**4. Reopen test** — added `test_audit_finding_apply_reopen_clears_closure`.

Also filed the out-of-scope observation mentally for slice C: `handleUpdateAuditFinding` wraps pool calls in `WithOrgTx` (ignoring the tx) — worth switching to the tx variants now they exist.

`just test-go` green; `test_update_writepath.py` 2/2 against the stack (×2).

unidoc-alip (dismissed):
looks good to me

PR #144, opened 2026-07-07
fix(#26): change_request status via suggestion-apply (slice C) — completes the epic

### LINKED ISSUE #143: #26 slice C: change_request status transition via suggestion-apply
Final slice of the #26 enforced-write-path epic. `applyChangeUpdate` handled type/priority/risk/rollback/assigned but silently ignored a `status` field (`UpdateChangeRequestTx` doesn't write status), so a suggestion/agent could never transition a change and `approved_at`/`approved_by` / `implemented_at` were never derived. HTTP routes status through `UpdateChangeRequestStatus`.

Fix: `UpdateChangeRequestStatusTx` (tx variant, same approved/implemented metadata + reverse-transition clearing); `applyChangeUpdate` routes a `status` field through it (validated against ChangeStatuses).

With this, #26 is complete: incident+CA (#123), asset/supplier/legal/system/objective create (slice A), audit_finding update (slice B), change_request status (this). risk + task need no change.

unidoc-ahall:
Closes #143. **Final slice of the #26 enforced-write-path epic.**

`applyChangeUpdate` silently ignored a `status` field (UpdateChangeRequestTx doesn't touch status), so a suggestion could never transition a change and `approved_at`/`approved_by` / `implemented_at` were never derived. Now a `status` field routes through `UpdateChangeRequestStatusTx` (tx variant of the HTTP path's `UpdateChangeRequestStatus`).

**#26 coverage after this:**
| entity | where |
|---|---|
| incident, corrective_action | #123 |
| asset, supplier, legal, system, objective (create) | slice A (#142) |
| audit_finding (update) | slice B |
| change_request (status) | this |
| risk, task | no change needed (verified) |

## Test plan
- [ ] `just test-restart` then `just test tests/test_changes.py` — incl. `test_suggestion_apply_status_stamps_approved` (green ×2 verified)
- [ ] apply a change suggestion with `{"status":"approved"}` → `approved_at` set (was ignored before)

### DIFF
--- a/internal/isms/api/api_suggestions.go
+++ b/internal/isms/api/api_suggestions.go
@@ -1102,6 +1102,20 @@ func applyChangeUpdate(ctx context.Context, tx pgx.Tx, s *Server, orgID int, sg 
 			cr.AssignedTo = sv
 		}
 	}
+	// Status transitions go through the shared enforced path so approved_at/by and
+	// implemented_at are derived exactly as the HTTP handler does — a plain field
+	// write (UpdateChangeRequestTx doesn't touch status) would skip that metadata.
+	if v, ok := payload.Fields["status"]; ok {
+		if sv, ok := v.(string); ok && sv != cr.Status {
+			if err := validateEnum("status", sv, db.ChangeStatuses); err != nil {
+				return "", err
+			}
+			if err := db.UpdateChangeRequestStatusTx(ctx, tx, orgID, cr.ID, sv, actor); err != nil {
+				return "", err
+			}
+			cr.Status = sv
+		}
+	}
 	if err := db.UpdateChangeRequestTx(ctx, tx, orgID, cr.ID, cr); err != nil {
 		return "", err
 	}
--- a/internal/isms/api/api_suggestions.go
+++ b/internal/isms/api/api_suggestions.go
@@ -337,6 +337,21 @@ func (s *Server) handleClaimEntitySuggestion(c echo.Context) error {
 	return c.JSON(http.StatusOK, map[string]string{"status": newStatus})
 }
 
+// applyPostCommitKey carries a queue of side-effect callbacks that must run only
+// AFTER the apply transaction commits. Apply handlers use registerApplyPostCommit
+// for effects that touch the pool directly or send mail (e.g. auto-created tasks),
+// which must not fire inside WithOrgTx — mirroring how the HTTP handlers run their
+// post-commit work. Keeps the HTTP and suggestion-apply paths behaving identically.
+type applyPostCommitKey struct{}
+
+// registerApplyPostCommit queues fn to run after the enclosing apply tx commits.
+// A no-op if the context carries no queue (defensive; the apply path always sets one).
+func registerApplyPostCommit(ctx context.Context, fn func()) {
+	if q, ok := ctx.Value(applyPostCommitKey{}).(*[]func()); ok {
+		*q = append(*q, fn)
+	}
+}
+
 func (s *Server) handleApplyEntitySuggestion(c echo.Context) error {
 	if err := requireRole(c, "admin", "manager"); err != nil {
 		return err
@@ -344,6 +359,11 @@ func (s *Server) handleApplyEntitySuggestion(c echo.Context) error {
 	orgID := getOrgID(c)
 	ctx := c.Request().Context()
 	actor := getUserEmail(c)
+
+	// Callbacks registered by apply handlers for side effects that must run only
+	// after the tx commits (task auto-creation, mail). Drained below on success.
+	var postCommit []func()
+	ctx = context.WithValue(ctx, applyPostCommitKey{}, &postCommit)
 
 	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
 	if err != nil {
@@ -419,7 +439,12 @@ func (s *Server) handleApplyEntitySuggestion(c echo.Context) error {
 		return echo.NewHTTPError(http.StatusInternalServerError, "apply failed: "+txErr.Error())
 	}
 
-	// Post-commit: activity log + notifications (non-transactional, OK to fail)
+	// Post-commit: run side effects handlers deferred until the tx committed
+	// (auto-created tasks, mail), then the activity log + notifications. All
+	// non-transactional and OK to fail — the entity mutation is already durable.
+	for _, fn := range postCommit {
+		fn()
+	}
 	s.logAndNotify(ctx, orgID, &db.Activity{
 		Actor:  actor,
 		Action: "suggestion_applied",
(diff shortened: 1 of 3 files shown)


unidoc-alip (changes requested):
Solid, well-scoped fix: `UpdateChangeRequestStatusTx` mirrors the HTTP `UpdateChangeRequestStatus` logic exactly (same approve/implement/reverse-transition branches, same org-scoped parameterized SQL), and the new test confirms `approved_at` is now stamped via suggestion-apply. Security review found no injection, authz-bypass, or org-isolation issues — all values are parameterized and `status` is enum-validated before use. Both findings below were verified live against a running build of this exact commit (abe65e6), with PoC transcripts included.

---

### 1. [should-fix] `internal/isms/api/api_suggestions.go` line 1116

The HTTP paths (`handleUpdateChange` and `handleUpdateChangeStatus` in `api_collab.go`) both call `s.createChangeFollowupTask(...)` when a change request transitions into `approved`, auto-creating an "Implement `<CR>`" task assigned to the requester/assignee. This suggestion-apply path doesn't call it, so approving a change via suggestion-apply silently skips that side effect — directly contradicting the comment above it ("...are derived exactly as the HTTP handler does") and the #26 epic's acceptance criterion that HTTP and suggestion-apply paths behave identically. Status transitions via suggestion-apply were previously a no-op (silently ignored), so this PR is what newly exposes the gap.

**Verified live** (local Postgres + server built from this commit, admin user, fresh org): traced `handleApplyEntitySuggestion` (api_suggestions.go:340-435) to confirm no hidden hook calls `createChangeFollowupTask` on change_request's behalf — the only two call sites are the two HTTP handlers in `api_collab.go`.

**PoC** — two change requests, same transition (`proposed` → `approved`), one via each path:

```
# Path A — HTTP status endpoint (baseline)
POST /api/v1/changes {"title":"CR via HTTP status endpoint", ...}   → CR-1 (id=1)
PUT  /api/v1/changes/1/status {"status":"approved"}                 → 200 {"status":"approved"}

GET /api/v1/tasks
→ TASK-1 "Implement CR-1: CR via HTTP status endpoint" (task_type: change_followup)   ← created, as expected

# Path B — suggestion-apply (this PR's new code path)
POST /api/v1/changes {"title":"CR via suggestion-apply", ...}       → CR-2 (id=2)
POST /api/v1/suggestions {
  "entity_type":"change_request","suggestion_type":"update","entity_id":"2",
  "payload":{"fields":{"status":"approved"}}, "rationale":"ready","title":"approve"
}                                                                    → suggestion #1
POST /api/v1/suggestions/1/apply {"force":true}                     → {"status":"applied","applied_entity_id":"CR-2"}

GET /api/v1/changes/2
→ {"status":"approved","approved_by":"poc-admin@example.com","approved_at":1783419981, ...}
  (this PR's fix DID work — approved_at/approved_by are correctly stamped)

GET /api/v1/tasks
→ still only TASK-1. No "Implement CR-2: ..." task was created.
```

**Suggested fix**: `createChangeFollowupTask` isn't safe to call as-is from inside `applyChangeUpdate`'s transaction — it uses `s.db.CreateTask` / `s.logChange` / `s.searchUpsert` / `s.mailer.Send...` directly against the pool (not `tx`-aware) and sends email, none of which should fire before the enclosing `WithOrgTx` (api_suggestions.go:390) actually commits. Recommended shape, mirroring the existing post-commit block at api_suggestions.go:422-429 (`logAndNotify`/`notifySuggestionResolved` already run after the tx succeeds):

```go
// 1. applyChangeUpdate (or the apply-handler contract) signals that an approval
//    transition happened, e.g. via a small post-commit callback queue threaded
//    through the tx context or the Server:
if err := db.UpdateChangeRequestStatusTx(ctx, tx, orgID, cr.ID, sv, actor); err != nil {
    return "", err
}
cr.Status = sv
if sv == "approved" {
    queuePostCommit(ctx, func() { s.createChangeFollowupTask(context.Background(), orgID, cr, actor) })
}

// 2. handleApplyEntitySuggestion runs queued callbacks right after txErr == nil,
//    alongside the existing logAndNotify/notifySuggestionResolved calls.
```

If follow-up-task auto-creation is intentionally an HTTP-only convenience (not a compliance-enforced rule), the alternative is to explicitly document that scoping decision against the #26 tracking issue rather than leave it as an undocumented gap, since the epic's stated acceptance criterion is "tests assert identical behavior ... across the HTTP and suggestion-apply paths."

---

### 2. [nit] `tests/test_changes.py` line 152

Only the approve path is tested. `UpdateChangeRequestStatusTx` duplicates the reverse-transition clearing logic (`approved`→`proposed`/`in_progress`/`rejected` clears `approved_at`/`approved_by`, and clears `implemented_at`) from `UpdateChangeRequestStatus` — a test covering a reverse transition via suggestion-apply would catch future drift between the two hand-duplicated functions (this duplication itself matches existing codebase convention across ~45 `*Tx` variants, so that part isn't a new problem).

**Verified live**: this is a coverage gap only, not a bug. Approved CR-2 from finding #1's PoC, then submitted a second suggestion transitioning it `approved` → `rejected` via suggestion-apply — `approved_at`/`approved_by` were correctly cleared, matching `UpdateChangeRequestStatusTx`'s `clearApproved` branch exactly as designed.

**PoC**:

```
# CR-2 is currently "approved" (from finding #1's PoC)
POST /api/v1/suggestions {
  "entity_type":"change_request","suggestion_type":"update","entity_id":"2",
  "payload":{"fields":{"status":"rejected"}}, "rationale":"reverse","title":"reject"
}                                                                → suggestion #2
POST /api/v1/suggestions/2/apply {"force":true}                 → {"status":"applied","applied_entity_id":"CR-2"}

GET /api/v1/changes/2
→ {"status":"rejected", ...}   (no approved_by / approved_at keys present — correctly cleared)
```

**Suggested fix**: add `test_suggestion_apply_status_clears_approved_on_reverse_transition` to `tests/test_changes.py` alongside `test_suggestion_apply_status_stamps_approved`: create a change, approve it via suggestion-apply (asserting `approved_at` is set), then submit and apply a second suggestion transitioning it to `rejected` (or `proposed`/`in_progress`), and assert `approved_at`/`approved_by` are absent/null afterward. This locks in the parity this PR establishes.

unidoc-ahall:
Thanks @unidoc-alip — should-fix + nit both done, and confirmed the reproduction you traced.

**1. Follow-up task parity** — you were right that this newly exposed the gap, and right that `createChangeFollowupTask` can't run inside the tx (it uses the pool + sends mail). Implemented the post-commit-queue shape you suggested: a callback list threaded through the request context (`registerApplyPostCommit`), `applyChangeUpdate` enqueues `createChangeFollowupTask` on an approval transition, and `handleApplyEntitySuggestion` drains it right after `txErr == nil`, next to the existing `logAndNotify`/`notifySuggestionResolved` block. So the task now fires post-commit for both paths — the generic queue also gives future apply handlers a clean home for the same pattern. Went with wiring it (not documenting-as-HTTP-only) since the epic's acceptance criterion is identical behaviour across paths.
**2. Reverse-transition test** — added `test_suggestion_apply_status_clears_approved_on_reverse_transition` (approve then reject via apply, asserts approved_at/by cleared), plus `test_suggestion_apply_approval_creates_followup_task` locking in finding 1.

`just test-go` green; the three change-status apply tests pass ×2 against a freshly restarted stack.

unidoc-alip (approved):
looks good to me

PR #146, opened 2026-07-07
feat(cli): CIA rating flags on supplier add/edit

### LINKED ISSUE #145: CLI: expose CIA rating flags on isms supplier add / edit
Supplier CIA rating (Confidentiality/Integrity/Availability, 0-5, NULL = not assessed) is a first-class column, on the Supplier struct, in the API create+update DTOs, and in the Suppliers web view — but the CLI doesn't expose it, so `isms supplier add` can't record a rating and `edit` can't change one. Operators seeding suppliers CLI-first (STS core suppliers) have to open the web UI or leave it NULL.

Add `--confidentiality` / `--integrity` / `--availability` (int 0-5) to `supplier add` and `edit`: set the pointer only when passed (unset stays NULL), validate 0-5 before the API call. No schema/API changes.

unidoc-ahall:
Closes #145. From the work order in temp/issue-supplier-cia-flags.md.

Supplier CIA rating (C/I/A, 0-5, NULL = not assessed) was first-class everywhere except the CLI. Adds `--confidentiality` / `--integrity` / `--availability` to `supplier add` and `edit`.

- Pointer set only when the flag is passed → unset stays NULL. The update DTO's `**int` treats a null rating as skip, so `supplier edit --criticality high` never disturbs an existing rating.
- 0-5 validated at the CLI before the API call; DB CHECK is the backstop.
- CLI-only — no schema, no API changes.

## Test plan
- [ ] `just test-go` green (supplier CIA CLI tests: sends ratings / null on unset / rejects out-of-range) — verified
- [ ] `isms supplier add --name X --type saas --criticality high --confidentiality 3 --integrity 4 --availability 5` → ratings recorded
- [ ] `isms supplier edit <id> --criticality high` leaves an existing rating untouched

### DIFF
--- a/cmd/isms/cli_relations_test.go
+++ b/cmd/isms/cli_relations_test.go
@@ -168,3 +168,44 @@ func TestReviewApproveUsesApprovalEndpoint(t *testing.T) {
 		t.Errorf("comment not forwarded: body=%v", r.body)
 	}
 }
+
+func TestSupplierAddSendsCIA(t *testing.T) {
+	srv, got := cliServer(t)
+	defer srv.Close()
+	cmd := supplierCmd()
+	cmd.SetArgs([]string{"add", "--name", "Acme", "--type", "saas", "--criticality", "high",
+		"--confidentiality", "3", "--integrity", "4", "--availability", "5"})
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	b := (*got)[len(*got)-1].body
+	if b["confidentiality"] != float64(3) || b["integrity"] != float64(4) || b["availability"] != float64(5) {
+		t.Errorf("CIA ratings not sent: %+v", b)
+	}
+}
+
+func TestSupplierAddOmitsUnsetCIA(t *testing.T) {
+	srv, got := cliServer(t)
+	defer srv.Close()
+	cmd := supplierCmd()
+	cmd.SetArgs([]string{"add", "--name", "Acme", "--type", "saas", "--criticality", "high"})
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("execute: %v", err)
+	}
+	// Unset → null (not assessed); the **int update DTO treats null as skip.
+	if b := (*got)[len(*got)-1].body; b["confidentiality"] != nil {
+		t.Errorf("unset confidentiality should be null, got %v", b["confidentiality"])
+	}
+}
+
+func TestSupplierAddRejectsOutOfRangeCIA(t *testing.T) {
+	srv, _ := cliServer(t)
+	defer srv.Close()
+	cmd := supplierCmd()
+	cmd.SilenceUsage, cmd.SilenceErrors = true, true
+	cmd.SetArgs([]string{"add", "--name", "Acme", "--type", "saas", "--criticality", "high",
+		"--confidentiality", "9"})
+	if err := cmd.Execute(); err == nil {
+		t.Error("expected an error for --confidentiality 9 (out of 0-5)")
+	}
+}
--- a/cmd/isms/cli_relations_test.go
+++ b/cmd/isms/cli_relations_test.go
@@ -209,3 +209,15 @@ func TestSupplierAddRejectsOutOfRangeCIA(t *testing.T) {
 		t.Error("expected an error for --confidentiality 9 (out of 0-5)")
 	}
 }
+
+func TestSupplierAddRejectsNegativeCIA(t *testing.T) {
+	srv, _ := cliServer(t)
+	defer srv.Close()
+	cmd := supplierCmd()
+	cmd.SilenceUsage, cmd.SilenceErrors = true, true
+	cmd.SetArgs([]string{"add", "--name", "Acme", "--type", "saas", "--criticality", "high",
+		"--confidentiality", "-1"})
+	if err := cmd.Execute(); err == nil {
+		t.Error("expected an error for --confidentiality -1 (out of 0-5)")
+	}
+}
--- a/cmd/isms/supplier.go
+++ b/cmd/isms/supplier.go
@@ -21,13 +21,16 @@ func supplierCmd() *cobra.Command {
 
 func supplierAddCmd() *cobra.Command {
 	var (
-		name         string
-		supplierType string
-		criticality  string
-		dataAccess   bool
-		contact      string
-		contractRef  string
-		notes        string
+		name            string
+		supplierType    string
+		criticality     string
+		dataAccess      bool
+		contact         string
+		contractRef     string
+		notes           string
+		confidentiality int
+		integrity       int
+		availability    int
 	)
 
 	cmd := &cobra.Command{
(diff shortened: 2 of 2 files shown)


unidoc-alip (changes requested):
Clean, narrowly-scoped CLI change: `--confidentiality/--integrity/--availability` flags on `supplier add`/`edit`, closing #145. Verified the core claim in the PR description empirically — Go's `**int` DTO field genuinely collapses "JSON null" and "key absent" to the same "skip" behavior for a top-level struct field, so an unset rating flag really does leave an existing rating untouched. DB CHECK (`BETWEEN 0 AND 5`) backstop confirmed in the initial schema migration. No security concerns — client-side validation is a UX nicety, the server enforces the real invariant. Two minor test-coverage/consistency nits below, plus one pre-existing (not introduced by this PR) latent issue in the same function worth a follow-up ticket.

---

### 1. [should-fix] `cmd/isms/cli_relations_test.go` line 201

`TestSupplierAddRejectsOutOfRangeCIA` only exercises the upper bound (`9`). `validateCIA` also rejects `v < 0`, but that branch has no test — worth adding a `--confidentiality -1` case alongside this one so the lower-bound check doesn't silently regress.

Verified: ran `--confidentiality -1` directly through validateCIA/pflag — it parses correctly as the flag's value (not swallowed as a shorthand flag) and is correctly rejected with "--confidentiality must be between 0 and 5". Behavior is already correct; this is purely a coverage gap.

Suggested fix (add to `cmd/isms/cli_relations_test.go`, next to the existing out-of-range test):

```go
func TestSupplierAddRejectsNegativeCIA(t *testing.T) {
	srv, _ := cliServer(t)
	defer srv.Close()
	cmd := supplierCmd()
	cmd.SilenceUsage, cmd.SilenceErrors = true, true
	cmd.SetArgs([]string{"add", "--name", "Acme", "--type", "saas", "--criticality", "high",
		"--confidentiality", "-1"})
	if err := cmd.Execute(); err == nil {
		t.Error("expected an error for --confidentiality -1 (out of 0-5)")
	}
}
```

---

### 2. [nit] `cmd/isms/supplier.go` line 224

`supplier edit`'s flag help text drops the "unset = not assessed" hint that `supplier add` has (line 84). Worth keeping consistent since it's the one piece of documentation a user sees for the NULL semantics.

Suggested fix (`cmd/isms/supplier.go:224-226`):

```go
cmd.Flags().IntVar(&confidentiality, "confidentiality", 0, "Confidentiality rating (0-5, unset = not assessed)")
cmd.Flags().IntVar(&integrity, "integrity", 0, "Integrity rating (0-5, unset = not assessed)")
cmd.Flags().IntVar(&availability, "availability", 0, "Availability rating (0-5, unset = not assessed)")
```

---

### 3. [should-fix] `cmd/isms/supplier.go` line 180

Not introduced by this PR, but worth flagging since it's the exact function this diff extends and you clearly already reasoned carefully about skip-vs-null semantics here: `update := &db.Supplier{}` builds a fresh zero-value struct and only overwrites fields whose flags were `Changed()`. That pattern is safe for the three new CIA fields (server DTO uses `**int`, which — as verified — collapses both "absent" and explicit "null" to "skip"). But `SupplierType`, `Criticality`, `Name`, `DataAccess`, and `Status` have no `omitempty` on `db.Supplier` and the server DTO uses single pointers (`*string`/`*bool`) for them — so an edit that only touches `--criticality` still sends `"supplier_type":"", "name":"", "status":"", "data_access":false` on the wire, and the server unmarshals those into non-nil pointers pointing at the zero value, which get written straight through (`if req.SupplierType != nil { updated.SupplierType = *req.SupplierType }`).

Verified empirically (via a throwaway unit test hitting `supplierEditCmd` with only `--criticality` set): `isms supplier edit <id> --criticality high` sends `supplier_type:""`, `name:""`, `status:""`, and `data_access:false` in the request body today. That silently wipes those fields on every partial edit that doesn't also pass them. Also confirmed this is a repo-wide pattern, not supplier-specific — `cmd/isms/asset.go`'s `assetEditCmd` has the identical `update := &db.Asset{}` + conditional-field-set shape, so it likely has the same bug. This predates PR #146 (the field-copying logic for Name/SupplierType/DataAccess/Status wasn't touched by this diff), so it's not a blocker here, and it doesn't affect the new CIA fields' correctness — the double-pointer DTO shape genuinely protects them. Given the author already dug into these exact null/skip mechanics for CIA, this is a natural companion fix for a follow-up, ideally shared across supplier.go and asset.go.

Note: `data_access` can't just get `omitempty` on the current `bool` field — that would make it impossible to ever explicitly turn data-access *off* via edit (false is `omitempty`'s zero value too), trading one silent-corruption bug for a silent-no-op bug.

Suggested fix — stop building the wire payload from `db.Supplier{}` zero values; use a dedicated partial-update struct with pointer fields (mirroring what the server's own `supplierUpdateRequest` already does), and only set the pointers that were `Changed()`:

```go
type supplierEditPayload struct {
	Name            *string `json:"name,omitempty"`
	SupplierType    *string `json:"supplier_type,omitempty"`
	Criticality     *string `json:"criticality,omitempty"`
	DataAccess      *bool   `json:"data_access,omitempty"`
	Contact         *string `json:"contact,omitempty"`
	Notes           *string `json:"notes,omitempty"`
	Confidentiality *int    `json:"confidentiality,omitempty"`
	Integrity       *int    `json:"integrity,omitempty"`
	Availability    *int    `json:"availability,omitempty"`
}
// ...
update := &supplierEditPayload{}
if cmd.Flags().Changed("name") {
	update.Name = &name
}
if cmd.Flags().Changed("data-access") {
	update.DataAccess = &dataAccess
}
// ... etc for the rest
```

`*bool` + `omitempty` works correctly here (unlike a raw `bool` field) since `omitempty` only drops the field when the *pointer itself* is nil — a non-nil `*bool` pointing at `false` still marshals as `false`, so explicit off-switches survive. `c.UpdateSupplier`'s signature would need to accept this payload type (or an `interface{}`) instead of `*db.Supplier`.

unidoc-ahall:
Thanks @unidoc-alip — findings 1 & 2 done, and filed finding 3.

**1. Negative bound** — added `TestSupplierAddRejectsNegativeCIA` (`--confidentiality -1`); confirmed pflag parses `-1` as the value (not a shorthand) and validateCIA rejects it.
**2. Edit help text** — restored `unset = not assessed` on the three `supplier edit` CIA flags, consistent with `supplier add`.
**3. Partial-edit field wipe** — verified exactly as you described (`db.Supplier` lacks omitempty on name/type/status/data_access + the CLI sends the struct directly + single-pointer DTO). Since it's pre-existing, spans `asset edit` too, and needs a client-signature change, I kept it out of this CIA PR and filed it as #147 (with the shared supplier+asset fix + the `*bool` data_access caveat). The CIA fields themselves are unaffected — the `**int` DTO genuinely protects them.

`just test-go` green.

unidoc-alip (approved):
Looks good to me

PR #149, opened 2026-07-07
fix(#127): Programs register + repo-wide status-badge consistency

### LINKED ISSUE #127: Objectives/Programs use a non-standard UI (no left nav, different creation) — align with the register pattern
Programs, objectives and check-ins all live in Objectives.vue under tab-nested routes (/objectives/:tab/:objId), unlike the standard registers (Risks, Incidents,
  …) which use /X + /X/:id with the left nav and the standard create flow. Result: no left bar and a different creation UX.

  First step: a component audit of Objectives.vue vs the standard register pattern (layout, left nav, list+detail, create modal) to list exactly what to align. Acceptance:
  objectives/programs match the standard register layout and creation flow.

unidoc-ahall:
Closes #127.

## Programs -> its own register
Programs lived as a tab inside `Objectives.vue` - no left-nav entry, no route, a bare table with inline Edit/Delete. Now:
- **New `Programs.vue`** - register header + Add + SuggestNewButton, search, standard table, detail slide-over (overview/notes/links/suggestions/comments/history + delete), create modal. Modeled on `Suppliers.vue`.
- **Routes** `/programs` + `/programs/:id`; **sidebar** entry; e2e smoke.
- **Objectives.vue** de-tabbed into a single clean register (Add in header, deep-link `/objectives/:objId`).
- Reuses existing `api.js` program CRUD; references key on the program `key`.

## Repo-wide status-badge consistency
A consistency audit found five views rendering status with local `statusClass` maps that drifted from the shared `<StatusBadge>` in their own detail panels - worst: **Incidents showed `open` red in the list but blue in the detail** (same status, two colours, one screen).
- `<StatusBadge>` is now the single source of truth: Incidents, Corrective Actions, Objectives, Tasks, Changes list rows all use it.
- Added the two missing cases (`todo`, `cancelled`) so nothing greys out.
- Removed the dead local `statusClass`/`statusDot` functions; completed the App.vue nav list.
- **Table-vs-card layouts are intentional** (registers vs operations modules) and left unchanged.

## Verify
`just build-web` green across all touched views; Programs chunk emitted; no dangling status refs. e2e runs in CI.

### DIFF
--- a/internal/isms/api/api_objectives.go
+++ b/internal/isms/api/api_objectives.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"bytes"
+	"context"
 	"crypto/sha256"
 	"fmt"
 	"io"
@@ -144,11 +145,26 @@ func (s *Server) handleCreateProgram(c echo.Context) error {
 	return c.JSON(http.StatusCreated, p)
 }
 
+// resolveProgramID accepts either a numeric program id or a program key (e.g.
+// "AWARE"). Cross-entity reference chips link programs by key, while the register's
+// own list links by id — both must reach these endpoints. Mirrors the id-or-key
+// fallback the reference title resolver already uses.
+func (s *Server) resolveProgramID(ctx context.Context, orgID int, raw string) (int64, error) {
+	if id, err := strconv.ParseInt(raw, 10, 64); err == nil {
+		return id, nil
+	}
+	p, err := s.db.GetProgramByKey(ctx, orgID, raw)
+	if err != nil {
+		return 0, err
+	}
+	return p.ID, nil
+}
+
 func (s *Server) handleGetProgram(c echo.Context) error {
 	orgID := getOrgID(c)
-	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	id, err := s.resolveProgramID(c.Request().Context(), orgID, c.Param("id"))
 	if err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, "invalid id")
+		return echo.NewHTTPError(http.StatusNotFound, "program not found")
 	}
 	p, err := s.db.GetProgram(c.Request().Context(), orgID, id)
 	if err != nil {
@@ -163,9 +179,9 @@ func (s *Server) handleUpdateProgram(c echo.Context) error {
 	}
 	orgID := getOrgID(c)
 	ctx := c.Request().Context()
-	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	id, err := s.resolveProgramID(ctx, orgID, c.Param("id"))
 	if err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, "invalid id")
+		return echo.NewHTTPError(http.StatusNotFound, "program not found")
 	}
 
 	old, err := s.db.GetProgram(ctx, orgID, id)
@@ -217,9 +233,9 @@ func (s *Server) handleDeleteProgram(c echo.Context) error {
 	}
 	orgID := getOrgID(c)
 	ctx := c.Request().Context()
-	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	id, err := s.resolveProgramID(ctx, orgID, c.Param("id"))
 	if err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, "invalid id")
+		return echo.NewHTTPError(http.StatusNotFound, "program not found")
 	}
 
 	if err := s.db.DeleteProgram(ctx, orgID, id); err != nil {
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -112,6 +112,7 @@ class TestSmoke:
         ("Corrective Actions", "Corrective"),
         ("Incidents", "Incident"),
         ("Objectives", "Objectives"),
+        ("Programs", "Program Register"),
         ("Tasks", "Tasks"),
         ("Inbox", "Inbox"),
         ("Admin", "Members"),
--- a/tests/test_objectives.py
+++ b/tests/test_objectives.py
@@ -37,6 +37,25 @@ class TestPrograms:
         })
         assert r.status_code == 403
 
+    def test_get_resolves_id_or_key(self, api_url, admin_headers):
+        """#149: program endpoints accept a numeric id OR the program key —
+        cross-entity reference chips deep-link programs by key."""
+        key = "BYKEY2026"
+        requests.post(f"{api_url}/programs", headers=admin_headers,
+                      json={"title": "By-key resolver test", "key": key})  # idempotent
+        progs = requests.get(f"{api_url}/programs", headers=admin_headers).json().get("data") or []
+        p = next((x for x in progs if x.get("key") == key), None)
+        assert p is not None, "program missing from list after create"
+        # by key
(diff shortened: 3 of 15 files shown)


unidoc-alip (changes requested):
Programs register and status-badge consolidation both look solid and well modeled on `Suppliers.vue`. But the route restructuring (`/objectives/:tab/:objId` → `/objectives/:objId`, plus new `/programs/:id`) wasn't propagated to the shared components that build cross-entity deep links. Three findings below are live, confirmed regressions; the fourth has the same stale mapping but is currently unreachable. All were re-verified against the current code (`router.js`, `api_objectives.go`, `api_search.go`, `api_references.go`, `Programs.vue`) before filing.

---

### 1. [must-fix] `web/src/components/EntityReferences.vue` line 141

This objective deep-link is dead: `router.js` changed `/objectives/:tab/:objId` to `/objectives/:objId`, so `/objectives/objectives/${numId}` no longer matches any route. Clicking an "OBJ" reference chip anywhere in the app (risks, suppliers, incidents, etc. that link to an objective) will silently fail to navigate.

Confirmed against the current `router.js` — only `/objectives` and `/objectives/:objId` remain, no two-segment pattern.

Suggested fix:
```js
// web/src/components/EntityReferences.vue:141
- return orgPath(`/objectives/objectives/${numId}`)
+ return orgPath(`/objectives/${numId}`)
```

---

### 2. [must-fix] `web/src/components/EntityReferences.vue` line 101

`program` still maps to the `objectives` route base here, and `program` isn't in the `deepLinkTypes` array above either, so it falls through to a bare `/objectives` with no id — clicking a "PROG" reference chip lands on the objectives list, not the program.

Fixing the route base alone isn't enough: `Programs.vue`'s route (`/programs/:id`) and its `handleGetProgram`/`handleUpdateProgram`/`handleDeleteProgram` backend handlers (`internal/isms/api/api_objectives.go`) do `strconv.ParseInt(c.Param("id"), ...)` and 400 on non-numeric input. But `Programs.vue` itself uses `selected.key` (e.g. "AWARE") as the `entityId` for its `ReferenceManager`/`SuggestionPanel`/`CommentsPanel` panels, so program references are keyed by the string `key`, not the numeric id. `Programs.vue`'s own `openFromRoute(id)` also does `parseInt(id)` before lookup, so a key like "AWARE" becomes `NaN` and the fetch fails silently. The backend's generic reference-title resolver (`api_references.go`, `resolveEntityTitle`) already handles this dual-identifier reality via a ParseInt-then-`GetProgramByKey` fallback — the direct program CRUD endpoints don't.

Needs a 3-part fix:

1. Backend — give the program-by-id endpoints the same key fallback as `resolveEntityTitle`:
```go
func (s *Server) resolveProgramID(ctx context.Context, orgID int, raw string) (int64, error) {
    if id, err := strconv.ParseInt(raw, 10, 64); err == nil {
        return id, nil
    }
    p, err := s.db.GetProgramByKey(ctx, orgID, raw)
    if err != nil {
        return 0, err
    }
    return p.ID, nil
}
```
Use it in `handleGetProgram`, `handleUpdateProgram`, and `handleDeleteProgram` instead of the bare `strconv.ParseInt` call, returning 404 (not 400) when resolution fails either way.

2. Frontend `web/src/views/Programs.vue` — `openFromRoute` currently discards non-numeric ids. Match by key too and pass the raw id through:
```js
async function openFromRoute(id) {
  const numId = parseInt(id)
  let p = programs.value.find(x => x.id === numId || x.key === id)
  if (!p) {
    try { p = await api.getProgram(id) } catch { return }
  }
  if (!p) return
  ...
}
```

3. This file — fix the map entry and add `program` to the deep-link branch (the existing `.replace(/^[A-Z]+-/, '')` is a harmless no-op on dash-less keys like "AWARE"):
```js
// line 101
- program: 'objectives',
+ program: 'programs',
```
```js
// deepLinkTypes array a few lines up
- const deepLinkTypes = ['change', 'incident', 'corrective_action', 'supplier', 'asset', 'task', 'risk', 'legal', 'system']
+ const deepLinkTypes = ['change', 'incident', 'corrective_action', 'supplier', 'asset', 'task', 'risk', 'legal', 'system', 'program']
```

---

### 3. [must-fix] `web/src/components/ReferenceManager.vue` line 123

Same stale mapping (`program: 'objectives'`), used by `refRoute()` (line 137) to build reference chip links inside the Links tab of every other entity. This one is live and reachable today: `Programs.vue`'s own "Links" tab renders `<ReferenceManager entityType="program" :entityId="selected.key" :editable="canWrite" />`, so users already create program↔other-entity references, and any entity on the other side (a risk, supplier, incident, etc.) renders the program back as a "PROG" chip through this same stale map.

Suggested fix (same map, copy-pasted into a third file):
```js
// web/src/components/ReferenceManager.vue:123
- objective: 'objectives', program: 'objectives', task: 'tasks',
+ objective: 'objectives', program: 'programs', task: 'tasks',
```
This alone still 400s until the backend key-fallback from finding #2 lands, since `refRoute()` here builds `/programs/<key>` via the same generic id logic.

Longer-term: consider extracting this entity-type→route-base map (and the sibling type-label/type-color maps) into one shared composable imported by `EntityReferences.vue`, `ReferenceManager.vue`, and `GlobalSearch.vue`, so a future route change only needs updating in one place instead of three.

---

### 4. [nit] `web/src/components/GlobalSearch.vue` line 139

Same stale `program: 'objectives'` mapping, but currently dead code: the backend's `buildSearchIndex` (`internal/isms/api/api_search.go`) never indexes programs or objectives at all (only documents, risks, legal, suppliers, assets, systems, incidents, tasks, changes, corrective actions), and nothing calls `searchUpsert`/`searchRemove` for programs either. So `item.type === 'program'` can't actually occur in global search results today. Worth fixing for consistency, but not blocking.

```js
// web/src/components/GlobalSearch.vue:139
- program: 'objectives',
+ program: 'programs',
```
Flag `navigate()`'s numeric-id assumption here too if programs are ever added to the search index — it'll need the same key→id resolution as `EntityReferences.vue`.

unidoc-ahall:
Thanks @unidoc-alip — sharp catch, all four fixed. The route change should have been grepped across the shared components; it wasn't.

**1 — EntityReferences objective chip** → `/objectives/${id}` (was the dead `/objectives/objectives/`).
**2 — program deep-link** → map corrected to `programs` + `program` added to `deepLinkTypes`. And you're right that programs are dual-identifier: reference chips key by `key`, the register lists by `id`. Added a backend `resolveProgramID` (ParseInt → `GetProgramByKey` fallback, returns 404 not 400) used by `handleGetProgram`/`handleUpdateProgram`/`handleDeleteProgram`; `Programs.vue` `openFromRoute` now matches by `key` too and passes the raw id through.
**3 — ReferenceManager** map → `programs` (the live one via the Programs Links tab).
**4 — GlobalSearch** map → `programs`.

Also swept `web/src` for any other `/objectives/objectives` or `program: 'objectives'` — none remain. Added a test: `/programs/<key>` and `/programs/<id>` both resolve, unknown key → 404.

On the longer-term suggestion — extracting the entity-type→route/label/colour maps into one shared composable (the triple duplication is exactly what let this drift) — agreed; I'll file that as a follow-up so it's not lost.

`just test-go` + `just build-web` green.

unidoc-alip (approved):
looks good to me

PR #151, opened 2026-07-07
fix(#147): partial supplier/asset edit no longer wipes unset fields

### LINKED ISSUE #147: CLI: partial `supplier edit` / `asset edit` silently wipes unset fields
Found during review of #146 (@unidoc-alip, finding 3 — pre-existing, not introduced by that PR).

## Problem
`isms supplier edit <id> --criticality high` silently wipes `name`, `supplier_type`, `status`, and `data_access` on the target supplier. Any partial edit that doesn't re-pass every field blanks the ones it omits.

## Root cause
- `db.Supplier` gives `Name`, `SupplierType`, `Criticality`, `DataAccess`, `Status` **no `omitempty`**, so they always serialize.
- The CLI (`supplierEditCmd`) builds `update := &db.Supplier{}` and `client.UpdateSupplier` sends that struct **directly** — so omitted flags go on the wire as `"name":""`, `"supplier_type":""`, `"status":""`, `"data_access":false`.
- The server DTO `supplierUpdateRequest` uses **single** pointers (`*string`/`*bool`) for these, so a present-but-empty value unmarshals to a non-nil pointer at the zero value and is written straight through.

The CIA ratings are safe — they use `**int` (null == skip), which is exactly why #146's new flags are correct. This bug is only the plain string/bool fields.

## Scope
Same `update := &db.Asset{}` shape exists in `cmd/isms/asset.go` (`assetEditCmd`) → almost certainly the same wipe. Fix both.

## Fix
Stop building the wire payload from a zero-value `db.X`. Use a dedicated partial-update struct with pointer fields + `omitempty`, set only the flags that were `Changed()`, and change `client.UpdateSupplier`/`UpdateAsset` to accept it. Note: `data_access` must be `*bool` (not raw `bool`+omitempty), else an explicit `--data-access=false` becomes a silent no-op.

## Acceptance
- `supplier edit <id> --criticality high` leaves name/type/status/data_access untouched (regression test asserting the wire body omits unchanged fields).
- Same for `asset edit`.

unidoc-ahall:
Closes #147.

Re-land of the partial-edit-wipe fix. The original PR was stacked on #146 and got closed unmerged when #146 squash-merged, so the fix never reached master — this re-opens it fresh on top of master.

## Bug
`isms supplier edit <id> --criticality high` (and the identical `asset edit`) sent a zero-value `db.Supplier`/`db.Asset`; `name`/`supplier_type`/`status`/`data_access` have no omitempty, so the server wrote the blanks over existing data. Every partial edit blanked the fields it didn't re-pass.

## Fix
- Dedicated pointer-payload structs (`supplierEditPayload` / `assetEditPayload`) with omitempty, populated only for `Changed()` flags → untouched fields are omitted from the wire.
- `client.UpdateSupplier`/`UpdateAsset` take `any`; the full-object import path and manager flow are unchanged.
- `data_access` is `*bool` so an explicit `--data-access=false` turns it off instead of being dropped.
- Regression tests: partial edit omits unchanged fields (supplier + asset), and `--data-access=false` is sent.

`just test-go` green.

### DIFF
--- a/cmd/isms/asset.go
+++ b/cmd/isms/asset.go
@@ -125,6 +125,24 @@ func assetListCmd() *cobra.Command {
 	return cmd
 }
 
+// assetEditPayload is the partial-update wire shape for `asset edit` — pointer
+// fields with omitempty so an unset flag is omitted rather than sent as a zero
+// value the server would write over an existing value (#147). Mirrors the
+// server's assetUpdateRequest.
+type assetEditPayload struct {
+	Name            *string   `json:"name,omitempty"`
+	AssetType       *string   `json:"asset_type,omitempty"`
+	Owner           *string   `json:"owner,omitempty"`
+	Description     *string   `json:"description,omitempty"`
+	Status          *string   `json:"status,omitempty"`
+	PrimaryLocation *string   `json:"primary_location,omitempty"`
+	Confidentiality *int      `json:"confidentiality,omitempty"`
+	Integrity       *int      `json:"integrity,omitempty"`
+	Availability    *int      `json:"availability,omitempty"`
+	NextReview      *db.Epoch `json:"next_review,omitempty"`
+	Notes           *string   `json:"notes,omitempty"`
+}
+
 func assetEditCmd() *cobra.Command {
 	var (
 		name            string
@@ -148,24 +166,26 @@ func assetEditCmd() *cobra.Command {
 			id := args[0]
 
 			c := requireAPI()
-			update := &db.Asset{}
+			// Only fields whose flag was set go on the wire; everything else is
+			// omitted so a partial edit never blanks an untouched field (#147).
+			update := &assetEditPayload{}
 			if cmd.Flags().Changed("name") {
-				update.Name = name
+				update.Name = &name
 			}
 			if cmd.Flags().Changed("type") {
-				update.AssetType = assetType
+				update.AssetType = &assetType
 			}
 			if cmd.Flags().Changed("owner") {
-				update.Owner = owner
+				update.Owner = &owner
 			}
 			if cmd.Flags().Changed("desc") {
-				update.Description = description
+				update.Description = &description
 			}
 			if cmd.Flags().Changed("status") {
-				update.Status = status
+				update.Status = &status
 			}
 			if cmd.Flags().Changed("location") {
-				update.PrimaryLocation = primaryLocation
+				update.PrimaryLocation = &primaryLocation
 			}
 			if cmd.Flags().Changed("confidentiality") {
 				update.Confidentiality = &confidentiality
@@ -184,7 +204,7 @@ func assetEditCmd() *cobra.Command {
 				update.NextReview = rd
 			}
 			if cmd.Flags().Changed("notes") {
-				update.Notes = notes
+				update.Notes = &notes
 			}
 			if _, err := c.UpdateAsset(id, update); err != nil {
 				return err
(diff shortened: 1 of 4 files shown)


unidoc-alip (approved):
Correct, narrowly-scoped fix for #147 (re-land of the fix originally reviewed as part of the #146 stack, now rebased fresh on master). `supplierEditPayload`/`assetEditPayload` use pointer fields with `omitempty`, populated only for `Changed()` flags, so an untouched field is genuinely dropped from the wire — verified against the server's `supplierUpdateRequest`/`assetUpdateRequest` handlers (`internal/isms/api/server.go`), and by building the branch and running `go test ./cmd/isms/... -run 'TestSupplierEdit|TestAssetEdit'` plus the full `cmd/isms` package suite, all green. `data_access` correctly uses `*bool` so an explicit `--data-access=false` is sent and processed. `client.UpdateAsset`/`UpdateSupplier` widening to `any` is safe — `Client.put` already took `interface{}`, and the only other caller (`supplier import`, cmd/isms/supplier.go:390) still passes a full `*db.Supplier` for the full-object PUT path, unchanged. No new security surface. Two non-blocking notes below.

---

### 1. [should-fix] `cmd/isms/supplier.go` line 161

`supplierEditPayload` (and `assetEditPayload` in `asset.go:132`) manually mirror the server's `supplierUpdateRequest`/`assetUpdateRequest` DTO shape field-by-field with no compile-time or test-time link between them. If the server DTO gains, renames, or drops a field later, nothing here breaks at compile time — the CLI will just silently fail to send (or keep sending) that field.

Suggested — cross-reference the DTO location directly in each payload struct's doc comment so a future DTO change is at least discoverable:

```go
// cmd/isms/asset.go — extend the existing doc comment on assetEditPayload:
// Mirrors the server's assetUpdateRequest (internal/isms/api/api_register_dtos.go:28) —
// if that DTO gains/renames/drops a field, update this struct and assetEditCmd to match.

// cmd/isms/supplier.go — extend the existing doc comment on supplierEditPayload:
// Mirrors the server's supplierUpdateRequest (internal/isms/api/api_register_dtos.go:64) —
// if that DTO gains/renames/drops a field, update this struct and supplierEditCmd to match.
```

Optional follow-up (not required for this PR): a reflection-based test asserting every JSON tag on `assetEditPayload`/`supplierEditPayload` exists on the corresponding server DTO, turning future drift into a test failure instead of a silent gap.

---

### 2. [nit] `cmd/isms/cli_relations_test.go` line 239

`TestSupplierEditOmitsUnchangedFields` only asserts `name/supplier_type/data_access/contact/notes` are omitted — it doesn't cover `confidentiality/integrity/availability`. Same gap in `TestAssetEditOmitsUnchangedFields` (line 274) for `confidentiality/integrity/availability/notes/next_review`. Not required to close #147 (the acceptance criteria's core fields are covered), but rounding these out would catch a future regression on the CIA/notes/next_review fields specifically.

Suggested — extend the existing omission-check slices in place; the same CLI invocation already leaves those fields unset, so no new test functions are needed:

```go
// TestSupplierEditOmitsUnchangedFields, in the for loop:
for _, k := range []string{"name", "supplier_type", "data_access", "contact", "notes",
    "confidentiality", "integrity", "availability"} {
    if _, ok := b[k]; ok {
        t.Errorf("unchanged field %q must be omitted, got %v", k, b[k])
    }
}

// TestAssetEditOmitsUnchangedFields, in the for loop:
for _, k := range []string{"name", "asset_type", "owner", "description", "primary_location",
    "confidentiality", "integrity", "availability", "notes", "next_review"} {
    if _, ok := b[k]; ok {
        t.Errorf("unchanged field %q must be omitted, got %v", k, b[k])
    }
}
```

PR #152, opened 2026-07-07
chore(devenv): just test-setup + mount isms-python in dev container

unidoc-ahall:
Two small devenv changes that were sitting outstanding.

- **`just test-setup`** — reproducibly (re)creates the Python test venv from `tests/requirements.txt`, exactly like CI does. A hand-made, never-upgraded `.venv` (old playwright wanting a Chromium build the image never had) is what broke local e2e; this is the missing "recreate it fresh" step. Nothing is version-pinned — the browser is always whatever the venv's own playwright resolves, which matches the image's baked build.
- **compose.yml** — mount `${ISMS_SRC_ROOT}/isms-python` into `isms-dev` so cross-repo work on the Python client is available in the container (same pattern as the isms / isms-templates mounts).

No app code touched.

### DIFF
--- a/devenv/Justfile
+++ b/devenv/Justfile
@@ -92,6 +92,22 @@ run +CMD:
 
 # ── Test environment ─────────────────────────────────────────────────────────
 
+# (Re)create the Python test venv from tests/requirements.txt — the same thing CI
+# does, so the .venv can't silently go stale (a hand-made, never-upgraded .venv is
+# what broke local e2e: it held an old playwright wanting a Chromium build the image
+# never had). `-U` pulls the current playwright, whose Chromium build matches the one
+# baked into the image (devenv/Dockerfile bakes it via the same-minor @playwright/test).
+#
+# If playwright ever ships a NEW minor before the image is rebuilt, its Chromium won't
+# be in the image yet — then fetch the exact matching build once, as root:
+#   just attach-root
+#   PLAYWRIGHT_BROWSERS_PATH=/opt/ms-playwright /home/isms/workspace/isms/.venv/bin/playwright install --with-deps chromium && chmod -R a+rX /opt/ms-playwright
+# (No version is hard-coded anywhere — the browser is always whatever the venv's own
+# playwright asks for.)
+test-setup:
+    {{compose}} exec -u isms isms bash -lc \
+        "cd /home/isms/workspace/isms && python3 -m venv .venv && .venv/bin/pip install --quiet -U -r tests/requirements.txt && .venv/bin/playwright --version"
+
 # Run pytest from inside the dev container against the test server.
 # Rebuilds web/dist first: isms-test serves the bundle live from disk, so an
 # e2e test would otherwise silently run against a stale UI after a .vue change
--- a/devenv/Justfile
+++ b/devenv/Justfile
@@ -24,10 +24,12 @@ compose := env_var_or_default('COMPOSE_CMD', 'docker compose')
 # ── Lifecycle ────────────────────────────────────────────────────────────────
 
 # Ensure optional bind-mount sources exist (compose errors if they don't).
-# isms-demo is private prospect content — contributors without it get an
-# empty dir so the mount resolves and the container can still boot.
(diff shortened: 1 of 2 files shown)


unidoc-alip (changes requested):
Solid, well-documented fix for the stale-`.venv`/Chromium-mismatch problem — the `test-setup` recipe and its comment clearly explain the failure mode and the escape hatch for a playwright minor-version drift. One issue blocks: the new `isms-python` bind mount isn't backed by the same `_ensure-paths` safety net `isms-demo` gets, and it'll break `just up`/`just build` for contributors who haven't cloned that repo.

---

### 1. [must-fix] `devenv/compose.yml` line 42

This bind mount has no `_ensure-paths` fallback, and `isms-python` isn't referenced anywhere else in the compose file or env config (no `ISMS_*_PATH` var, unlike `isms-templates` via `ISMS_TEMPLATE_PATH`). That makes it look optional — like `isms-demo` — rather than required.

The existing comment on `_ensure-paths` (`devenv/Justfile`) states plainly: "Ensure optional bind-mount sources exist (compose errors if they don't)." Contributors who haven't cloned `isms-python` will hit that same `just up`/`just build` failure that `_ensure-paths` was specifically introduced to prevent for `isms-demo`.

Verified: `_ensure-paths` still only creates `isms-demo`; `isms-python` has no other reference in the repo (confirmed via grep), unlike `isms-templates` which is load-bearing via `ISMS_TEMPLATE_PATH`; and `git log -p` confirms `_ensure-paths` was added by the maintainers specifically because "compose errors if [optional bind-mount sources] don't [exist]."

Suggested fix — add `isms-python` to `_ensure-paths`, same treatment as `isms-demo`:

```diff
 # Ensure optional bind-mount sources exist (compose errors if they don't).
 # isms-demo is private prospect content — contributors without it get an
 # empty dir so the mount resolves and the container can still boot.
 _ensure-paths:
     @mkdir -p "$ISMS_SRC_ROOT/isms-demo"
+    @mkdir -p "$ISMS_SRC_ROOT/isms-python"
```

If `isms-python` is instead meant to become a hard requirement for every contributor going forward, skip the mkdir and call that out explicitly in `devenv/README.md` instead — but that's a bigger behavioral change than "no app code touched" suggests, so the mkdir is the safer default.

---

### 2. [nit] `devenv/Justfile` line 108

`python3 -m venv .venv` on an already-existing `.venv` doesn't `--clear` it first, so a `.venv` left over from a different Python version (mismatched symlinks) wouldn't be reliably fixed by re-running `test-setup`. Not a regression from this PR, just worth tightening since "make it reproducible" is the explicit goal here.

Verified empirically: a scratch venv with `requests==2.25.0` installed still reported that version after a bare re-run of `python3 -m venv`; re-running with `--clear` wiped it back to "not found." Practical impact here is limited since `pip install -U -r tests/requirements.txt` re-pins the packages that matter (pytest/requests/playwright) regardless — the gap is only for stray packages outside `tests/requirements.txt`.

Suggested fix:

```diff
 test-setup:
     {{compose}} exec -u isms isms bash -lc \
-        "cd /home/isms/workspace/isms && python3 -m venv .venv && .venv/bin/pip install --quiet -U -r tests/requirements.txt && .venv/bin/playwright --version"
+        "cd /home/isms/workspace/isms && python3 -m venv --clear .venv && .venv/bin/pip install --quiet -U -r tests/requirements.txt && .venv/bin/playwright --version"
```

unidoc-ahall:
Thanks @unidoc-alip — both done.

**1. `_ensure-paths`** — added `mkdir -p "$ISMS_SRC_ROOT/isms-python"` right alongside `isms-demo`, so the new mount gets the same optional-source safety net and won't break `just up`/`just build` for anyone who hasn't cloned it. (Keeping it optional, not a hard requirement — matches the "no behavioral change" intent.)
**2. `--clear`** — `test-setup` now runs `python3 -m venv --clear .venv`, so a re-run reliably rebuilds a venv left over from a different Python version, not just re-pins the tracked deps.

unidoc-alip (approved):
looks good to me

PR #158, opened 2026-07-10
feat(ui): roll slim click-to-filter stat strip out to all registers

unidoc-ahall:
Rolls the slim, always-visible **StatStrip** (introduced for Tasks in #157) out to the remaining registers, closing #156.

Each register opened with a grid of big stat cards that pushed the list well below the fold. This swaps them for the shared chip strip: count + label, click-to-filter, active chip highlighted — same data, same filtering, a fraction of the height.

**Converted:** Suppliers, Assets, Legal, Systems, Incidents, Corrective Actions, Objectives, Changes, Risks.

**Notes**
- `StatStrip` gains an optional per-chip `static` flag — display-only chips for derived metrics that aren't a filter dimension (Critical / Severe / High).
- **Risks:** the four summary chips span different filter dimensions (level vs status), so they're display-only; the Risk Map stays in its collapsible "More".
- **Corrective Actions:** its summary cards were plain (non-clickable) before — they're now click-to-filter, consistent with the rest.
- **Audit** is intentionally left out: it has two stat sections plus a boolean "overdue" toggle mixed into the chips, which doesn't fit the single-`v-model` strip. Can follow up separately if wanted.

`just build-web` green. No API/logic changes beyond wiring the existing filter refs to the strip.

### DIFF
--- a/web/src/components/StatStrip.vue
+++ b/web/src/components/StatStrip.vue
@@ -2,26 +2,38 @@
   <!-- Slim, always-visible register stat strip. Each stat is click-to-filter:
        clicking toggles the active key via v-model. Replaces the tall stat-card
        grids at the top of the registers so the list gets the vertical space,
-       while counts + filtering stay glanceable. -->
+       while counts + filtering stay glanceable.
+
+       A stat with `static: true` renders as a display-only chip (no click) — for
+       derived metrics that aren't a filter dimension (e.g. Critical / Severe). -->
   <div class="flex flex-wrap gap-2">
-    <button
-      v-for="s in stats"
-      :key="s.key"
-      type="button"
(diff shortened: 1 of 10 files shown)


unidoc-alip (approved):
Clean, mechanical rollout of the shared `StatStrip` component (from #157) across all nine remaining registers, closing #156. `StatStrip` gained a well-scoped `static` flag for display-only chips (Critical/Severe/High), and every view's `statusStats`/`summaryStats` computed correctly mirrors the pre-existing `stats` fields, filter keys, and colors from the card grid it replaces — no dropped states, no orphaned `filterStatus` wiring, no `v-html`/new input surfaces. Verified `npm run build` is green and traced each converted view's fallback logic (`stats.total || total || 0`, etc.) against the prior markup; all matches. 

No must-fix or should-fix issues found.

LGTM

PR #165, opened 2026-07-14
feat(ui): refresh button on all list pages

### LINKED ISSUE #164: Refresh button on list pages (no more F5)
There's no way to refresh a list when someone else has changed data — you have to hit F5, which reloads the whole SPA (loses scroll position, selection, filters, and does a fresh auth round-trip).

**Change:** a small shared **RefreshButton** on every list page — registers (Tasks, Risks, Suppliers, Assets, Legal, Systems, Incidents, Corrective Actions, Changes, Objectives, Programs, Audit) plus Reviews and Inbox. The icon spins while the page re-fetches its list + stats in place; scroll/selection/filters are preserved. No full reload.

Dashboard is intentionally left for a follow-up (it's a summary that loads on navigation, not a stale-list surface).

unidoc-ahall:
Closes #164 — there's no way to refresh a list without F5 (which reloads the whole SPA and loses scroll/selection/filters).

Adds a shared **`components/RefreshButton.vue`** and wires it into every list page:
- **Registers:** Tasks, Risks, Suppliers, Assets, Legal, Systems, Incidents, Corrective Actions, Changes, Objectives, Programs, Audit.
- **Also:** Reviews, Inbox.

Each view gets a `reload()` that re-fetches its list (the register loaders already cascade to stats) using a dedicated `refreshing` ref — the icon spins, the button disables, and the list updates in place. No full reload, no auth round-trip, scroll/selection/filters preserved.

**Dashboard** is intentionally deferred — it's a summary that loads on navigation, not a stale-list surface. Easy follow-up if wanted.

`just build-web` green.

### DIFF
--- a/web/src/components/RefreshButton.vue
+++ b/web/src/components/RefreshButton.vue
@@ -1,0 +1,21 @@
+<template>
+  <!-- Reload the current page's list(s) without a full F5. Icon spins + button
+       disables while loading. Wire @refresh to the view's loader and :loading to
+       its loading ref. -->
+  <button
+    type="button"
+    :disabled="loading"
+    @click="$emit('refresh')"
+    class="inline-flex items-center justify-center rounded-lg border border-slate-800 bg-slate-900 p-1.5 text-slate-400 transition-colors hover:border-slate-700 hover:text-slate-200 disabled:opacity-60 disabled:cursor-not-allowed"
+    :title="loading ? 'Refreshing…' : 'Refresh'"
+    aria-label="Refresh">
+    <svg class="h-4 w-4" :class="{ 'animate-spin': loading }" fill="none" viewBox="0 0 24 24" stroke="currentColor" stroke-width="2">
+      <path stroke-linecap="round" stroke-linejoin="round" d="M4 4v5h.582m15.356 2A8.001 8.001 0 004.582 9m0 0H9m11 11v-5h-.581m0 0a8.003 8.003 0 01-15.357-2m15.357 2H15" />
+    </svg>
+  </button>
+</template>
+
+<script setup>
+defineProps({ loading: { type: Boolean, default: false } })
+defineEmits(['refresh'])
+</script>
--- a/web/src/views/Assets.vue
+++ b/web/src/views/Assets.vue
@@ -8,8 +8,9 @@
 
     <!-- Error -->
     <div v-else-if="error" class="max-w-5xl mx-auto px-8 py-12">
-      <div class="bg-red-950/40 border border-red-900/50 rounded-lg p-6 text-red-300 text-sm">
-        {{ error }}
+      <div class="bg-red-950/40 border border-red-900/50 rounded-lg p-6 text-red-300 text-sm flex items-center justify-between gap-4">
+        <span>{{ error }}</span>
+        <RefreshButton :loading="refreshing" @refresh="reload" />
       </div>
     </div>
 
@@ -437,8 +438,11 @@ const loading = ref(true)
 const refreshing = ref(false)
 async function reload() {
   refreshing.value = true
+  error.value = null
   try {
     await loadAssets()
+  } catch (e) {
+    error.value = e.message
   } finally {
     refreshing.value = false
   }
--- a/web/src/views/Assets.vue
+++ b/web/src/views/Assets.vue
@@ -80,6 +80,7 @@
       <!-- Search + Filters -->
       <div class="space-y-3">
         <div class="flex flex-wrap items-center gap-3">
+          <RefreshButton :loading="refreshing" @refresh="reload" class="shrink-0" />
           <div class="relative flex-1 max-w-xs">
             <svg class="absolute left-3 top-1/2 -translate-y-1/2 w-4 h-4 text-slate-500" fill="none" viewBox="0 0 24 24" stroke="currentColor" stroke-width="2">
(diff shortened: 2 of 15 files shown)


unidoc-alip (changes requested):
Clean, mechanical rollout of a shared `RefreshButton.vue` wired into 14 list views, matching issue #164's scope exactly. Verification turned up one real must-fix: in four views, `reload()` routes through a `loading`-toggling helper, which brings back the exact full-list-disappears flicker the PR is meant to eliminate. Also flagging an unaddressed gap (button unreachable after a failed initial load, on 12 of 14 views) and a minor duplicate-fetch nit. No security concerns — this is pure UI re-fetch wiring, no new endpoints or unescaped content.

---

### 1. [must-fix] `web/src/views/CorrectiveActions.vue` line 571

`reload()` calls `await loadAll()`, but `loadAll()` itself does `loading.value = true` / `loading.value = false` around the fetch. The template is `v-if="loading"` → `ListSkeleton` / `v-else-if="error"` / `v-else` → main content (which is where `RefreshButton` and the list live). So every refresh click flips `loading` back to `true`, unmounts the entire list + the `RefreshButton` itself, and shows the skeleton in its place — the same "list disappears, scroll resets" behavior the PR is meant to eliminate. The other views affected by the identical pattern: `Incidents.vue` (`loadAll` at line 644), `Objectives.vue` (`loadAll` at line 623), `Programs.vue` (`loadAll` at line 342).

Suggested fix — extract the fetch body out of `loadAll()` so `reload()` can call it without touching `loading`:

```js
async function loadAll() {
  loading.value = true
  error.value = null
  try {
    await fetchAll()
  } catch (e) {
    error.value = e.message
  } finally {
    loading.value = false
  }
}

async function fetchAll() {
  await Promise.all([loadActions(), loadStats()])
}

async function reload() {
  refreshing.value = true
  try {
    await fetchAll()
  } finally {
    refreshing.value = false
  }
}
```

`loadActions()`/`loadStats()` already have their own try/catch that sets `error.value`, so error handling on refresh is unaffected. Apply the same split in `Incidents.vue` (`fetchAll` = `Promise.all([loadIncidents(), loadStats()])`), `Objectives.vue` (`fetchAll` = the `api.getPrograms()` + `loadObjectives()` body), and `Programs.vue` (extract the `api.getPrograms()` call into a `fetchPrograms()` that `reload()` can call directly).

---

### 2. [should-fix] `web/src/views/Assets.vue` line 10

The template gates on `v-if="loading"` / `v-else-if="error"` / `v-else` (main content), and `RefreshButton` only lives in the `v-else` (success) branch. So if the *initial* list load fails, the user lands on the error screen with no `RefreshButton` in sight — the only way to retry is still a full F5, which is exactly the problem issue #164 was filed to fix. This affects every view using this same three-way template split: Assets, Audit, CorrectiveActions, Incidents, Legal, Objectives, Programs, Reviews, Risks, Suppliers, Systems, and Inbox (Changes.vue and Tasks.vue are fine — they use toast-based errors instead of a blocking error screen, so the button stays visible there).

Suggested fix — add a retry control to the error branch:

```html
<div v-else-if="error" class="max-w-5xl mx-auto px-8 py-12">
  <div class="bg-red-950/40 border border-red-900/50 rounded-lg p-6 text-red-300 text-sm flex items-center justify-between gap-4">
    <span>{{ error }}</span>
    <RefreshButton :loading="refreshing" @refresh="reload" />
  </div>
</div>
```

This alone isn't sufficient: none of Assets, Legal, Risks, Suppliers, Systems, Audit, or Inbox's primary loaders ever reset `error.value = null` on entry, so a successful retry would leave the stale error banner showing forever since `v-else-if="error"` stays true. Those loaders need `error.value = null` added at the top of the fetch too (CorrectiveActions/Incidents/Objectives/Programs already do this inside `loadAll()`, so they're covered by the fix above once `fetchAll()` is wired to `reload()`).

---

### 3. [nit] `web/src/views/Risks.vue` line 432

`reload()` does `await loadRisks(); loadAllAdvisories()` — but `loadRisks()` already fires `loadAllAdvisories()` itself (fire-and-forget) as part of its own body (line 982), so every refresh click triggers two advisory fetches instead of one.

Suggested fix — drop the redundant call since `loadRisks()` already covers it:

```js
async function reload() {
  refreshing.value = true
  try {
    await loadRisks()
  } finally {
    refreshing.value = false
  }
}
```

unidoc-ahall:
Thanks for the thorough review @unidoc-alip — all three addressed:

**1. [must-fix] loadAll() skeleton flicker on refresh.** Fixed exactly as suggested. Extracted a `fetchAll()` (the fetch body, no `loading` toggle) in `CorrectiveActions`, `Incidents`, `Objectives` and `Programs`; `loadAll()` keeps the `loading` + `error` handling and calls `fetchAll()`, and `reload()` calls `fetchAll()` directly. Refresh now updates in place — the list and the button stay mounted.

**2. [should-fix] No retry after a failed initial load.** Added `<RefreshButton :loading="refreshing" @refresh="reload" />` to the error branch of all 12 error-screen views. To make the retry actually recover (not leave a stale banner), `reload()` now resets `error.value = null` on entry and catches, so a successful retry clears the error and a failed one re-shows it. (Changes/Tasks unchanged — toast-based, button already stays visible.)

**3. [nit] Risks double advisory fetch.** `reload()` now just `await loadRisks()` — dropped the redundant `loadAllAdvisories()` since `loadRisks()` already fires it.

`just build-web` green. Pushed on `feat/refresh-button`.

unidoc-alip (approved):
looks good to me

PR #170, opened 2026-07-14
fix(auth): style buttons/inputs on all auth screens, not just login

### LINKED ISSUE #169: Forgot/reset password: make the submit a proper button (styling)
The submit/reset action on the forgot-password (reset password) screen isn't styled as a proper button — it looks unfinished. Make it a standard primary button, consistent with the rest of the auth screens (login, signup).

unidoc-ahall:
Closes #169 — the reset/forgot-password button looked unstyled ("ugly").

**Root cause (not just that one button):** the brand-* auth classes — `.brand-btn`, `.brand-text`, `.brand-text-hover`, `.brand-focus`, `.brand-info-box` — were defined in `Login.vue`'s `<style scoped>`. Scoped styles only apply to Login, so **ForgotPassword, Signup, VerifyEmail and VerifyEmailChange** all use those classes but never received the CSS — their primary buttons rendered with white text and no background, and inputs had no brand focus ring.

**Fix:** move the shared brand-* rules to global `src/style.css` (with a `var(--brand-color, #3b82f6)` fallback so they work even when an org hasn't set a branding color), and remove the scoped block from Login. One change fixes all four screens; Login is unchanged visually.

`just build-web` green. No logic/API change.

### DIFF
--- a/web/src/style.css
+++ b/web/src/style.css
@@ -673,3 +673,30 @@
     page-break-after: avoid;
   }
 }
+
+/* ── Brand-colored auth controls ────────────────────────────────────────────
+   Shared by every auth screen (login, signup, forgot/reset password, email
+   verification). Previously scoped to Login.vue, so the OTHER auth screens
+   rendered their buttons/inputs unstyled ("ugly"). Global now; --brand-color
+   falls back to blue-500 when no org branding is configured. */
+.brand-focus:focus {
+  border-color: var(--brand-color, #3b82f6);
+  box-shadow: 0 0 0 1px color-mix(in srgb, var(--brand-color, #3b82f6) 30%, transparent);
+}
+.brand-btn {
+  background-color: var(--brand-color, #3b82f6);
+}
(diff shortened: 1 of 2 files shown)


unidoc-alip (approved):
Closes #169. Root cause confirmed: `.brand-btn`/`.brand-text`/`.brand-text-hover`/`.brand-focus`/`.brand-info-box` were scoped to Login.vue, so ForgotPassword, Signup, VerifyEmail and VerifyEmailChange referenced the same classes but never got the CSS — verified via grep that all five views use these classes. Moving the rules to global `src/style.css` with a `var(--brand-color, #3b82f6)` fallback fixes all four screens at once; the fallback matches App.vue's existing `--brand-color: #3b82f6` root default, so Login renders identically to before. No other view defines conflicting `.brand-*` classes, so going global introduces no collisions. Pure CSS/markup change — no logic, no user input, no security implications. Straightforward, well-scoped fix.

looks good to me

Issue #173, opened 2026-07-14
Asset identifier minted inconsistently: AST- (suggestion-apply) vs ASSET- (HTTP) — standardize on ASSET-


unidoc-ahall:
Assets were minted `ASSET-N` on the HTTP path but `AST-N` on the suggestion-apply path (`suggestions_tx.go` hardcoded it), and `parseID` only stripped `ASSET-`. Standardize on `ASSET-` by routing the suggestion path through the shared `nextIdentifierTx(…, "asset")` helper; `parseID` keeps stripping `AST-` for any legacy rows.

PR #174, opened 2026-07-14
fix(api): entity endpoints accept identifier form + int64 id consistency

### LINKED ISSUE #172: API: entity endpoints accept identifier form (TASK-6), not just numeric id (+ int64 id consistency)
Task, incident, corrective-action, audit-finding and legal endpoints parsed the URL id with raw `strconv.Atoi`, so only a numeric id worked — the identifier form (TASK-6, INC-3, CA-2, FIND-5, LEGAL-4) returned 400, unlike suppliers/assets/risks/systems which accept both via `parseID`.

Fix: extend `parseID` to strip all identifier prefixes `NextIdentifier` mints; route the affected get/update/status/delete handlers through it; standardize those entities' id to int64 to match the rest of the codebase. Verified by `go build`; regression test asserts identifier and numeric forms resolve the same record.

unidoc-ahall:
Closes #172 and #173.

**Identifier URLs + int64.** Task/incident/corrective-action/audit-finding/legal endpoints parsed the `:id` param with raw `strconv.Atoi`, so only a numeric id worked and the identifier form (TASK-6, INC-3, …) returned 400 — asymmetric with suppliers/assets/risks/systems. Now they route through `parseID` (extended to strip every prefix `NextIdentifier` mints), and those five entities' id is standardized to **int64** to match the rest of the codebase so the helper drops in cleanly.

**AST- → ASSET-.** Suggestion-applied assets were minted `AST-N` while HTTP-created ones were `ASSET-N`. The suggestion path now uses the shared `nextIdentifierTx(…, "asset")` helper (→ `ASSET-`); `parseID` still strips `AST-` so any legacy rows keep resolving.

Regression test (`tests/test_identifier_urls.py`) asserts the identifier and numeric forms resolve the same record for tasks/incidents/CAs/legal (GET/PUT/DELETE), and that garbage ids still 400. `go build` + `just test-go` green.

### DIFF
--- a/internal/isms/api/api_audit.go
+++ b/internal/isms/api/api_audit.go
@@ -641,7 +641,7 @@ func (s *Server) handlePaginatedAuditFindings(c echo.Context) error {
 
 func (s *Server) handleGetAuditFinding(c echo.Context) error {
 	orgID := getOrgID(c)
-	id, err := strconv.Atoi(c.Param("id"))
+	id, err := parseID(c.Param("id"))
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "invalid finding id")
 	}
@@ -726,7 +726,7 @@ func (s *Server) handleUpdateAuditFinding(c echo.Context) error {
 		return err
 	}
 	orgID := getOrgID(c)
-	id, err := strconv.Atoi(c.Param("id"))
+	id, err := parseID(c.Param("id"))
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "invalid finding id")
 	}
@@ -791,7 +791,7 @@ func (s *Server) handleUpdateAuditFindingStatus(c echo.Context) error {
 		return err
 	}
 	orgID := getOrgID(c)
-	id, err := strconv.Atoi(c.Param("id"))
+	id, err := parseID(c.Param("id"))
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "invalid finding id")
 	}
@@ -833,7 +833,7 @@ func (s *Server) handleDeleteAuditFinding(c echo.Context) error {
 		return err
 	}
 	orgID := getOrgID(c)
-	id, err := strconv.Atoi(c.Param("id"))
+	id, err := parseID(c.Param("id"))
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "invalid finding id")
 	}
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -2140,7 +2140,7 @@ func (s *Server) handleUpdateTaskStatus(c echo.Context) error {
 	}
 	orgID := getOrgID(c)
 	ctx := c.Request().Context()
-	id, err := parseID(c.Param("id"))
+	id, err := s.resolveTaskID(c.Request().Context(), orgID, c.Param("id"))
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "invalid task id")
 	}
@@ -2184,7 +2184,7 @@ func (s *Server) handleUpdateTaskStatus(c echo.Context) error {
 
 func (s *Server) handleGetTask(c echo.Context) error {
 	orgID := getOrgID(c)
-	id, err := parseID(c.Param("id"))
+	id, err := s.resolveTaskID(c.Request().Context(), orgID, c.Param("id"))
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "invalid task id")
 	}
@@ -2201,7 +2201,7 @@ func (s *Server) handleUpdateTask(c echo.Context) error {
 	}
 	orgID := getOrgID(c)
 	ctx := c.Request().Context()
-	id, err := parseID(c.Param("id"))
+	id, err := s.resolveTaskID(c.Request().Context(), orgID, c.Param("id"))
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "invalid task id")
 	}
@@ -2295,7 +2295,7 @@ func (s *Server) handleDeleteTask(c echo.Context) error {
 	}
 	orgID := getOrgID(c)
 	ctx := c.Request().Context()
-	id, err := parseID(c.Param("id"))
+	id, err := s.resolveTaskID(c.Request().Context(), orgID, c.Param("id"))
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "invalid task id")
 	}
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -2141,8 +2141,10 @@ func (s *Server) handleUpdateTaskStatus(c echo.Context) error {
 	orgID := getOrgID(c)
 	ctx := c.Request().Context()
 	id, err := s.resolveTaskID(c.Request().Context(), orgID, c.Param("id"))
-	if err != nil {
+	if errors.Is(err, errInvalidID) {
 		return echo.NewHTTPError(http.StatusBadRequest, "invalid task id")
+	} else if err != nil {
+		return echo.NewHTTPError(http.StatusNotFound, "task not found")
 	}
 	var req struct {
 		Status string `json:"status"`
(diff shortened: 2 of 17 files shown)


unidoc-ahall:
Heads-up on a real bug the identifier-URL tests caught (thanks to the regression test): `parseID` stripped the prefix and treated the numeric **suffix** as the primary key. That only holds when the per-org `identifier_sequences` value happens to equal the table `id`. For **task/incident/corrective-action/legal** (minted `PREFIX-<seq>`) they diverge, so `GET /tasks/TASK-8` was looking up `id=8` → 404.

Fixed by resolving the identifier form via `GetXByIdentifier` lookups (`resolve*ID` helpers) instead of stripping. **Findings keep `parseID`** — they're `FIND-<id>` (the suffix *is* the id), so that path is correct.

Note for a separate ticket: the pre-existing `supplier`/`asset`/`risk`/`system` handlers use the same `parseID`-then-GetByID pattern, so they have the same latent bug when `seq != id` — just never surfaced because their identifier-URL path isn't exercised by tests or the SPA (which resolves from the loaded list). Out of scope for #174.

`go build` + `just test-go` green.

unidoc-alip (changes requested):
Solid fix, and the follow-up commit shows good judgment — the author's own regression test caught that `parseID` treats the numeric suffix as the PK, which only holds when the per-org identifier sequence happens to equal the id; task/incident/CA/legal now correctly resolve via `GetXByIdentifier` lookups instead. `go build` is clean and the new test file covers GET/PUT/DELETE for both id forms. Two non-blocking gaps below: a status-code inconsistency in the new `resolve*ID` helpers, and a pre-existing instance of the same bug class left un-flagged in the suggestion-apply path.

---

### 1. [should-fix] `internal/isms/api/server.go` line 319

`resolveTaskID` (and its incident/CA/legal siblings) return the raw `GetXByIdentifier` error for a syntactically valid but nonexistent identifier (e.g. `GET /tasks/TASK-9999`), and every call site maps that straight to `400 invalid task id`. Compare with the numeric path: a nonexistent numeric id sails through this check and correctly 404s at the subsequent `GetTask` call (api_collab.go:2155-2157). So `GET /tasks/999999` → 404 but `GET /tasks/TASK-9999` → 400 for what's semantically the same "not found" case.

Suggested fix — check the identifier's own prefix before hitting the DB, so a malformed param (no digits, no recognized prefix) still 400s early, but a well-formed-but-missing identifier falls through to the existing 404 at the `Get*` call:

```go
func (s *Server) resolveTaskID(ctx context.Context, orgID int, param string) (int64, error) {
    if n, err := strconv.ParseInt(param, 10, 64); err == nil {
        return n, nil
    }
    if !strings.HasPrefix(param, "TASK-") {
        return 0, fmt.Errorf("%w: %q", errInvalidID, param)
    }
    t, err := s.db.GetTaskByIdentifier(ctx, orgID, param)
    if err != nil {
        return 0, err // not-found — let the caller 404, not 400
    }
    return t.ID, nil
}
```

Then at each call site, only 400 when the error is `errInvalidID`; anything else (a real `GetTaskByIdentifier` miss) should 404:

```go
id, err := s.resolveTaskID(c.Request().Context(), orgID, c.Param("id"))
if errors.Is(err, errInvalidID) {
    return echo.NewHTTPError(http.StatusBadRequest, "invalid task id")
} else if err != nil {
    return echo.NewHTTPError(http.StatusNotFound, "task not found")
}
```

Same shape applies to `resolveIncidentID` ("INC-"), `resolveCorrectiveActionID` ("CA-"), and `resolveLegalID` ("LEGAL-"). `errInvalidID` can be a single package-level `errors.New("invalid id")` sentinel declared near `parseID`.

---

### 2. [should-fix] `internal/isms/api/api_suggestions.go` line 791

`applyIncidentUpdate`/`applyCorrActiveUpdate`/`applyTaskUpdate` still resolve `sg.EntityID` via `parseEntityID`, which strips the identifier prefix and treats the numeric suffix directly as the primary key passed to `GetIncident`/`GetCorrectiveAction`/`GetTask`. That's the exact bug class this PR's second commit just fixed for URL params (identifier suffix is a per-org `identifier_sequences` value, not the row id, for task/incident/CA/legal) — just on the suggestion-apply path instead of the HTTP `:id` param path. Confirmed `sg.EntityID` is always the identifier form here (e.g. `INC-3`), not a bare numeric id — the Vue side passes `selectedIncident.identifier` / `selectedTask.identifier` / `selectedCA.identifier` as `entityId` into `SuggestionPanel`/`CommentsPanel` (Incidents.vue:475, Tasks.vue:342, CorrectiveActions.vue:356), which becomes `entity_id` on the suggestion.

The codebase already has the correct pattern two callers away — `applyRiskUpdate` / `applySupplierUpdate` / `applyLegalReading` resolve via `GetRiskByIdentifier` / `GetSupplierByIdentifier` / `GetLegalRequirementByIdentifier` directly. Suggested fix — mirror that pattern (all three `GetXByIdentifier` helpers already exist, including the `GetTaskByIdentifier` this PR just added):

```go
// applyIncidentUpdate
inc, err := s.db.GetIncidentByIdentifier(ctx, orgID, sg.EntityID)
if err != nil {
    return "", fmt.Errorf("incident %s not found: %w", sg.EntityID, err)
}
// drop the `parseEntityID(sg.EntityID)` call above it — incID is no longer needed

// applyCorrActiveUpdate
ca, err := s.db.GetCorrectiveActionByIdentifier(ctx, orgID, sg.EntityID)
if err != nil {
    return "", fmt.Errorf("corrective action %s not found: %w", sg.EntityID, err)
}

// applyTaskUpdate
t, err := s.db.GetTaskByIdentifier(ctx, orgID, sg.EntityID)
if err != nil {
    return "", fmt.Errorf("task %s not found: %w", sg.EntityID, err)
}
```

---

### 3. [nit] `internal/isms/api/server.go` line 306

This comment states findings and SUPPLIER-/ASSET-/RISK-/SYSTEM- identifiers "work" with `parseID` because the suffix IS the primary key — true for findings (`FIND-<id>` is built directly from the row id, per `SoftDeleteAuditFinding`), but for the other four `nextIdentifierTx` mints them from the same per-org `identifier_sequences` counter used for task/incident/CA/legal, so they have the same theoretical seq-vs-id divergence risk, just not yet observed/tested.

Suggested fix — soften the comment so it doesn't read as a structural guarantee:

```go
// parseID accepts a bare numeric id, or an identifier whose numeric SUFFIX is used
// directly as the primary key: findings (FIND-<id>, built from the row id itself — see
// SoftDeleteAuditFinding) and, so far incidentally rather than by design, the
// pre-existing SUPPLIER-/ASSET-/RISK-/SYSTEM- handlers (same identifier_sequences
// mechanism as TASK-/INC-/CA-/LEGAL-, just not yet observed to diverge — see #174 PR
// discussion). It does NOT work for seq-based identifiers where the suffix is known to
// diverge from the id (TASK-/INC-/CA-/LEGAL-) — those resolve via the resolve*ID lookups
// below.
```

unidoc-ahall:
Thanks @unidoc-alip — all three addressed, and a note on CI:

**1. [should-fix] 404 vs 400 in resolve*ID.** Added an `errInvalidID` sentinel + a prefix check: a malformed param (no digits, wrong/no prefix) returns `errInvalidID` → the call sites 400; a well-formed-but-missing identifier returns the `GetXByIdentifier` lookup error → 404, matching the numeric path.

**2. [should-fix] suggestion-apply seq bug.** `applyIncidentUpdate`/`applyCorrActiveUpdate`/`applyTaskUpdate` now resolve via `GetIncidentByIdentifier`/`GetCorrectiveActionByIdentifier`/`GetTaskByIdentifier` instead of `parseEntityID`-suffix. Good catch — it's actually a bit broader: `change_request`/`objective`/`system`/`asset` apply paths have the same pattern (system/asset have the helper; change/objective need one added). Filed as a follow-up (https://github.com/unidoc/isms/issues/177) rather than widen this PR.

**3. [nit] parseID comment.** Softened per your wording.

**On CI:** the red integration job is a flake, not this PR — the one failure is `test_e2e_browser.py::TestTaskEditStatusPersists`. That path is `PUT /tasks/{numeric-id}` → `handleUpdateTask`, and for a numeric id `resolveTaskID` is behaviourally identical to the old `strconv.Atoi` (same value, same `GetTask`/`UpdateTask`). master passes this test and this PR is backend-only, so the numeric task-update path is unchanged. Re-running the job.

`go build` + `just test-go` green.

unidoc-alip (dismissed):
looks good to me

unidoc-ahall:
CI fix — and it was my regression from the review commit, not a flake (thanks for pushing back on that).

My earlier commit routed suggestion-apply through `GetXByIdentifier(sg.EntityID)`. But `sg.EntityID` isn't always the identifier form: `test_corrective_writepath` / `test_incident_writepath` post `entity_id: str(ca_id)` — the **numeric** id — so `GetCorrectiveActionByIdentifier("7")` looked up `WHERE identifier='7'` → no rows → the `corrective action 7 not found` 500. (The SPA sends the identifier form, which is why the task suggestion test passed — different form.)

Fixed by routing through the `resolve*ID` helpers, which accept **either** a numeric id or the `PREFIX-<seq>` identifier (numeric → straight to `Get*`; identifier → `GetXByIdentifier`) — the same both-forms resolution the URL handlers use. Covers both the numeric-entity_id tests and the identifier-form suggestions.

Note: the follow-up issue I filed for change/objective/system/asset should use this same both-forms `resolve*ID` pattern, not `GetXByIdentifier` alone. `go build` + `just test-go` green.

PR #179, opened 2026-07-15
feat(tasks): per-task private visibility + org default

### LINKED ISSUE #178: Task visibility: per-task private flag + org default
## Why
Today every member of an org sees **every** task — `handleListTasks` → `PaginatedTasks` filter only by `organization_id`, and the `assignee` param is an optional UI convenience, not a boundary. A design partner needs task **privacy**: managers create tasks for people, and contributors/readers should not see tasks that aren't theirs (HR, security, disciplinary work). An ISMS platform should be privacy-respecting about who-sees-what.

(Task *create* is already manager/admin-only via `requireRole`. The `tasks` table already has `assignee_id` NOT NULL + `created_by_user_id`, so no new columns are needed for the rule itself — only the flag.)

## Model
- **Column** `tasks.private BOOLEAN NOT NULL DEFAULT false` (in the v0.7.1 migration). **Public by default → zero behaviour change for existing deployments; existing tasks stay public.**
- **Visibility predicate (one rule, app-layer — no RLS):**
  `NOT private OR assignee_id = :me OR created_by_user_id = :me OR role >= manager`
  applied via a shared helper in the task read paths in `db/tasks.go`. A hidden `GET /tasks/:id` returns **404, not 403** (don't reveal a private task exists).
- **Org setting** `task_default_private` (default `false`): decides the value when a create request omits `private`. An org that wants privacy-by-default flips this one setting — no product-wide change.
- **UI/API:** create form shows a **Public / Private** choice + a manager can toggle it after creation; the Python/API client gets an optional `private` field — omit it and the org default applies (backward-compatible).
- Managers/admins always see all (oversight); the assignee always sees their own; the creator always sees what they created.

### API tokens & visibility (no new grant — anchored to the owning user's role)
The predicate keys off the **resolved role of the token's user** (`GetUserRole(org, tok.UserID)`), never off a token flag. Consequences, all safe by construction:
- A **reader**'s token sees only that reader's own private tasks + public — it can **never** see others' private tasks. The same holds for a **contributor/agent** token (agents can't snoop private tasks). No new attribute is added to grant private visibility — an independent "see private" token flag would be a privilege-escalation hole (it could land on a low-role token) and is **explicitly rejected**.
- Token `permissions` (`read` / `write` / `read-write`) only **narrows** (AND), never widens: `permissions=read` removes mutation but does **not** expand visibility.
- **To get a token that reads ALL private tasks** (backup/sync/integration): issue it against a **manager/admin service user** and set `permissions=read`. See-all comes from the role; read-only comes from the token permission. **Zero new code** — both mechanisms already exist.

## Scope checklist (fan-out — everything that returns/counts tasks must honour the predicate; this is where it leaks if missed)
- [ ] `db/tasks.go` read fns (`ListTasks`, `GetTask`, `GetTaskByIdentifier`, `OverdueTasks`, `ListTasksWhere`, `PaginatedTasks`) + shared predicate helper + unit test
- [ ] Task stats / counts (`TaskStats`, dashboard rollups)
- [ ] Calendar view
- [ ] Overdue rollups
- [ ] Notification / mention / activity feed (don't surface a task the recipient can't see)
- [ ] **Entity-references in comments** — a private task's title must not leak via a comment ref (interaction with the comment-references work)
- [ ] Suggestions raised on a task
- [ ] Create form Public/Private control + manager edit toggle + a "Private" badge in lists
- [ ] Org setting `task_default_private` + settings UI
- [ ] Python/API client optional `private` field
- [ ] Regression tests: private hidden from a non-owner contributor; visible to assignee / creator / manager; public unchanged; hidden `GET` → 404
- [ ] Token regression tests: reader/contributor token can't see others' private tasks; manager/admin **read-only** token (`permissions=read`) sees all private but gets 403 on write; `permissions=read` never widens visibility

## Decisions (settled)
- **Product default = PUBLIC everywhere** (no behaviour change); per-org opt-in via `task_default_private`. Existing tasks remain public.
- Hidden task `GET` → **404**.

## Non-goals
RLS (app-layer enforcement only — task reads don't run in RLS context today); per-field ACLs; sharing with named individuals (visibility is role + assignee + creator only); a token attribute that grants private visibility independently of the owning user's role (rejected — see "API tokens & visibility"; use a manager/admin service user + read-only token instead).

## Known limitation (follow-up)
Universal search uses an org-wide shared/cached index that can't be filtered per viewer at build time, so **private tasks are excluded from universal search** (public tasks only) to avoid leaking titles. Making a user's own private tasks searchable needs query-time visibility filtering — tracked separately.

unidoc-ahall:
Closes #178

## What
Tasks can be marked **private**. A private task is visible only to its **assignee, its creator, and managers/admins**; everyone else (readers, contributors, agents) can't see it. **Public by default** — existing tasks and existing deployments are unchanged. An org opts into privacy-by-default by flipping the **`task_default_private`** setting (Admin → Settings).

## Why
Today every member of an org sees **every** task (`handleListTasks` → `PaginatedTasks` filter only by `organization_id`; the `assignee` param is a UI convenience, not a boundary). A design partner needs task privacy — managers create tasks for people, and contributors/readers shouldn't see tasks that aren't theirs (HR, security, disciplinary work).

## How
- **Migration (v0.7.1):** `tasks.private` column + a `task_default_private` settings-catalog row (the catalog-driven Admin UI renders it as a boolean toggle automatically).
- **One predicate, app-layer, no RLS:** `db.TaskViewer{Email, CanSeeAll}` → `NOT private OR assignee_id=me OR created_by_user_id=me`, applied in the task read paths (`ListTasks`, `PaginatedTasks`, `ListTasksWhere`). Managers/admins (`CanSeeAll`) leave queries byte-for-byte unchanged. (RLS is only tenant-isolation here and reads don't run in its context, so this stays app-layer.)
- **Single by-id fetches** enforce `canViewTask` in the handler → **404, not 403** (don't reveal a private task exists). Entity-reference title resolution hides a private task's title from non-viewers.
- **Token safety:** visibility keys off the **token user's role**, never a token flag — a reader/contributor/agent token can't see others' private tasks. There is deliberately **no "see-private" token attribute** (that would be privilege escalation); to read all private tasks use a **manager/admin service user + read-only token**.
- **UI/API:** create form gets a **Public / Private** control seeded from the org default (exposed via `/config`); create & update accept an optional `private` field (omit → org default). A "Private" badge shows on private tasks.

## Testing
- `db.TaskViewer.visibilityClause` (CanSeeAll → no predicate; restricted → predicate + email bound once) and `api.canViewTask` (public visible to all; private visible to manager/admin/assignee/creator; hidden from unrelated reader/contributor incl. agents) — all green.
- `just fmt`, `just build-go`, `just test-go` (full suite), `just build-web` all pass.

## Known limitation (follow-up)
Universal search uses an org-wide shared/cached index that can't be filtered per viewer at build time, so **private tasks are excluded from universal search** (public only) to avoid leaking titles. Making a user's own private tasks searchable needs query-time filtering — tracked separately.

No breaking change for existing clients (public default; token-only flows unchanged).

### DIFF
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -2076,7 +2076,7 @@ func (s *Server) handleListTasks(c echo.Context) error {
 
 func (s *Server) handleTaskStats(c echo.Context) error {
 	orgID := getOrgID(c)
-	stats, err := s.db.TaskStats(c.Request().Context(), orgID)
+	stats, err := s.db.TaskStats(c.Request().Context(), orgID, taskViewer(c))
 	if err != nil {
 		return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
 	}
@@ -2143,7 +2143,11 @@ func (s *Server) handleCreateTask(c echo.Context) error {
 
 	s.createReferencesForEntity(ctx, orgID, "task", t.Identifier, t.CreatedBy, req.References)
 
-	s.searchUpsert(orgID, "task", t.Identifier, t.Title, t.Identifier+" "+t.Title+" "+t.Description)
+	// A private task must not surface its title through the shared search index or
+	// the org-wide activity feed / chat webhooks (#178 review).
+	if !t.Private {
+		s.searchUpsert(orgID, "task", t.Identifier, t.Title, t.Identifier+" "+t.Title+" "+t.Description)
+	}
 
 	s.logChange(ctx, orgID, &db.ChangelogEntry{
 		EntityType: "task",
@@ -2151,10 +2155,14 @@ func (s *Server) handleCreateTask(c echo.Context) error {
 		Action:     "create",
 		ChangedBy:  t.CreatedBy,
 	})
+	createDetail := fmt.Sprintf("Task: %s assigned to %s", t.Title, t.Assignee)
+	if t.Private {
+		createDetail = "Private task created"
+	}
 	s.logAndNotify(ctx, orgID, &db.Activity{
 		Actor:  t.CreatedBy,
 		Action: "task_created",
-		Detail: fmt.Sprintf("Task: %s assigned to %s", t.Title, t.Assignee),
+		Detail: createDetail,
 	})
 
 	// Email assignee if set
@@ -2317,12 +2325,22 @@ func (s *Server) handleUpdateTask(c echo.Context) error {
 	diffs := db.DiffFields("task", int64(id), user, "", old.ToChangeMap(), t.ToChangeMap())
 	s.logChanges(ctx, orgID, diffs)
 
-	s.searchUpsert(orgID, "task", old.Identifier, t.Title, old.Identifier+" "+t.Title+" "+t.Description)
+	// Keep the search index and activity feed clear of a private task's title;
+	// a public→private flip must also remove any existing index entry (#178 review).
+	if t.Private {
+		s.searchRemove(orgID, "task", old.Identifier)
+	} else {
+		s.searchUpsert(orgID, "task", old.Identifier, t.Title, old.Identifier+" "+t.Title+" "+t.Description)
+	}
 
+	updateDetail := fmt.Sprintf("Task #%d: %s", id, t.Title)
+	if t.Private {
+		updateDetail = fmt.Sprintf("Task #%d updated (private)", id)
+	}
 	s.logAndNotify(ctx, orgID, &db.Activity{
 		Actor:  user,
 		Action: "task_updated",
-		Detail: fmt.Sprintf("Task #%d: %s", id, t.Title),
+		Detail: updateDetail,
 	})
 
 	// Reload to get computed fields
@@ -2362,10 +2380,14 @@ func (s *Server) handleDeleteTask(c echo.Context) error {
 		Action:     "delete",
 		ChangedBy:  user,
 	})
+	deleteDetail := fmt.Sprintf("%s: %s", task.Identifier, task.Title)
+	if task.Private {
+		deleteDetail = fmt.Sprintf("%s deleted (private)", task.Identifier)
+	}
 	s.logAndNotify(ctx, orgID, &db.Activity{
 		Actor:  user,
 		Action: "task_deleted",
-		Detail: fmt.Sprintf("%s: %s", task.Identifier, task.Title),
+		Detail: deleteDetail,
 	})
 
 	s.searchRemove(orgID, "task", task.Identifier)
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -195,6 +195,31 @@ func getUserEmail(c echo.Context) string {
 	return ""
 }
 
+// taskViewer builds the task-visibility scope for the current caller: managers
+// and admins see every task; everyone else sees only public tasks plus their own
+// (assigned to them or created by them).
+func taskViewer(c echo.Context) db.TaskViewer {
+	role, _ := c.Get("user_role").(string)
+	return db.TaskViewer{
+		Email:     getUserEmail(c),
+		CanSeeAll: role == "manager" || role == "admin",
+	}
+}
+
+// canViewTask mirrors db.TaskViewer for single by-id fetches (which aren't
+// SQL-filtered): reports whether the caller may see this task under the privacy
+// rule. A hidden task is treated as not-found (404), never 403 — don't reveal it.
+func canViewTask(c echo.Context, t *db.Task) bool {
+	if !t.Private {
+		return true
+	}
+	if role, _ := c.Get("user_role").(string); role == "manager" || role == "admin" {
+		return true
+	}
+	email := getUserEmail(c)
+	return t.Assignee == email || t.CreatedBy == email
+}
+
 // getOrgID returns the organization ID set by AuthMiddleware, or 0 if not set.
 func getOrgID(c echo.Context) int {
 	if id, ok := c.Get("org_id").(int); ok {
@@ -2031,7 +2056,7 @@ func (s *Server) handleListTasks(c echo.Context) error {
 		TaskType: c.QueryParam("task_type"),
 		Assignee: c.QueryParam("assignee"),
 	}
-	items, total, err := s.db.PaginatedTasks(c.Request().Context(), orgID, params)
+	items, total, err := s.db.PaginatedTasks(c.Request().Context(), orgID, taskViewer(c), params)
 	if err != nil {
 		return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
 	}
@@ -2105,6 +2130,13 @@ func (s *Server) handleCreateTask(c echo.Context) error {
 		return err
 	}
 	ctx := c.Request().Context()
+	// Visibility: an explicit flag wins; otherwise fall back to the org default
+	// (public unless the org has opted into task_default_private).
+	if req.Private != nil {
+		t.Private = *req.Private
+	} else if v, _ := s.db.GetOrgSetting(ctx, orgID, "task_default_private"); v == "true" {
+		t.Private = true
+	}
 	if err := s.db.CreateTask(ctx, orgID, &t); err != nil {
 		return pgxHTTPError(err)
 	}
@@ -2196,6 +2228,9 @@ func (s *Server) handleGetTask(c echo.Context) error {
 	if err != nil {
 		return echo.NewHTTPError(http.StatusNotFound, "task not found")
 	}
+	if !canViewTask(c, task) {
+		return echo.NewHTTPError(http.StatusNotFound, "task not found")
+	}
 	return c.JSON(http.StatusOK, task)
 }
 
@@ -2269,6 +2304,9 @@ func (s *Server) handleUpdateTask(c echo.Context) error {
 	}
 	if req.Notes != nil {
 		t.Notes = *req.Notes
+	}
+	if req.Private != nil {
+		t.Private = *req.Private
 	}
 
 	if err := s.db.UpdateTask(ctx, orgID, &t); err != nil {
@@ -2347,6 +2385,7 @@ type taskCreateRequest struct {
 	DueDate        *db.Epoch        `json:"due_date"`
 	RecurrenceDays *int             `json:"recurrence_days"`
 	Notes          string           `json:"notes"`
+	Private        *bool            `json:"private"`
 	References     []ReferenceInput `json:"references"`
 }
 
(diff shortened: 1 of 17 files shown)


unidoc-alip (changes requested):
The SQL-filtered list paths (`ListTasks`/`PaginatedTasks`/`ListTasksWhere`), the by-id 404-guard (`canViewTask`), and the entity-reference title guard are all correctly implemented and tested — nice work on those. However, this review found six other surfaces that read or write the `tasks` table and were not updated to honor the new privacy predicate, each leaking a private task's title, existence, or aggregate presence to users who shouldn't see it. Three of these are exactly the items issue #178's own scope checklist flagged as required ("Task stats / counts", "Notification / mention / activity feed", "Suggestions raised on a task") and left unchecked; a fourth is a live search-index path that bypasses the public-only safeguard this PR itself added to the index-rebuild path. All findings below were verified against current `master` source, not just this diff, with file:line citations and a concrete fix per finding.

---

### 1. [must-fix] `internal/isms/db/overdue.go` line 144

`GetOverdueSummary`'s task query (used by `GET /overdue` → `handleOverdueSummary`, which has no role restriction) selects `t.title` and the assignee's email for every overdue task in the org, with no `TaskViewer`/`private` filter:

```go
rows5, err := d.pool.Query(ctx, `
    SELECT t.id::text, t.title,
        COALESCE((SELECT email FROM users WHERE id = t.assignee_id), ''),
        t.due_date, t.priority
    FROM tasks t
    WHERE t.organization_id = $1
        AND t.status NOT IN ('done', 'cancelled')
        AND t.due_date IS NOT NULL
        AND t.due_date < $2
        AND t.deleted_at IS NULL
`, orgID, now)
```

An overdue private task's title and assignee are returned verbatim to any authenticated org member via the overdue widget/summary — the most direct leak in this review.

**Suggested fix:** thread a `TaskViewer` through:

```go
func (d *DB) GetOverdueSummary(ctx context.Context, orgID int, viewer TaskViewer) (*OverdueSummary, error) {
    ...
    vis, vargs := viewer.visibilityClause(3)
    rows5, err := d.pool.Query(ctx, `
        SELECT t.id::text, t.title,
            COALESCE((SELECT email FROM users WHERE id = t.assignee_id), ''),
            t.due_date, t.priority
        FROM tasks t
        WHERE t.organization_id = $1
            AND t.status NOT IN ('done', 'cancelled')
            AND t.due_date IS NOT NULL
            AND t.due_date < $2
            AND t.deleted_at IS NULL`+vis, append([]interface{}{orgID, now}, vargs...)...)
```

Update call sites: `handleOverdueSummary` (api/server.go:3050) → `s.db.GetOverdueSummary(ctx, orgID, taskViewer(c))`; `CreateOverdueReviewTasks` (db/overdue.go:295) → `TaskViewer{CanSeeAll: true}` (it needs full visibility for correct dedup). Add a regression test asserting a private task assigned to A is absent from the summary for viewer B but present for A and for `CanSeeAll`.

---

### 2. [must-fix] `internal/isms/api/api_collab.go` line 2155

`handleCreateTask`, `handleUpdateTask`, and `handleDeleteTask` all call `logAndNotify` unconditionally with the task title, regardless of `Private`:

```go
// handleCreateTask (line 2155)
s.logAndNotify(ctx, orgID, &db.Activity{
    Actor:  t.CreatedBy,
    Action: "task_created",
    Detail: fmt.Sprintf("Task: %s assigned to %s", t.Title, t.Assignee),
})
// handleUpdateTask (line ~2321): Detail: fmt.Sprintf("Task #%d: %s", id, t.Title)
// handleDeleteTask (line ~2364): Detail: fmt.Sprintf("%s: %s", task.Identifier, task.Title)
```

`logAndNotify` writes to the shared `activity` table (read back by `RecentActivity`, which has no visibility filter, served by `GET /activity` with no role restriction) and broadcasts `Detail` to the org's Slack webhook / Matrix room if configured. A private task's title and assignee is immediately visible in every reader/contributor's activity feed and posted to the whole org's chat channel on create/update/delete.

**Suggested fix:** redact the title/assignee for private tasks at all three call sites, e.g.:

```go
detail := fmt.Sprintf("Task: %s assigned to %s", t.Title, t.Assignee)
if t.Private {
    detail = "Task created (private)"
}
s.logAndNotify(ctx, orgID, &db.Activity{Actor: t.CreatedBy, Action: "task_created", Detail: detail})
```

Apply the equivalent guard in `handleUpdateTask` (on `t.Private`) and `handleDeleteTask` (on `task.Private`). This fixes both the activity-feed leak and the Slack/Matrix leak at the source. Add a regression test asserting `Activity.Detail` never contains the title when the task is private.

---

### 3. [must-fix] `internal/isms/api/api_suggestions.go` line 197

`handleListEntitySuggestions` / `handleGetEntitySuggestion` return suggestion records — including the suggestion's own `Title` field — with no check against the target task's `private` flag or the requester's relationship to it:

```go
suggestions, err := s.db.ListSuggestions(c.Request().Context(), orgID, filters)
```

Since suggestion creation is open to any user, a pending suggestion against a private task is visible to any org member calling plain `GET /suggestions` — no `entity_id` filter or prior knowledge of the private task's identifier needed. This is issue #178's unchecked "Suggestions raised on a task."

**Suggested fix:** in `db/suggestions.go`'s `ListSuggestions`, join to `tasks` and apply the same visibility predicate when `entity_type = 'task'` and the caller isn't `CanSeeAll` — or, if the join is awkward against the string `entity_id`, post-filter in `handleListEntitySuggestions`/`handleGetEntitySuggestion`: for each `EntityType == "task"` suggestion, resolve via `GetTaskByIdentifier` and drop/404 it using the same rule as `canViewTask`. Cover both the list and single-get handlers. Add a regression test: a suggestion on a private task is visible to the assignee/creator/manager and absent for an unrelated reader/contributor.

---

### 4. [must-fix] `internal/isms/api/api_search.go` line 374

`searchUpsert` writes directly into the live, queryable search index and is called unconditionally from `handleCreateTask` and `handleUpdateTask` with the task's real title, regardless of `Private`:

```go
// handleCreateTask (api_collab.go:2149)
s.searchUpsert(orgID, "task", t.Identifier, t.Title, t.Identifier+" "+t.Title+" "+t.Description)
// handleUpdateTask (api_collab.go:2320)
s.searchUpsert(orgID, "task", old.Identifier, t.Title, old.Identifier+" "+t.Title+" "+t.Description)
```

This bypasses the exact safeguard this PR added in `buildSearchIndex` (the public-only filter on full-index rebuilds) — the incremental upsert path has no such check, so a private task's title becomes universally searchable via `GET /search?q=...` the moment it's created or edited, contradicting the PR description's own claim that private tasks are excluded from universal search.

**Suggested fix:**

```go
if !t.Private {
    s.searchUpsert(orgID, "task", t.Identifier, t.Title, t.Identifier+" "+t.Title+" "+t.Description)
} else {
    s.searchRemove(orgID, "task", t.Identifier) // covers public→private flips on update
}
```

Apply at both create and update call sites (`handleDeleteTask`'s unconditional `searchRemove` is already correct — removal is always safe). Add a test asserting a private task never appears in the search index, and that flipping an existing public task to private removes its entry.

---

### 5. [should-fix] `internal/isms/api/api_collab.go` line 2079

`handleTaskStats` calls `s.db.TaskStats(ctx, orgID)` with no `TaskViewer`, and `TaskStats` counts by status/priority across the whole org with no `private` predicate. `GET /tasks/stats` has no role gate, so any reader/contributor/agent gets aggregate counts that include private tasks — no title leak, but it does leak the existence and volume/priority mix of private tasks, matching issue #178's unchecked "Task stats / counts" item.

**Suggested fix:**

```go
func (d *DB) TaskStats(ctx context.Context, orgID int, viewer TaskViewer) (*TaskStats, error) {
    vis, vargs := viewer.visibilityClause(2)
    err := d.pool.QueryRow(ctx, `
        SELECT count(*), ...
        FROM tasks t
        WHERE t.organization_id = $1 AND t.deleted_at IS NULL`+vis,
        append([]interface{}{orgID}, vargs...)...,
    ).Scan(...)
```

Update `handleTaskStats` to call `s.db.TaskStats(ctx, orgID, taskViewer(c))`. Add a unit test: a manager/admin viewer's total includes a private task; a reader/contributor viewer's counts exclude it.

---

### 6. [nit] `internal/isms/db/tasks.go` line 246

`ListTasksWhere`'s visibility clause is hard-coded to placeholder `$2` regardless of how many placeholders the caller's `where` string uses. Not currently exploitable — both current call sites (`OverdueTasks`, and the manual dedup query in `CreateOverdueReviewTasks`) pass a literal `where` with no `$N` placeholders — but it's a latent footgun for a future parameterized caller.

**Suggested fix:** either take the next-free placeholder index as an explicit parameter, or (lower effort, proportionate to current risk) add a doc comment on `ListTasksWhere` stating that `where` must not contain positional placeholders since `visibilityClause` always claims `$2`.

unidoc-ahall:
Thanks for the thorough review — all six are real and now fixed in the latest commit. Point by point:

**1. `GetOverdueSummary` (overdue.go) — [must-fix] ✅** Threaded a `TaskViewer`; the overdue task query now appends `visibilityClause($3)`. `handleOverdueSummary` passes `taskViewer(c)`; `CreateOverdueReviewTasks` passes `TaskViewer{CanSeeAll: true}` (system dedup needs full visibility).

**2. `logAndNotify` title leak (api_collab.go) — [must-fix] ✅** `handleCreateTask`/`handleUpdateTask`/`handleDeleteTask` now redact the title/assignee in `Activity.Detail` when the task is private (`"Private task created"`, `"Task #N updated (private)"`, `"<ID> deleted (private)"`). Fixes the activity feed and the Slack/Matrix webhook at the source.

**3. Suggestions on a private task (api_suggestions.go) — [must-fix] ✅** New `taskSuggestionHidden` helper: `handleListEntitySuggestions` drops task-type suggestions on a private task the caller can't see, and `handleGetEntitySuggestion` returns 404. It resolves the task by numeric id or identifier and reuses `canViewTask`; short-circuits for `CanSeeAll` (no extra DB hits for managers).

**4. `searchUpsert` incremental path (api_collab.go) — [must-fix] ✅** Create only upserts when `!Private`; update removes the entry when private (covers a public→private flip) and upserts otherwise. Delete's unconditional `searchRemove` was already correct.

**5. `TaskStats` (tasks.go) — [should-fix] ✅** Takes a `TaskViewer` and appends the visibility clause (aliased the table to `t`); `handleTaskStats` passes `taskViewer(c)`, so a non-manager's counts exclude private tasks.

**6. `ListTasksWhere` `$2` hard-coding — [nit] ✅** Added a doc comment stating `where` must not contain positional placeholders since `visibilityClause` always claims `$2` (proportionate to current risk — both callers pass literal predicates).

Tests: `tests/test_task_visibility_surfaces.py` covers all five leak surfaces (search / activity / overdue / suggestions / stats), viewer-scoped where applicable. `just fmt`, `just build-go`, `just test-go` (full suite) pass.

unidoc-alip (dismissed):
looks good to me

PR #180, opened 2026-07-15
fix(suggestions): resolve seq-based identifiers on apply (#177)

### LINKED ISSUE #177: Suggestion-apply: resolve seq-based identifiers via lookup (change/objective/system/asset)
Follow-up from #174 review. The suggestion-apply path resolves `sg.EntityID` (the identifier form) via `parseEntityID`, which treats the numeric suffix as the primary key. #174 fixed this for incident/corrective-action/task (and the URL handlers). The same pattern remains in the other seq-based apply functions:

- `applyChangeRequestUpdate` → `GetChangeRequest(suffix)`
- `applyObjectiveUpdate` → `GetObjective(suffix)`
- `applySystemUpdate` → `GetSystem(suffix)`
- `applyAssetUpdate` → `GetAsset(suffix)`

`system`/`asset` already have `GetXByIdentifier`; `change_request`/`objective` need one added. Findings (`FIND-<id>`) are correct as-is. Low observed impact (needs seq!=id in the org), but it's the same latent bug.

unidoc-ahall:
Closes #177

## What
Suggestion-**apply** now resolves the target entity by lookup for **change / objective / system / asset**, so an update suggestion addressing an entity by its per-org **identifier form** lands on the right row.

## Why
`applyChange/Objective/System/AssetUpdate` did `parseEntityID(sg.EntityID)` (strip prefix → int) then `GetX(thatInt)`. The identifier suffix is a **per-org sequence, not the primary key** — on a shared/persistent stack they differ, so an identifier-form suggestion (`CR-n`, `SYSTEM-n`, `ASSET-n`, objective `display_id`) silently updated the **wrong entity**, or errored "not found". This is the same class of bug #174 fixed for task / incident / corrective_action.

## How
- New `resolveChangeID` / `resolveObjectiveID` / `resolveSystemID` / `resolveAssetID`: numeric → id; otherwise look up (`GetChangeRequestByIdentifier` [new] / `GetObjectiveByDisplayID` / `GetSystemByIdentifier` / `GetAssetByIdentifier`). They deliberately skip a fixed-prefix check — asset ids exist as `ASSET-`/`AST-`, systems as `SYSTEM-`, and objective identifiers are program-key display IDs (e.g. `ISMS-1`), so a hardcoded prefix would be wrong.
- The four apply handlers resolve via those helpers instead of `parseEntityID`.

## Testing
- `tests/test_suggestion_apply_identifier.py`: for each of the four types, create it, apply an update addressing it **by its identifier form**, assert the correct row changed (`force=true` to bypass stale-detection).
- `just fmt`, `just build-go`, `just test-go` pass. The Python test runs against the stack (integration).

## Noted, not fixed here
The stale-detection path (`parseEntityID` at the top of the apply/get flow) has the same mis-resolution for these types, but it only affects staleness-warning accuracy, not which row is mutated — separate follow-up if we want it.

### DIFF
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -195,6 +195,31 @@ func getUserEmail(c echo.Context) string {
 	return ""
 }
 
+// taskViewer builds the task-visibility scope for the current caller: managers
+// and admins see every task; everyone else sees only public tasks plus their own
+// (assigned to them or created by them).
+func taskViewer(c echo.Context) db.TaskViewer {
+	role, _ := c.Get("user_role").(string)
+	return db.TaskViewer{
+		Email:     getUserEmail(c),
+		CanSeeAll: role == "manager" || role == "admin",
+	}
+}
+
+// canViewTask mirrors db.TaskViewer for single by-id fetches (which aren't
+// SQL-filtered): reports whether the caller may see this task under the privacy
+// rule. A hidden task is treated as not-found (404), never 403 — don't reveal it.
+func canViewTask(c echo.Context, t *db.Task) bool {
+	if !t.Private {
+		return true
+	}
+	if role, _ := c.Get("user_role").(string); role == "manager" || role == "admin" {
+		return true
+	}
+	email := getUserEmail(c)
+	return t.Assignee == email || t.CreatedBy == email
+}
+
 // getOrgID returns the organization ID set by AuthMiddleware, or 0 if not set.
 func getOrgID(c echo.Context) int {
 	if id, ok := c.Get("org_id").(int); ok {
@@ -2031,7 +2056,7 @@ func (s *Server) handleListTasks(c echo.Context) error {
 		TaskType: c.QueryParam("task_type"),
 		Assignee: c.QueryParam("assignee"),
 	}
-	items, total, err := s.db.PaginatedTasks(c.Request().Context(), orgID, params)
+	items, total, err := s.db.PaginatedTasks(c.Request().Context(), orgID, taskViewer(c), params)
 	if err != nil {
 		return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
 	}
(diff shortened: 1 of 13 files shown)


unidoc-alip (approved):
Closes #177 cleanly and follows the #174 precedent (`resolveTaskID` et al.). `resolveChangeID`/`resolveObjectiveID`/`resolveSystemID`/`resolveAssetID` correctly fall back to a per-org-unique lookup (`identifier`/`display_id` all carry `UNIQUE(organization_id, ...)` constraints) when the suggestion's `EntityID` isn't numeric, fixing the wrong-row-on-apply bug for change/objective/system/asset updates. The new Python integration test exercises all four types end-to-end.

  Note: this branch is stacked on the unmerged `feat/task-visibility` branch (commit `b0646f5`), so the diff GitHub shows also includes that unrelated commit. That work has its own review; this review scopes only to `7bc8297`, the actual #177 fix.

  ---

  ### 1. [nit] `internal/isms/api/server.go` line 3218

  Unlike `resolveTaskID`/`resolveIncidentID`/etc. (which 400 via `errInvalidID` on a malformed, non-numeric, wrong-prefix param), these four helpers skip the prefix check entirely and always fall through to the DB lookup — so a garbage param and a well-formed-but-missing identifier both surface as the same "not found" error class. The PR description explains this is intentional (no fixed prefix works for all four: assets use both `ASSET-`/`AST-`, and objective identifiers are program-key-scoped). Worth a one-line code comment calling out that this is a deliberate 400-vs-404 tradeoff, not an oversight, since it's easy for a future reader to "fix" it back to the `resolveTaskID` shape and reintroduce a wrong assumption.

  Not a correctness bug — org-scoped `UNIQUE` constraints on `identifier`/`display_id` mean there's no ambiguity risk, and every caller already treats the resolve error as "not found" regardless of cause.

  ---

  ### 2. [nit] `internal/isms/api/api_suggestions.go` line 542

  `parseEntityID` is still used for stale-detection at lines 219/389/420/1671 (and for other entity types' apply paths, which are fine since they resolve by primary key). The PR body flags this as a known, deliberately out-of-scope follow-up for staleness-warning accuracy on change/objective/system/asset — no action needed here, just confirming the described scope matches the diff.

  ---

PR #181, opened 2026-07-15
fix(suggestions): confirm a submitted suggestion is in review (#167)

### LINKED ISSUE #167: Suggestions: confirm a new suggestion is in review with managers (don't just disappear)
After creating a suggestion it disappears from the creator's view with no feedback, so it feels lost/dropped. Give clear confirmation that it was submitted and is now **in review with managers** — a success toast, and ideally a visible "in review" state so the contributor can see their suggestion is pending, not gone.

unidoc-ahall:
Closes #167

## What
Submitting a suggestion now shows a confirmation: **"Suggestion submitted — a manager will review it before it takes effect."**

## Why
`SuggestNewButton` closed the form and emitted `created` with no feedback. On a list view the new suggestion isn't rendered there (it goes to the entity's suggestions / the inbox for manager review), so from the submitter's side it looked like nothing happened — the suggestion "just disappeared."

## How
A success toast on both create paths — `SuggestNewButton` (list-view "Suggest new") and the `SuggestionPanel` inline create form. Frontend-only; no API change.

## Testing
`just build-web` passes. Manual: submit a suggestion as a contributor → toast confirms it's pending manager review.

### DIFF
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -195,6 +195,31 @@ func getUserEmail(c echo.Context) string {
 	return ""
 }
 
+// taskViewer builds the task-visibility scope for the current caller: managers
+// and admins see every task; everyone else sees only public tasks plus their own
+// (assigned to them or created by them).
+func taskViewer(c echo.Context) db.TaskViewer {
+	role, _ := c.Get("user_role").(string)
+	return db.TaskViewer{
+		Email:     getUserEmail(c),
+		CanSeeAll: role == "manager" || role == "admin",
+	}
+}
+
+// canViewTask mirrors db.TaskViewer for single by-id fetches (which aren't
+// SQL-filtered): reports whether the caller may see this task under the privacy
+// rule. A hidden task is treated as not-found (404), never 403 — don't reveal it.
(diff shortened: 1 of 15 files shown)


unidoc-alip (approved):
Clean, well-scoped fix for #167. Both suggestion-creation paths (`SuggestNewButton`, `SuggestionPanel`) now show a success toast confirming the suggestion is pending manager review, addressing the "it just disappeared" complaint. The toast message is a static string (no injection surface), and fires independent of component teardown since `useToast`'s state is a module-level singleton rendered at `App.vue`'s root — it survives whatever the parent does in its `created` handler. No correctness, security, or test-coverage concerns in the isolated diff.

  This is only a partial close of #167: the issue also asked "ideally" for a persistent visible in-review state on the suggestion itself, which this PR doesn't add (toast-only, disappears after 3s). Worth a one-line call-out in case a follow-up should be tracked rather than treating the issue as fully resolved.

  ---

 ### 1. [nit] `web/src/components/SuggestionPanel.vue` line 161

 Minor style inconsistency: `SuggestNewButton.vue` explains the "why" behind this toast in a multi-line comment above the call, while here it's just a trailing `// #167` tag. Not blocking — just noting for consistency if you touch this again.

PR #182, opened 2026-07-15
feat(suggestions): 'Suggested by' label on suggestion lists (#168)

### LINKED ISSUE #168: Show "Suggested by" on suggestions
Suggestions don't show who created them. Surface **"Suggested by"** wherever suggestions are listed and reviewed — the creating user, and human vs agent (via `suggested_by` / `suggested_by_type`, which are already stored). Lets managers see the source at a glance. Display-only; the data already exists.

unidoc-ahall:
Closes #168

## What
An explicit **"Suggested by"** label on the creator line of suggestions.

## Why / status
Assessed the issue: suggestions are shown in exactly two places — the per-entity **SuggestionPanel** (used by all 11 modules) and the **Inbox** — and both already surfaced the creator (`suggested_by`) and human-vs-agent (a purple **AI** badge). So the data was there, but the creator line was a bare email/name with no label. This adds the literal **"Suggested by"** prefix so the attribution reads clearly "at a glance" per the issue.

## How
Display-only: prefix the creator line with "Suggested by" in `SuggestionPanel` and `Inbox` (panel also appends "(AI)" for agent suggestions). No API change — the data already exists.

## Testing
`just build-web` passes.

### DIFF
--- a/web/src/components/SuggestionPanel.vue
+++ b/web/src/components/SuggestionPanel.vue
@@ -20,7 +20,7 @@
           <span v-if="sg.suggested_by_type === 'agent'" class="px-1 py-0.5 rounded text-[9px] bg-purple-500/15 text-purple-400">AI</span>
         </div>
         <div v-if="sg.rationale" class="text-sm text-slate-500">{{ sg.rationale }}</div>
-        <div class="text-[10px] text-slate-600">{{ sg.suggested_by }} · {{ formatDate(sg.created_at) }}</div>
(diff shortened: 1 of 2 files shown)


unidoc-alip (approved):
Small, display-only change that does exactly what issue #168 asked for: prefixes the suggestion creator line with "Suggested by" in both places suggestions are rendered (`SuggestionPanel.vue`, used by all 11 modules, and `Inbox.vue`). Verified via grep that these are the only two components referencing `suggested_by`, so coverage is complete. No API or data changes — `suggested_by` / `suggested_by_type` were already fetched and rendered, just unlabeled. No correctness, security, or test-coverage concerns for a change this size.

looks good to me

PR #184, opened 2026-07-15
fix(objectives): don't reuse a soft-deleted display_id on create

### LINKED ISSUE #183: Objective create reuses a soft-deleted display_id after a delete (duplicate-key 409)
`CreateObjective` (internal/isms/db/objectives.go) computes the next sequence as `SELECT MAX(seq_number) FROM objectives WHERE program_id = $1 AND deleted_at IS NULL`, then builds `display_id = "<program.key>-<seq>"`. But the table constraint is `UNIQUE(organization_id, display_id)` with **no** partial `WHERE deleted_at IS NULL`, so a soft-deleted objective still occupies its `display_id`.

Result: after any objective under a program is (soft-)deleted, the next create regenerates a `display_id` that is still taken by the deleted row → `duplicate key value violates unique constraint "objectives_organization_id_display_id_key"` (HTTP 409).

Surfaces on a persistent DB once an objective has been deleted (e.g. `test_create_writepath::test_objective_apply_matches_http` on the shared isms-test stack). CI on a fresh DB does not hit it.

Present in BOTH objective-create paths: `CreateObjective` (objectives.go) and the suggestion-apply create (suggestions_tx.go). Every other entity allocates via the `identifier_sequences` counter (`NextIdentifier`), which never reuses — objectives are the only exception.

**Fix:** count all rows (drop the `deleted_at IS NULL` filter) in both paths so `seq_number` is monotonic per program and a `display_id` is never reused. No schema change.

unidoc-ahall:
Closes #183

Pre-existing bug, independent of the 0.7.1 feature PRs — surfaced on the persistent isms-test DB (CI's fresh DB doesn't hit it).

## Cause
`CreateObjective` took the next seq as `MAX(seq_number) WHERE deleted_at IS NULL`, but `UNIQUE(organization_id, display_id)` is not partial — a soft-deleted objective still holds its `display_id`. After any delete, the next create regenerated a taken id → duplicate-key 409.

## Fix
Count all rows (drop the `deleted_at IS NULL` filter) so `seq_number` is monotonic per program; a `display_id` is never reused. Fixed in **both** objective-create paths (`CreateObjective` and the suggestion-apply create in `suggestions_tx.go`) — every other entity already uses the `identifier_sequences` counter, which never reuses. No schema change.

## Testing
`just build-go`, `just test-go` pass; the existing `test_create_writepath` objective test passes reliably once seq stops reusing a deleted display_id.

### DIFF
--- a/internal/isms/db/objectives.go
+++ b/internal/isms/db/objectives.go
@@ -74,10 +74,13 @@ func (d *DB) CreateObjective(ctx context.Context, orgID int, o *Objective) error
 		return fmt.Errorf("program not found: %w", err)
 	}
 
-	// Get next seq number for this program
+	// Get next seq number for this program. Count ALL rows incl. soft-deleted:
+	// display_id is UNIQUE(organization_id, display_id) with no partial index, so a
+	// soft-deleted objective still holds its display_id. Excluding deleted rows here
+	// let MAX+1 regenerate a taken display_id after any delete → duplicate-key 409.
 	var maxSeq int
 	_ = d.pool.QueryRow(ctx,
-		`SELECT COALESCE(MAX(seq_number), 0) FROM objectives WHERE program_id = $1 AND deleted_at IS NULL`,
+		`SELECT COALESCE(MAX(seq_number), 0) FROM objectives WHERE program_id = $1`,
 		o.ProgramID).Scan(&maxSeq)
 	o.SeqNumber = maxSeq + 1
 	o.DisplayID = fmt.Sprintf("%s-%d", prog.Key, o.SeqNumber)
--- a/internal/isms/db/objectives.go
+++ b/internal/isms/db/objectives.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+
+	"github.com/jackc/pgx/v5"
 )
 
 // Allowed enum values for objective fields. Mirrors schema CHECK constraints.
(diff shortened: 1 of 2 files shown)


unidoc-alip (changes requested):
Correct, minimal fix for the soft-delete display_id reuse bug — dropping the `deleted_at IS NULL` filter so `seq_number` is monotonic per program is the right call given the non-partial `UNIQUE(organization_id, display_id)` constraint, and it's applied consistently in both create paths. One residual concern, verified against the code below: the underlying allocation is still a non-atomic `SELECT MAX` + separate `INSERT`, so a concurrent-create race can still produce the same duplicate-key 409 this PR is fixing, just via a different trigger. Not a regression from this PR, so not blocking, but worth a follow-up given `identifier_sequences`/`NextIdentifier` already exists and even lists `"objective"` in its entity-type prefix map.

---

### 1. [should-fix] `internal/isms/db/objectives.go` line 81

This still allocates `seq_number` via `SELECT MAX(seq_number)` followed by a separate `INSERT`, with no row lock or atomic counter in between. Both `UNIQUE(organization_id, display_id)` and `UNIQUE(program_id, seq_number)` are non-partial, so two concurrent `CreateObjective` calls for the same program can read the same `maxSeq` and then collide on insert — reproducing the same duplicate-key 409 this PR sets out to fix, just via a race instead of soft-delete reuse.

Verified: `handleCreateObjective` (`internal/isms/api/api_objectives.go:311`) has no locking/serialization before calling `s.db.CreateObjective` — no mutex, no advisory lock, no per-program queueing. The schema confirms both constraints are non-partial (`migrations/20260327000000_initial_schema.sql:971,973`). `pgxHTTPError` (`internal/isms/api/middleware.go:662-663`) maps Postgres code `23505` (unique violation) straight to HTTP 409 — so a genuine concurrent-create race on the same program reproduces the exact duplicate-key 409 symptom this PR eliminates for the soft-delete case. This is a real, currently-reachable gap, not theoretical.

Cheapest correct fix without a schema change: serialize concurrent creates on the same program by taking a row lock before computing `maxSeq`, all inside one transaction:

```go
return d.pool.BeginFunc(ctx, func(tx pgx.Tx) error {
    // Lock the program row so concurrent creates for it serialize.
    if _, err := tx.Exec(ctx, `SELECT id FROM programs WHERE id = $1 FOR UPDATE`, o.ProgramID); err != nil {
        return fmt.Errorf("lock program: %w", err)
    }
    var maxSeq int
    if err := tx.QueryRow(ctx,
        `SELECT COALESCE(MAX(seq_number), 0) FROM objectives WHERE program_id = $1`,
        o.ProgramID).Scan(&maxSeq); err != nil {
        return err
    }
    o.SeqNumber = maxSeq + 1
    o.DisplayID = fmt.Sprintf("%s-%d", prog.Key, o.SeqNumber)
    // ... existing INSERT, run via tx.QueryRow instead of d.pool.QueryRow
    return tx.QueryRow(ctx, `INSERT INTO objectives (...) ... RETURNING id, created_at, updated_at`,
        /* args */).Scan(&o.ID, &o.CreatedAt, &o.UpdatedAt)
})
```

A more thorough alternative matching the rest of the codebase's convention: add a program-scoped sequence table (mirroring `identifier_sequences` but keyed by `program_id` instead of `(organization_id, entity_type)`, since `display_id` here is `{program.Key}-{seq}`, not org-scoped) and allocate via the same `INSERT ... ON CONFLICT DO UPDATE ... RETURNING` pattern `NextIdentifier` already uses. That's a schema migration, so only worth it if objectives are expected to see real concurrent-create traffic per program; the `FOR UPDATE` lock above is sufficient otherwise.

---

### 2. [nit] `internal/isms/db/suggestions_tx.go` line 662

Same non-atomic MAX+1 pattern duplicated here as in `CreateObjective` — if the fix above lands, this call site needs the same treatment (or should be refactored to share one implementation, since the two functions are now near-identical modulo the parameterized query interface).

Verified: `CreateObjectiveTx` has exactly one caller, `internal/isms/api/api_suggestions.go:1373` (the suggestion-apply handler), which already runs inside a transaction (`tx pgx.Tx` is passed in) but never takes a lock on the program row before reading `maxSeq` — same TOCTOU window as finding 1, just inside an externally-supplied transaction rather than one this function owns.

Since this function already receives a live `tx pgx.Tx`, the fix is simpler than finding 1 — no new `BeginFunc` needed, just add the row lock before computing `maxSeq`, inside the same transaction:

```go
// Lock the program row so concurrent creates for it serialize within this tx.
if _, err := tx.Exec(ctx, `SELECT id FROM programs WHERE id = $1 FOR UPDATE`, o.ProgramID); err != nil {
    return fmt.Errorf("lock program: %w", err)
}
var maxSeq int
_ = tx.QueryRow(ctx,
    `SELECT COALESCE(MAX(seq_number), 0) FROM objectives WHERE program_id = $1`,
    o.ProgramID).Scan(&maxSeq)
```

Note this lock is only effective against other transactions that also take it — so this fix and finding 1's fix must land together, otherwise `CreateObjective` (unlocked) can still race against `CreateObjectiveTx` (locked). If both call sites converge on the same locking discipline, consider factoring the "lock program + compute next seq + build display_id" logic into one shared helper taking either `*pgxpool.Pool` or `pgx.Tx` to prevent future drift.

unidoc-ahall:
Good catch — the soft-delete fix left the concurrent-create race open. Both findings addressed in the latest commit:

**1. `CreateObjective` non-atomic MAX+1 — [should-fix] ✅** Allocation + insert now run in one transaction (`pool.Begin`), and a new shared helper `nextObjectiveSeqTx(tx, programID)` takes `SELECT id FROM programs WHERE id = $1 FOR UPDATE` before computing `MAX(seq_number)+1`. Concurrent creates for the same program now serialize on the program row instead of racing.

**2. `CreateObjectiveTx` same pattern — [nit] ✅** Routed through the same `nextObjectiveSeqTx` (it already runs inside a caller-supplied `tx`, so it just takes the lock there). Since both paths take the **same** program-row lock, `CreateObjective` and `CreateObjectiveTx` also serialize against each other — landing together as you flagged.

Factored the "lock program + next seq" logic into the single `nextObjectiveSeqTx` helper so the two call sites can't drift, per your suggestion. Went with the `FOR UPDATE` lock rather than a new program-scoped sequence table — no schema change, and objectives don't see high per-program create concurrency.

No dedicated race test (it'd be inherently flaky); the existing objective + `test_create_writepath` tests pass, and `just fmt` / `build-go` / `test-go` are green.

unidoc-alip (approved):
looks good to me

PR #186, opened 2026-07-16
feat(comments): reference entities with # + add objectives/programs to search (#171)

### LINKED ISSUE #171: Reference entities (tasks, CAs…) in comments + add missing entity pickers
People want to reference another entity from a comment — e.g. commenting on TASK-1 "closing this, I split it into TASK-9" with TASK-9 as a link. Today that isn't possible.

**Two gaps:**

1. **Comments don't support entity references at all.** `CommentsPanel` uses `MentionTextarea`, which only does `@` user-mentions. There's no way to link a task (or any register entity) from a comment.

2. **The slash-command entity pickers are missing several types.** `useSlashCommands.js` has pickers for risk / legal / asset / supplier / system / document / incident — but **not task, corrective_action, change, objective, program**. So even in description/notes fields (MarkdownField) you can't `/task`.

**Scope:**
- Add the missing pickers to `useSlashCommands.js` + `entityFieldMap` (task, corrective_action, change, objective, program).
- Bring entity-reference support into the comment input (either swap `MentionTextarea` for `MarkdownField`, or add the slash/picker affordance to it) while keeping `@` user-mentions.
- Links should use the entity **identifier** (build on the deep-links work) — and note `incident`'s picker currently uses the numeric `id`, which should move to identifier too.

No schema change.

unidoc-ahall:
Closes #171

## What
Reference other entities from inside a comment, and fill the two register types that couldn't be picked as reference targets.

## How
- **`#` entity picker** (`MentionTextarea.vue`): typing `#` opens a searchable dropdown backed by the universal-search index and inserts the entity identifier (e.g. `#TASK-6`). `@` still mentions users — same component, second trigger. Works everywhere `MentionTextarea` is used (entity comments + document comments).
- **Deep-linked render** (`useMention.js` `renderMention`): a known-prefix identifier (`#RISK-3`, `#TASK-6`, …) becomes a link to the entity's page via the #166 identifier deep-links (org-aware `currentOrgPath`); unknown forms stay a highlighted reference.
- **Search picker gap** (`api_search.go`): objectives (`display_id`) and programs (`key`) were missing from the search index, so `ReferenceManager` couldn't find them to link — added both. `resolveEntityTitle` already resolves those identifier forms, so reference chips display correctly.

## Scope / follow-ups (small, tracked separately)
- Audit + audit findings aren't in search yet — needs an org-wide findings list method and handling their special `/audit/:tab/:itemId` routing.
- Objectives/programs enter the index on its TTL rebuild rather than an immediate `searchUpsert` (fine for a reference picker; other types upsert immediately).
- Comment reference links are plain `<a href>` (full navigation) rather than SPA-router pushes.
- Create-time linking via the (still-unwired) `ReferenceCollector` component.

## Testing
`just fmt` / `build-go` / `build-web` / `test-go` pass. The `#` picker and linkified render are interactive frontend with no JS test runner in the repo — verified manually. (A search-index integration test would be TTL-cache flaky, so not added.)

### DIFF
--- a/internal/isms/api/api_search.go
+++ b/internal/isms/api/api_search.go
@@ -364,6 +364,44 @@ func (s *Server) buildSearchIndex(orgID int) {
 		return entries
 	})
 
+	// Objectives (#171: were missing from search, so couldn't be picked as a
+	// reference target). display_id is the identifier form resolveEntityTitle +
+	// deep-links use.
+	collect(func() []SearchEntry {
+		items, err := s.db.ListObjectives(ctx, orgID, 0, "")
+		if err != nil {
+			return nil
+		}
+		entries := make([]SearchEntry, 0, len(items))
+		for _, o := range items {
+			entries = append(entries, SearchEntry{
+				Type:   "objective",
+				ID:     o.DisplayID,
+				Title:  o.Title,
+				Search: strings.ToLower(o.DisplayID + " " + o.Title + " " + o.Description),
+			})
+		}
+		return entries
+	})
+
+	// Programs (#171). Key is the identifier form resolveEntityTitle + deep-links use.
+	collect(func() []SearchEntry {
+		items, err := s.db.ListPrograms(ctx, orgID)
+		if err != nil {
+			return nil
+		}
+		entries := make([]SearchEntry, 0, len(items))
+		for _, p := range items {
+			entries = append(entries, SearchEntry{
+				Type:   "program",
+				ID:     p.Key,
+				Title:  p.Title,
+				Search: strings.ToLower(p.Key + " " + p.Title + " " + p.Description),
+			})
+		}
+		return entries
+	})
+
 	wg.Wait()
 
 	s.searchIndex.Set(orgID, all)
--- a/internal/isms/api/api_search.go
+++ b/internal/isms/api/api_search.go
@@ -145,6 +145,36 @@ func (idx *SearchIndex) Remove(orgID int, entryType, entryID string) {
 }
 
 // buildSearchIndex loads all entity types for an org and populates the index.
+// objectiveSearchEntries / programSearchEntries build the index rows for these
+// two types. Kept pure (no DB) so they're unit testable (#171 review). ID is the
+// identifier form (objective display_id, program key) that resolveEntityTitle and
+// the deep-link routes resolve.
+func objectiveSearchEntries(items []db.Objective) []SearchEntry {
+	entries := make([]SearchEntry, 0, len(items))
+	for _, o := range items {
+		entries = append(entries, SearchEntry{
+			Type:   "objective",
+			ID:     o.DisplayID,
+			Title:  o.Title,
+			Search: strings.ToLower(o.DisplayID + " " + o.Title + " " + o.Description),
+		})
+	}
+	return entries
+}
+
+func programSearchEntries(items []db.Program) []SearchEntry {
+	entries := make([]SearchEntry, 0, len(items))
+	for _, p := range items {
+		entries = append(entries, SearchEntry{
+			Type:   "program",
+			ID:     p.Key,
+			Title:  p.Title,
+			Search: strings.ToLower(p.Key + " " + p.Title + " " + p.Description),
+		})
+	}
+	return entries
+}
+
 func (s *Server) buildSearchIndex(orgID int) {
 	start := time.Now()
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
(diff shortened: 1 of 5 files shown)


unidoc-alip (changes requested):
Solid, well-scoped diff that reuses `MentionTextarea` for both `@` and `#` triggers and wires up entity search nicely. However, the PR's core stated goal — making objectives and programs pickable *and linkable* as reference targets — doesn't actually work in the normal case: `ENTITY_ROUTES` in `useMention.js` is a static map of fixed system prefixes, but `Program.Key` is an arbitrary per-org user string and `Objective.DisplayID` is built from it, so neither will match the `PROG` entry outside of an org literally naming a program `"PROG"`. Also a minor debounce race in the new entity search. No security issues found (separate security-focused pass came back clean).

---

### 1. [must-fix] `web/src/composables/useMention.js` line 7

`ENTITY_ROUTES` is a static map keyed by literal, fixed system-generated prefixes (`RISK`, `TASK`, `CA`, …, corresponding to `Identifier` values like `TASK-6` from `NextIdentifier`). Programs and objectives don't fit this model:

- `Program.Key` is an arbitrary, per-org, user-chosen string (e.g. `ISMS`, `SEC`), not the literal string `PROG` — `PROG` is only the prefix of `Program.Identifier`, a separate field the frontend doesn't use anywhere (`Programs.vue` uses `.key` throughout: route, `entityId`, lookups). So `ENTITY_ROUTES['PROG']` only matches if an org happens to name a program's key literally `"PROG"`.
- Referencing a program via the new `#` picker inserts `#ISMS ` (bare key, no digit suffix), which doesn't even match the linkification regex (`#([A-Z][A-Z0-9]*)-(\d+)`), so it renders as plain unstyled text — not even highlighted.
- `Objective.DisplayID` (`fmt.Sprintf("%s-%d", programKey, seqNumber)`, e.g. `ISMS-1`) does match the regex, but since `ENTITY_ROUTES` has no entry for the actual per-org program key (only the unrelated literal `PROG`), it falls into the "unknown form" branch and renders as a highlighted, non-clickable span.

Net effect: in the common case, neither programs nor objectives picked via the new search-backed `#` picker produce a working deep-link — which is exactly the gap this PR claims to close ("fill the two register types that couldn't be picked as reference targets"). This isn't listed in the PR's "Scope/follow-ups" section as a known limitation.

Suggested fix — the static-prefix-map approach can't work for these two types because their identifiers are dynamic and per-org. Recommended fix, following the existing `useMembers.js` shared-cache pattern already used for `@`-mention data in this same component family:

1. Add a small composable, e.g. `web/src/composables/usePrograms.js`, mirroring `useMembers.js`: a module-level `ref([])` cache populated once via the existing `api.getPrograms()` (already used by `Programs.vue`), exposing the list of `{ key, ... }`.
2. In `useMention.js`, import that cache and change `renderMention`'s entity-link logic to:
   - First check the static `ENTITY_ROUTES` map (unchanged, covers the fixed-prefix types).
   - If the matched prefix isn't in `ENTITY_ROUTES`, check it against the cached program keys: if `prefix` matches a known program key, link `#KEY-N` to `/objectives/${prefix}-${num}` (an objective) rather than `/programs/...`.
   - Add a second, narrower regex pass (or extend the existing one to make the `-num` suffix optional) to catch bare `#KEY` references with no digit suffix, matched only against the cached program-key set, and link those to `/programs/${key}`.
   - Keep the existing "unknown → highlighted span" fallback for anything matching neither a fixed prefix nor a cached program key (handles the cache-not-yet-loaded / genuinely unknown case gracefully, same failure mode as today).
3. The cache staleness caveat is acceptable — worst case is graceful degradation to a highlighted, unlinked reference, matching current behavior for genuinely unmapped types.

---

### 2. [should-fix] `web/src/components/MentionTextarea.vue` line 110

`searchEntities`'s `setTimeout`-based debounce only cancels *unfired* timers via `clearTimeout`. If the network round-trip for `api.search` exceeds the 200ms debounce window, two requests can be in flight concurrently; a slower earlier response resolving after a faster later one will overwrite `entityResults` with stale results for a now-different query. Low severity (self-corrects on the next keystroke), but worth guarding with a query token / call-id check before assigning `entityResults.value`.

Suggested fix — add a monotonically increasing request sequence number and ignore stale responses:

```js
let searchTimer = null
let searchSeq = 0
function searchEntities(q) {
  clearTimeout(searchTimer)
  if (!q) { entityResults.value = []; return }
  searchTimer = setTimeout(async () => {
    const seq = ++searchSeq
    try {
      const res = await api.search(q)
      if (seq !== searchSeq) return // a newer search has since started; drop this response
      const data = (res && (res.data || res)) || []
      entityResults.value = data.slice(0, 8)
    } catch {
      if (seq === searchSeq) entityResults.value = []
    }
  }, 200)
}
```

Equivalent alternative: use an `AbortController` per request and abort the previous one when a new search starts, if `api.search` is extended to accept a signal. The sequence-number approach above is the minimal fix and needs no change to the shared `api` client.

---

### 3. [nit] `internal/isms/api/api_search.go` line 367

No new Go test covers the two added search-index collectors (objectives/programs now appearing in `buildSearchIndex`/`handleUniversalSearch` results). The PR body explains the lack of JS tests (no test runner) but the backend addition here is straightforward to unit test and isn't mentioned. Note: there's no existing test file for `api_search.go` at all, so this isn't a regression specific to this PR — just a low-cost opportunity given how mechanically testable the new collectors are.

Suggested fix — add `internal/isms/api/api_search_test.go` covering `buildSearchIndex`, asserting:
- An objective's `SearchEntry` has `Type: "objective"`, `ID` equal to its `DisplayID` (not numeric ID), and `Search` contains the lowercased title.
- A program's `SearchEntry` has `Type: "program"`, `ID` equal to its `Key`, and `Search` contains the lowercased title.
- Soft-deleted objectives/programs (`deleted_at IS NOT NULL`) are excluded (already true via the underlying queries — worth locking in with a test given the prior soft-delete/display_id-reuse bug fixed in PR #184).

If a full DB-backed test is too heavy for this file's existing conventions, at minimum extract the entry-building loop into a small pure helper function (`[]Objective`/`[]Program` → `[]SearchEntry`) so it's testable without a DB fixture.

unidoc-ahall:
All three addressed — you were right that the static prefix map couldn't work for the two dynamic-identifier types.

**1. Objective/program don't link — [must-fix] ✅** Replaced the `PROG` map entry with a live program-key cache: new `usePrograms.js` (mirrors `useMembers.js`) holds the org's program keys. `renderMention` now resolves in a single pass (`#([A-Z][A-Z0-9]*)(?:-(\d+))?`):
- fixed system prefix (`TASK`, `RISK`, …) → its register;
- else if `prefix` is a known program key → `#KEY-N` is an **objective** (`display_id = KEY-seq`) → `/objectives/KEY-N`, and bare `#KEY` is the **program** → `/programs/KEY`;
- otherwise → highlighted span / untouched (also the graceful path while the cache is still loading).
Single pass so a linkified `#KEY-N` can't be re-matched as a bare `#KEY`.

**2. Debounce out-of-order race — [should-fix] ✅** Added a `searchSeq` token; a response is dropped if a newer search has started before it resolves.

**3. No Go test for the new collectors — [nit] ✅** Extracted `objectiveSearchEntries` / `programSearchEntries` as pure helpers and added `api_search_test.go` asserting `Type`, the **identifier-form** `ID` (display_id / key, not the numeric id), and `Search` contents (+ empty-input).

`just fmt` / `build-go` / `build-web` / `test-go` pass. The `#`-picker + linkified render remain manually verified (no JS test runner).

unidoc-alip (approved):
looks good to me

PR #189, opened 2026-08-04
build(freebsd): jail support — goreleaser target + rc.d service + docs

### LINKED ISSUE #188: FreeBSD jail support: goreleaser freebsd build + rc.d service + install docs
Run ISMS from FreeBSD jails. The binary is pure-Go (CGO off) and already
cross-compiles to freebsd/amd64+arm64 cleanly; the app is self-contained (web UI +
migrations are embedded). What's missing is packaging:

- **goreleaser**: add `freebsd` to the build `goos` (openbsd is already there), so
  the release produces `isms_<ver>_freebsd_{amd64,arm64}.tar.gz`.
- **rc.d service script** so a jail install is `service isms enable` + rc.conf,
  shipped inside the archive.
- **Config sample + install docs** (`DATABASE_URL` is not `ISMS_`-prefixed; the
  listen address is the `--addr` flag; templates path on disk).

Build/packaging only — no app behaviour change, no schema/migration, so safe in
0.7.1 (an adopter needs it to run the system on FreeBSD).

### Out of scope (separate follow-up)
Native `pkg install isms`. goreleaser/nfpm can't build a FreeBSD `.pkg` (deb/rpm/
apk only), so the release ships only the tarball. A `.pkg` is a separate
`pkg create` step **on FreeBSD** (staged binary + rc.d + `+MANIFEST`), then
`pkg repo` to build the catalog, hosted over HTTP as a self-hosted pkg repo — or,
later, an official FreeBSD port (`USES=go:modules,nodejs`, submitted to
freebsd-ports, committer-approved + maintainer-tracked). Neither is needed to run
in a jail.

unidoc-ahall:
Closes #188

## What
Run ISMS from FreeBSD jails. **Build/packaging only** — no app code change, no schema/migration — so it's safe in 0.7.1 (needed to actually run the system on FreeBSD).

## How
- **goreleaser** (`.goreleaser.yaml`): add `freebsd` to the build `goos` (OpenBSD was already there; the binary is pure-Go/CGO-off and cross-compiles cleanly). The existing `tar.gz` archive now yields `isms_<ver>_freebsd_amd64.tar.gz` + `arm64`, and every archive ships the rc.d script + config sample + `FREEBSD.md`.
- **rc.d service** (`contrib/rc.d/isms`): `rc.subr` script running `isms server` under `daemon(8)` as the `isms` user, sourcing `ISMS_*` / `DATABASE_URL` from an env file.
- **Docs** (`contrib/freebsd/`): `isms.env.sample` + `README.md`. Notes the gotchas: `DATABASE_URL` is **not** `ISMS_`-prefixed, and the listen address is the `--addr` flag (default `:8080`), not an env var.

## Out of scope (separate follow-up)
Native `pkg install isms` — goreleaser can't build a FreeBSD `.pkg`, so the release ships only the tarball. A `.pkg` is a separate `pkg create` step on FreeBSD, then `pkg repo` + host as a self-hosted repo (or an official FreeBSD port later). Not needed to run in a jail.

## Testing
`goreleaser check` passes; `freebsd/amd64` + `freebsd/arm64` cross-compile clean; `just build-go` passes. Release artifacts are produced by `just release`.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -37,9 +37,15 @@ universal_binaries:
 archives:
   - formats: [tar.gz]
     name_template: "isms_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
-    # Ship the FreeBSD rc.d service script + a config sample so a jail install is
-    # copy-into-place. Harmless on other OSes (they just ignore rc.d/isms).
+    # Specifying `files` REPLACES goreleaser's default list, so the defaults
+    # (LICENSE/README/CHANGELOG) must be re-listed explicitly or every platform's
+    # tarball silently loses them — a licence-distribution regression, not just a
+    # FreeBSD one. Keep the defaults, then add the FreeBSD rc.d script + config
+    # sample so a jail install is copy-into-place (harmless on other OSes).
     files:
+      - LICENSE*
+      - README*
+      - CHANGELOG*
       - src: contrib/freebsd/rc.d/isms
         dst: rc.d/isms
       - src: contrib/freebsd/isms.env.sample
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -19,6 +19,7 @@ builds:
       - linux
       - darwin
       - openbsd
+      - freebsd
     goarch:
       - amd64
       - arm64
@@ -36,6 +37,15 @@ universal_binaries:
 archives:
   - formats: [tar.gz]
     name_template: "isms_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
+    # Ship the FreeBSD rc.d service script + a config sample so a jail install is
+    # copy-into-place. Harmless on other OSes (they just ignore rc.d/isms).
+    files:
+      - src: contrib/freebsd/rc.d/isms
+        dst: rc.d/isms
+      - src: contrib/freebsd/isms.env.sample
+        dst: isms.env.sample
+      - src: contrib/freebsd/README.md
+        dst: FREEBSD.md
 
 checksum:
   name_template: checksums.txt
--- a/contrib/freebsd/README.md
+++ b/contrib/freebsd/README.md
@@ -1,4 +1,7 @@
-# Running ISMS on FreeBSD (jails)
+# Running ISMS on FreeBSD
+
+Works on any FreeBSD system — a bare-metal/VM host or a jail; there is nothing
+jail-specific here (a jail is just a FreeBSD userland with the same rc(8) system).
 
 ISMS ships as a single self-contained binary — the web UI and DB migrations are
 embedded, so there is nothing else to deploy for the app itself. You still need a
--- a/contrib/freebsd/README.md
+++ b/contrib/freebsd/README.md
@@ -1,0 +1,78 @@
+# Running ISMS on FreeBSD (jails)
+
+ISMS ships as a single self-contained binary — the web UI and DB migrations are
+embedded, so there is nothing else to deploy for the app itself. You still need a
+PostgreSQL database (commonly its own jail/host) and the templates repo on disk.
+
+The release tarball `isms_<version>_freebsd_amd64.tar.gz` (also `arm64`) contains:
+
+```
+isms              # the binary
+rc.d/isms         # the service script (this dir)
+isms.env.sample   # configuration sample
+FREEBSD.md        # this file
+```
+
+## Install
+
+```sh
+# In the jail, as root:
+fetch https://github.com/unidoc/isms/releases/download/vX.Y.Z/isms_X.Y.Z_freebsd_amd64.tar.gz
+tar xzf isms_X.Y.Z_freebsd_amd64.tar.gz
+
+install -m 755 isms /usr/local/bin/isms
+install -m 755 rc.d/isms /usr/local/etc/rc.d/isms
+
+# Service account (no login, no home needed — git storage lives under a data dir
+# you choose via config):
+pw useradd isms -d /nonexistent -s /usr/sbin/nologin -c "ISMS service"
+
+# Configuration:
+install -m 600 -o isms isms.env.sample /usr/local/etc/isms.env
+$EDITOR /usr/local/etc/isms.env        # set DATABASE_URL, ISMS_TEMPLATE_PATH, ISMS_BASE_URL
+```
+
+## Database + templates
+
+- Create the PostgreSQL role/database and point `DATABASE_URL` at it (note:
+  `DATABASE_URL`, not `ISMS_DATABASE_URL`).
+- Apply the schema — migrations are embedded:
+
+  ```sh
+  env $(grep -v '^#' /usr/local/etc/isms.env | xargs) isms migrate
+  ```
+
+- Put the standard templates repo at `ISMS_TEMPLATE_PATH`.
+
+## Enable + run
+
+```sh
+sysrc isms_enable=YES
+service isms start
+service isms status
+```
+
+By default it listens on `:8080` and runs as the `isms` user. To change the
+listen address (e.g. bind to loopback behind a reverse proxy):
+
+```sh
+sysrc isms_args="server --addr 127.0.0.1:8080"
+service isms restart
+```
+
+## rc.conf variables
+
+| variable        | default                      | meaning                          |
+| --------------- | ---------------------------- | -------------------------------- |
+| `isms_enable`   | `NO`                         | enable the service               |
+| `isms_user`     | `isms`                       | user to run as                   |
+| `isms_env_file` | `/usr/local/etc/isms.env`    | file with `ISMS_*` / `DATABASE_URL` config |
+| `isms_args`     | `server`                     | `isms` subcommand + flags        |
+
+Logs go to syslog via `daemon(8)`; the pidfile is `/var/run/isms.pid`.
+
+## `pkg install` (later)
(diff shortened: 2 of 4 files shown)


unidoc-alip (changes requested):
Good idea and the packaging shape is right — `freebsd` cross-compiles clean (verified `freebsd/amd64` + `freebsd/arm64`, CGO off), the rc.d script is valid `sh` (`sh -n` passes), and the archive `src`/`dst` renames land exactly as intended, mode 755 preserved (verified with a real `goreleaser release --snapshot`).

But the service as shipped does not start ISMS. `isms server` is a cobra *group* with no `Run` — it prints help and exits 0 (verified by running it), so `isms_args="server"` makes `service isms start` report success and start nothing. The subcommand is `isms server serve`. Two more blockers: the sample env file omits `ISMS_DATA_DIR`, which `blob.NewFromEnv` treats as required and `log.Fatal`s on; and adding `archives.files` replaces goreleaser's default file list, so **every** platform's tarball now loses `LICENSE` and `README.md` (verified: master's tarball ships both, this branch's ships neither). There is also a local privilege-escalation path — the env file is installed writable by the `isms` service user but sourced as **root** in `start_precmd`.

What was actually executed for this review: `GOOS=freebsd GOARCH={amd64,arm64} CGO_ENABLED=0 go build` (both pass), `sh -n contrib/freebsd/rc.d/isms` (passes), `goreleaser release --snapshot` against this branch's config and master's config to diff archive contents, and `isms server` against a locally built binary. I could **not** execute `rc.subr`/`daemon(8)` — no FreeBSD host — so the runtime behaviour of the script itself is reviewed statically; the command-path defect below is confirmed against the actual binary.

---

### 1. [must-fix] `contrib/freebsd/rc.d/isms` line 32

`isms server` does not start the server — it's a cobra command group with no `Run`, so it prints its help text and exits 0 (`cmd/isms/main.go:86-107`). The serve subcommand is `serve` (`cmd/isms/serve.go:27`).

Effect: `service isms start` succeeds, `daemon(8)` reaps an immediately-exited child, and nothing is listening. Verified by running the built binary:

```
$ isms server; echo "EXIT=$?"
Server-side commands for managing the ISMS platform. ...
EXIT=0
```

```sh
: ${isms_args:="server serve"}
```

The header comment on line 20 needs the same fix.

---

### 2. [must-fix] `contrib/freebsd/isms.env.sample` line 11

`ISMS_DATA_DIR` is missing, and it is not optional — `blob.NewFromEnv()` returns `ISMS_DATA_DIR is required` when unset, and `api.New` handles that with `log.Fatal` (`internal/isms/api/server.go:255-257`). It's also what locates the per-org bare git repos (`$ISMS_DATA_DIR/repos/{slug}.git`, see `cmd/isms/org.go:105`).

So once finding #1 is fixed, a by-the-book install still dies at startup. Add it to the `--- Required ---` block:

```sh
# Base data directory: org git repos ($ISMS_DATA_DIR/repos/{slug}.git) and,
# with the file storage backend, branding + evidence blobs.
ISMS_DATA_DIR=/var/db/isms
```

Worth noting in `README.md` too — the `pw useradd` comment says "git storage lives under a data dir you choose via config" but never names the variable, and the service user gets `-d /nonexistent`, so the directory needs creating and chowning to `isms`.

---

### 3. [must-fix] `.goreleaser.yaml` line 42

Specifying `archives.files` **replaces** goreleaser's default list (`LICENSE*`, `README*`, `CHANGELOG*` and lowercase variants) rather than adding to it — so this drops `LICENSE` and `README.md` from every platform's tarball, not just FreeBSD's. Both exist at repo root, so this is a live regression (and a licence-distribution one).

Verified with a real snapshot build of both configs:

```
# master config
$ tar tzf dist/isms_..._linux_amd64.tar.gz
LICENSE
README.md
isms

# this branch
$ tar tzf dist/isms_..._linux_amd64.tar.gz
FREEBSD.md
isms.env.sample
rc.d/isms
isms          <- LICENSE and README.md gone
```

Fix — re-add the defaults explicitly:

```yaml
files:
  - LICENSE*
  - README*
  - CHANGELOG*
  - src: contrib/freebsd/rc.d/isms
    dst: rc.d/isms
  - src: contrib/freebsd/isms.env.sample
    dst: isms.env.sample
  - src: contrib/freebsd/README.md
    dst: FREEBSD.md
```

Side note on the PR's testing claim: `goreleaser check` only validates config schema — it says nothing about archive contents, which is why this slipped through.

---

### 4. [must-fix] `contrib/freebsd/README.md` line 34

`-o isms` makes the env file **writable by the service account**, and `isms_precmd` sources it as **root** (`. "${isms_env_file}"` — arbitrary shell, run before privileges are dropped). That's a local privilege-escalation path: anything that can write as `isms` (a compromised web process, a path-traversal write bug) appends a command to `isms.env` and gets root on the next `service isms restart`.

The `isms` user never needs to read this file — `daemon(8)` inherits the environment from the root-side `start_precmd` and passes it down. So:

```sh
install -m 600 -o root -g wheel isms.env.sample /usr/local/etc/isms.env
```

---

### 5. [must-fix] `contrib/freebsd/README.md` line 45

Wrong command path — `migrate` lives under the `server` group (`cmd/isms/main.go:93-102`), so this is `isms server migrate`. `isms migrate` exits non-zero with `unknown command`.

```sh
env $(grep -v '^#' /usr/local/etc/isms.env | xargs) isms server migrate
```

Same defect at line 62 and in `isms.env.sample:18`: `sysrc isms_args="server --addr 127.0.0.1:8080"` must be `sysrc isms_args="server serve --addr 127.0.0.1:8080"` (see finding #1).

---

### 6. [should-fix] `contrib/freebsd/README.md` line 75

"Logs go to syslog via `daemon(8)`" isn't true for this invocation. `daemon -f` *discards* stdin/stdout/stderr to `/dev/null`; syslog forwarding needs `-S` (or `-o file`). As written, ISMS's log output — including the `log.Fatal` from finding #2 — vanishes, which makes a failed start undiagnosable.

Either add `-S` (and optionally `-T isms` for the tag) to `command_args` in the rc.d script:

```sh
command_args="-f -S -T ${name} -p ${pidfile} -u ${isms_user} ${procname} ${isms_args}"
```

or correct this line.

---

### 7. [should-fix] `contrib/freebsd/rc.d/isms` line 14

`ISMS_DATABASE_URL` doesn't exist — the variable is the unprefixed `DATABASE_URL` (`cmd/isms/serve.go:30`). Ironic given this is the exact gotcha the PR description and `isms.env.sample` call out. Also worth adding `ISMS_DATA_DIR` here (finding #2):

```
# isms.env.sample. Required at minimum: DATABASE_URL, ISMS_DATA_DIR and
# ISMS_TEMPLATE_PATH.
```

---

### 8. [nit] `contrib/freebsd/rc.d/isms` line 37

Consider `daemon -r` (restart the child if it exits) so a crashed ISMS comes back rather than leaving a dead service. Worth noting it would have made finding #1 loop on the help-printer, so add it only after that's fixed.

---

### 9. [nit] `.goreleaser.yaml` line 40

Confirmed harmless but noisy: `FREEBSD.md`, `isms.env.sample` and `rc.d/isms` land in the linux, darwin and openbsd tarballs too (verified in the snapshot build). If you'd rather keep those clean, a second `archives` entry keyed on `{{ eq .Os "freebsd" }}` would scope them — happy to leave as-is given the comment already flags it.

unidoc-ahall:
Thanks — every finding was correct, and the real-snapshot verification caught two things `goreleaser check` can't. All addressed in the latest push.

**must-fix**
1. `server serve` — you're right, `isms server` is a group that prints help and exits 0; the service started nothing. Fixed in the rc.d default, the header comment, and every doc/sample occurrence (incl. `server migrate`, finding #5).
2. `ISMS_DATA_DIR` added to the Required block. While verifying I hit the next one right behind it: `ISMS_STORAGE_BACKEND` is *also* required (`blob.NewFromEnv` errors on empty), so a by-the-book install died there too — it was commented under Optional. Both are now Required, and the README creates + chowns the data dir.
3. `archives.files` replacing the defaults — confirmed and fixed by re-listing `LICENSE*`/`README*`/`CHANGELOG*`. Verified with a real `goreleaser release --snapshot`: both the linux and freebsd tarballs now contain `LICENSE` + `README.md` again (and `rc.d/isms`). Good catch on the licence-distribution angle.
4. Env-file privesc — fixed to `install -m 600 -o root -g wheel`. The `isms` user never needs to read it; `daemon(8)` inherits the environment from the root-side `start_precmd`.
5. Command paths (`isms server migrate`, `server serve --addr …`) fixed in the README and `isms.env.sample`.

**should-fix**
6. `daemon -f` discarding output — added `-S -T isms` so it forwards to syslog (which also makes the finding #2 fatal visible if misconfigured). Updated the "logs go to syslog" line to match.
7. `ISMS_DATABASE_URL` → `DATABASE_URL` in the rc.d header; added `ISMS_DATA_DIR` to the "required at minimum" note.

**nits**
8. Added `daemon -r` (restart on exit) — safe now that #1 is fixed so it won't loop on the help-printer.
9. Left the FreeBSD files landing in the other tarballs as-is per your offer; the config comment flags it.

Separately: this branch was carrying two already-merged #171 commits (patch-identical to master), so I rebased onto master — the PR now shows only the FreeBSD change.

unidoc-alip (approved):
looks good to me

PR #190, opened 2026-08-04
fix(web): store SSO hash token before the auth guard bounces to /login (#187)

### LINKED ISSUE #187: OIDC SSO login redirect loop  ,router guard redirects to /login before hash token is processed
### What happened?

After configuring an OIDC provider (Microsoft 365 / Entra ID) with Auto-add users enabled, the OIDC login flow completes successfully on the backend — a valid session JWT with all correct claims (email, name, user_id, org_id, role, org_slug) is generated every time. However, the browser never actually gets logged in. It ends up back on the generic /login page in a loop, with the valid token trapped inside a `redirect` query parameter instead of being processed. This makes SSO login completely unusable for end users, even though the OIDC handshake itself (IdP config, client credentials, JIT provisioning, org auto-add) works correctly.

### Steps to reproduce

1. Configure an OIDC provider for an org (tested with the "Microsoft 365" preset against Entra ID), with "Auto-add users" enabled
2. From the org login page, click "Sign in with Microsoft 365"
3. Complete authentication with the IdP (Entra ID) successfully
4. Observe the browser redirected to:
   https://<host>/login?redirect=/<org-slug>/login%23token=<JWT>&role=<role>
   instead of being logged into the app
5. Decoding the JWT confirms it is fully valid and correctly populated (correct email, name, user_id, org_id, role, org_slug)
6. Manually navigating to the "inner" URL (https://<host>/<org-slug>/login#token=<JWT>&role=<role>) reproduces the exact same redirect-to-/login loop — retrying does not help, this is fully deterministic

### Expected vs. actual

Expected: After a successful OIDC login, the SPA reads the token from the URL hash fragment, stores it, and the user lands on the org dashboard, logged in.

Actual: The Vue Router's global `beforeEach` guard (web/src/router.js) checks `getApiToken()` (which reads localStorage key `isms_api_token`) immediately on every navigation. Since the destination route (`/<org-slug>/login#token=...`) hasn't had a chance to run its own token-parsing logic yet, `getApiToken()` returns empty, and the guard immediately redirects to `/login?redirect=<original target>` — before the token in the hash is ever read or stored. This happens on every single OIDC login attempt, not just intermittently.

Confirmed workaround (proves this is purely a frontend routing bug, not a backend/auth issue): manually running this in the browser console before navigating fixes the login:
  localStorage.setItem('isms_api_token', '<token-from-url>')

This is the same class of bug already fixed for Cloudflare Access in PR #100 ("CF Access authenticated the API/CLI/git layer, but the web SPA gated on a local token and bounced to /login") — it looks like the same fix pattern was never applied to the regular OIDC login path.

### Component / surface

Web UI

### Version

0.7.0

### Anything else?

Self-hosted deployment, built from source per the Dockerfile in this repo. Fully reproducible in a fresh incognito browser session, every time, with a brand-new group-restricted Entra ID app registration (not a caching/stale-session issue).

Related: PR #109 / issue #108 (orgTokenRedirectURL subdomain vs path routing) is a different, already-fixed issue and does not address this — this is a client-side router guard ordering problem, not a URL-construction problem.

unidoc-ahall:
Closes #187

## What
OIDC (Entra ID / Microsoft 365) and any SSO that returns the session token in the URL hash was stuck in a redirect loop back to `/login`, with a valid JWT trapped in a `?redirect=` param — SSO login unusable, exactly as reported.

## Root cause
The OIDC callback lands the browser on `/<org>/login#token=<JWT>`. The router's global `beforeEach` guard runs on that navigation and checks `getApiToken()` (localStorage) **first** — which is empty, because the token is sitting in the URL hash and nothing has read it yet. The guard immediately redirects to `/login?redirect=…`. On top of that, in **path mode** `/<org>/login` isn't even a matched route (`orgScopedRoutes` has no `login` entry), so `Login.vue`'s hash-parsing (`onMounted`) never runs. Deterministic on every attempt; same class of bug fixed for Cloudflare Access in #100, never applied to the OIDC path.

## Fix
At the very top of `beforeEach` (before any auth gate), if the incoming URL hash contains `token=`, store the token and continue to the org's overview with the hash stripped (`/login`→`/overview`, `/<org>/login`→`/<org>/overview`). This runs before the redirect decision and works in both subdomain and path mode. Mirrors the CF Access handling from #100.

## Testing
`build-web` passes. The guard is interactive frontend with no JS test runner in the repo, so it's verified against the reporter's deterministic repro (both the OIDC flow and the manual `/<org>/login#token=` navigation now log in and land on the dashboard). Backend/auth were already correct (the JWT was valid) — this is purely the client-side guard ordering.

### DIFF
--- a/tests/poc_190_router_guard.mjs
+++ b/tests/poc_190_router_guard.mjs
@@ -1,0 +1,352 @@
+// PoC harness for PR #190 (web/src/router.js SSO hash-token guard).
+//
+// The routes table and the beforeEach body below are TRANSPLANTED VERBATIM from
+// web/src/router.js @ e57f8ca — only the Vue components are replaced by inert
+// markers and the api.js/localStorage deps are injected, so the control flow
+// under test is the real one. Run with:
+//   node poc190.mjs
+import {
+  createRouter,
+  createMemoryHistory,
+} from '../web/node_modules/vue-router/vue-router.node.mjs'
+
+// ---------------------------------------------------------------- fake api.js
+let store = {}
+const getApiToken = () => store['isms_api_token'] || ''
+const setApiToken = t => { store['isms_api_token'] = t }
+const clearApiToken = () => {
+  delete store['isms_api_token']
+  delete store['isms_user_email']
+  delete store['isms_user_name']
+}
+// getMe(): 'ok' → any token validates. '401' → only the freshly-minted SSO token
+// (NEWJWT) validates; the pre-existing local token is expired, so the server 401s
+// and api.js fetchRaw's 401 branch clears it (api.js:46-52).
+let getMeMode = 'ok'
+const api = {
+  async getMe() {
+    if (getMeMode === '401' && getApiToken() !== 'NEWJWT') {
+      clearApiToken()
+      const e = new Error('Authentication required'); e.status = 401; throw e
+    }
+    return { email: 'user@example.com', organization_slug: 'acme' }
+  },
+  async cfSession() { return null }, // not behind CF Access
+}
+
+// -------------------------------------------------- routes (verbatim, ll.35-100)
+const C = name => ({ __marker: name, render: () => null })
+const Login = C('Login')
+const Dashboard = C('Dashboard')
+const orgScopedRoutes = [
+  { path: '/overview', component: Dashboard },
+  { path: '/documents', component: C('Documents') },
+  { path: '/documents/:docId', component: C('Documents') },
+  { path: '/inbox', component: C('Inbox') },
+  { path: '/inbox/:tab?', component: C('Inbox') },
+  { path: '/risks', component: C('Risks') },
+  { path: '/risks/:id', component: C('Risks') },
+  { path: '/settings', component: C('Settings') },
+  { path: '/admin', component: C('Admin') },
+  { path: '/admin/:tab', component: C('Admin') },
+]
+function buildOrgRoutes(subdomainMode) {
+  return orgScopedRoutes.map(r => ({
+    path: subdomainMode ? r.path : '/:org' + r.path,
+    component: r.component,
+    meta: { orgScoped: true },
+  }))
+}
+function buildRoutes(subdomainMode, opts = {}) {
+  const extra = []
+  // Finding #2's proposed additions, under test in section E.
+  if (opts.orgLoginRoute && !subdomainMode) {
+    extra.push({ path: '/:org/login', component: Login, meta: { public: true } })
+  }
+  if (opts.catchAll) {
+    extra.push({ path: '/:pathMatch(.*)*', redirect: '/login' })
+  }
+  return [
+    ...extra,
+    { path: '/', component: C('Landing'), meta: { public: true } },
+    { path: '/login', component: Login, meta: { public: true } },
+    { path: '/signup', component: C('Signup'), meta: { public: true } },
+    { path: '/forgot-password', component: C('ForgotPassword'), meta: { public: true } },
+    { path: '/verify-email', component: C('VerifyEmail'), meta: { public: true } },
+    { path: '/verify-email-change', component: C('VerifyEmailChange'), meta: { public: true } },
+    { path: '/organizations', component: C('Organizations') },
+    ...buildOrgRoutes(subdomainMode),
+  ]
+}
+
+// ------------------------------------------- guard (verbatim, ll.113-184)
+// `variant`: 'head'  = PR #190 as merged  (gated on !getApiToken())
+//            'fixed' = review finding #1's proposed change (hash always wins)
+//            'base'  = master, before PR #190 (no hash block at all)
+function makeRouter({ subdomainMode, orgFromSubdomain, variant, routeOpts }) {
+  const router = createRouter({ history: createMemoryHistory(), routes: buildRoutes(subdomainMode, routeOpts) })
+  let sessionValidated = false
+  let cfTried = false
+
+  router.beforeEach(async (to, from) => {
+    if (variant === 'head') {
+      // ---- PR #190, web/src/router.js:114-127 (verbatim) ----
+      if (!getApiToken() && to.hash && to.hash.includes('token=')) {
+        const token = new URLSearchParams(to.hash.replace(/^#/, '')).get('token')
+        if (token) {
+          setApiToken(token)
+          return { path: to.path.replace(/\/login$/, '/overview') }
+        }
+      }
+    } else if (variant === 'fixed') {
+      // ---- proposed: drop the !getApiToken() gate, reset sessionValidated ----
+      if (to.hash && to.hash.includes('token=')) {
+        const token = new URLSearchParams(to.hash.replace(/^#/, '')).get('token')
+        if (token) {
+          setApiToken(token)
+          sessionValidated = false
+          return { path: to.path.replace(/\/login$/, '/overview') }
+        }
+      }
+    }
+
+    if (to.path === '/organizations' && orgFromSubdomain()) {
+      return getApiToken() ? { path: '/overview' } : { path: '/login' }
+    }
+    if (to.meta.public) {
+      if (to.path === '/' && getApiToken()) {
+        const slug = orgFromSubdomain()
+        if (slug) return { path: '/overview' }
+        return { path: '/organizations' }
+      }
+      return true
+    }
+    if (!getApiToken()) {
+      if (!cfTried) {
+        cfTried = true
+        const cf = await api.cfSession()
+        if (cf?.token) {
+          setApiToken(cf.token)
+          if (cf.email) store['isms_user_email'] = cf.email
+          if (cf.name) store['isms_user_name'] = cf.name
+        }
+      }
+      if (!getApiToken()) {
+        return { path: '/login', query: to.fullPath !== '/' ? { redirect: to.fullPath } : undefined }
+      }
+    }
+    if (!sessionValidated) {
+      try {
+        const me = await api.getMe()
+        if (me?.email) { sessionValidated = true; return true }
+      } catch (e) {
+        if (e?.status !== 401 && e?.isNetwork) return true
+      }
+      clearApiToken()
+      sessionValidated = false
+      return { path: '/login', query: to.fullPath !== '/' ? { redirect: to.fullPath } : undefined }
+    }
+    return true
+  })
+  return router
+}
+
+// -------------------------------------------------------------- Login.vue stub
+// Mirrors Login.vue onMounted (ll.230-248): if the mounted route IS Login and a
+// hash token is present, it stores the token and routes to the org overview.
+// This is the safety net that exists in subdomain mode and NOT in path mode.
+async function mountIfLogin(router) {
+  const cur = router.currentRoute.value
+  const isLogin = cur.matched.some(m => m.components?.default?.__marker === 'Login')
+  if (!isLogin) return { mounted: cur.matched.length ? 'other' : 'NOTHING (blank page)' }
+  // Login.vue reads window.location.hash — i.e. the hash of where we LANDED, not
+  // of the original SSO URL. A token folded into ?redirect= is invisible to it.
+  const hash = cur.hash
+  const tok = hash ? new URLSearchParams(hash.replace(/^#/, '')).get('token') : null
+  if (tok) {
+    setApiToken(tok)
+    await router.push(cur.params.org ? `/${cur.params.org}/overview` : '/overview')
+    return { mounted: 'Login → handled hash' }
+  }
+  return { mounted: 'Login (form)' }
+}
+
+// ------------------------------------------------------------------- scenarios
+let pass = 0, fail = 0
+function check(label, got, want) {
+  const ok = got === want
+  ok ? pass++ : fail++
+  console.log(`   ${ok ? 'PASS' : 'FAIL'}  ${label}\n         got:  ${got}\n         want: ${want}`)
+}
+
+async function run({ title, subdomainMode, url, preToken, getMe = 'ok', variant, routeOpts, expect }) {
+  store = {}
+  if (preToken) store['isms_api_token'] = preToken
+  getMeMode = getMe
+  const router = makeRouter({
+    subdomainMode,
+    orgFromSubdomain: () => (subdomainMode ? 'acme' : null),
+    variant,
+    routeOpts,
+  })
+  // The SSO callback is a full browser navigation: this IS the initial nav.
+  try { await router.push(url) } catch (e) { /* navigation failure */ }
+  const view = await mountIfLogin(router)   // then the destination component mounts
+  const landed = router.currentRoute.value  // read the URL AFTER the view has run
+  const outcome = `${landed.fullPath} | view=${view.mounted} | token=${getApiToken() || '(none)'}`
+  const ro = routeOpts ? `, routes+=${Object.keys(routeOpts).join('+')}` : ''
+  console.log(`\n${title}\n   push: ${url}  [${variant}, ${subdomainMode ? 'subdomain' : 'path'} mode, preToken=${preToken || 'none'}, getMe=${getMe}${ro}]`)
+  check(expect.label, outcome, expect.outcome)
+  return outcome
+}
+
+const JWT = 'NEWJWT'
+
+console.log('='.repeat(78))
+console.log('A. Fresh session — the flow PR #190 targets (must work)')
+console.log('='.repeat(78))
+await run({
+  title: 'A1 path mode, /<org>/login#token= — BASE (master, pre-PR)', variant: 'base',
+  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`,
+  expect: { label: '#187 reproduces on master', outcome: '/login?redirect=/acme/login%23token=NEWJWT%26role=reader | view=Login (form) | token=(none)' },
+})
+await run({
+  title: 'A2 path mode, /<org>/login#token= — HEAD (PR #190)', variant: 'head',
+  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`,
+  expect: { label: 'PR fixes the reported flow', outcome: '/acme/overview | view=other | token=NEWJWT' },
+})
+await run({
+  title: 'A3 subdomain mode, /login#token= — HEAD', variant: 'head',
+  subdomainMode: true, url: `/login#token=${JWT}&role=reader`,
+  expect: { label: 'subdomain mode works', outcome: '/overview | view=other | token=NEWJWT' },
+})
+await run({
+  title: 'A4 apex fallback shape /#token= (api_oidc.go:351) — HEAD', variant: 'head',
+  subdomainMode: false, url: `/#token=${JWT}&role=reader`,
+  expect: { label: 'lands on org picker, not a dashboard', outcome: '/organizations | view=other | token=NEWJWT' },
+})
+
+console.log('\n' + '='.repeat(78))
+console.log('B. Finding #1 — a token already in localStorage')
+console.log('='.repeat(78))
+await run({
+  title: 'B1 path mode, live session, /<org>/login#token= — HEAD', variant: 'head',
+  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, preToken: 'OLDJWT', getMe: 'ok',
+  expect: { label: 'blank page, fresh token dropped', outcome: '/acme/login#token=NEWJWT&role=reader | view=NOTHING (blank page) | token=OLDJWT' },
+})
+await run({
+  title: 'B2 path mode, stale/expired token (getMe 401) — HEAD', variant: 'head',
+  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, preToken: 'STALEJWT', getMe: '401',
+  expect: { label: '#187 loop returns, token trapped in ?redirect=', outcome: '/login?redirect=/acme/login%23token=NEWJWT%26role=reader | view=Login (form) | token=(none)' },
+})
+await run({
+  title: 'B3 subdomain mode, live session — HEAD (Login.vue is the safety net)', variant: 'head',
+  subdomainMode: true, url: `/login#token=${JWT}&role=reader`, preToken: 'OLDJWT', getMe: 'ok',
+  expect: { label: 'subdomain mode recovers via Login.vue', outcome: '/overview | view=Login → handled hash | token=NEWJWT' },
+})
+await run({
+  title: 'B4 path mode, live session — FIXED variant', variant: 'fixed',
+  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, preToken: 'OLDJWT', getMe: 'ok',
+  expect: { label: 'proposed fix resolves it', outcome: '/acme/overview | view=other | token=NEWJWT' },
+})
+await run({
+  title: 'B5 path mode, stale token — FIXED variant', variant: 'fixed',
+  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, preToken: 'STALEJWT', getMe: '401',
+  expect: { label: 'fix survives an expired local token too', outcome: '/acme/overview | view=other | token=NEWJWT' },
+})
+await run({
+  title: 'B6 fresh session, FIXED variant — no regression vs HEAD', variant: 'fixed',
+  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`,
+  expect: { label: 'fix keeps the happy path', outcome: '/acme/overview | view=other | token=NEWJWT' },
+})
+
+console.log('\n' + '='.repeat(78))
+console.log('C. Finding #2 — unmatched org-scoped paths in path mode')
+console.log('='.repeat(78))
+{
+  const r = makeRouter({ subdomainMode: false, orgFromSubdomain: () => null, variant: 'head' })
+  check('/acme/login matches no route in path mode', String(r.resolve('/acme/login').matched.length), '0')
+  check('/login matches Login at top level', String(r.resolve('/login').matched.length), '1')
+  check('/acme/typo (any unmatched org path) matches nothing', String(r.resolve('/acme/typo').matched.length), '0')
+  const rs = makeRouter({ subdomainMode: true, orgFromSubdomain: () => 'acme', variant: 'head' })
+  check('/login matches Login in subdomain mode too', String(rs.resolve('/login').matched.length), '1')
+}
+await run({
+  title: 'C1 path mode, unmatched org path with a valid session — HEAD', variant: 'head',
+  subdomainMode: false, url: '/acme/typo', preToken: 'OLDJWT', getMe: 'ok',
+  expect: { label: 'no catch-all → blank page', outcome: '/acme/typo | view=NOTHING (blank page) | token=OLDJWT' },
+})
+
+console.log('\n' + '='.repeat(78))
+console.log('D. Findings #3 / #4 — identity keys, and hash accepted on any path')
+console.log('='.repeat(78))
+await run({
+  title: 'D1 identity keys after a hash login — HEAD', variant: 'head',
+  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`,
+  expect: { label: 'only the token is stored', outcome: '/acme/overview | view=other | token=NEWJWT' },
+})
+check('finding #3: isms_user_email set by the hash path?', String(store['isms_user_email']), 'undefined')
+check('finding #3: isms_user_name set by the hash path?', String(store['isms_user_name']), 'undefined')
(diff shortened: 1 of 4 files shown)


unidoc-alip (changes requested):
The diagnosis in the description is accurate and it holds up against the code: `beforeEach` gated on `getApiToken()` before anything read the hash, and `orgScopedRoutes` really has no `login` entry (nor any catch-all — no `addRoute` anywhere in `web/src`), so in path mode `/<org>/login#token=` matches no route and `Login.vue`'s `onMounted` hash parser never runs. Handling the hash at the top of the guard is the right fix for the reported flow, and it works in all three redirect shapes `orgTokenRedirectURL` produces: subdomain `/login#token=`, path `/<org>/login#token=`, and the malformed-base `/#token=` fallback (`api_oidc.go:351`), which lands on `/` and then bounces to `/organizations`.

One gap blocks it. The new block is gated on `!getApiToken()`, so a user who already has a session and lands on `/<org>/login#token=NEW` in path mode has the fresh token silently discarded — the unmatched route renders the old org's app shell around an empty content area and they stay on the old identity. That is the reporter's own step-6 repro for anyone currently logged in. Dropping the `!getApiToken()` precondition fixes it: a hash token is a fresh server-minted assertion and should always win over older localStorage state.

Every finding below is backed by an executable PoC — `tests/poc_190_router_guard.mjs`, 30 assertions, all passing — that transplants the real routes table and the real `beforeEach` body into `vue-router` under a memory history and replays master, this PR, and the proposed fix side by side. Details in "Verification" at the end.

Two corrections to the description. First, this does not fully mirror #100: #100's CF probe sits *inside* the `if (!getApiToken())` branch and also persists `isms_user_email`/`isms_user_name`. The placement is the same shape, but a CF probe is idempotent and re-runnable while a hash token is one-shot, so inheriting #100's no-token gate is exactly what causes the gap above. Second, `build-web` is not CI-verified — `.github/workflows/tests.yml` runs only `go-unit` and `integration`, and those are the only two checks on this PR, so that was a local build.

---

### 1. [must-fix] `web/src/router.js` line 120

The `!getApiToken()` precondition makes this fix incomplete: when a token is already in localStorage, the fresh SSO token in the hash is dropped.

In **path mode** there is no fallback, because `/<org>/login` matches no route and there is no catch-all. A user with a live session — already logged into org A, or performing the reporter's step-6 manual navigation to `/<org>/login#token=…` — hits this: the block is skipped, `to.meta.public` is falsy (`to.matched` is empty), the auth branch sees a valid token, `getMe()` succeeds, the guard returns `true`. The fresh token is silently discarded, the destination never loads, and the user stays on the previous session's identity.

Concretely: `App.vue`'s `isPublicRoute` (`App.vue:567-574`) is a hardcoded path list that does not include `/<org>/login`, so the app-layout branch renders — sidebar and header for the *old* org around an empty `<main><router-view/></main>` — and `loadAppData()` runs against the old token (`route.params.org` is empty on an unmatched route, so neither the org-mismatch redirect nor `switch-org` in `loadUserData` fires).

Reproduced with the real routes table and the real guard body under `vue-router`:

```
A2  fresh session, path mode, this PR
    -> /acme/overview | view=other | token=NEWJWT              (PR works)
B1  live session (OLDJWT valid), path mode, this PR
    -> /acme/login#token=NEWJWT | nothing rendered | token=OLDJWT (new one lost)
B2  stale token (getMe 401), path mode, this PR
    -> /login?redirect=/acme/login%23token=NEWJWT | token=(none)   (#187 again)
B3  live session, SUBDOMAIN mode, this PR
    -> /overview | view=Login handled the hash | token=NEWJWT   (Login.vue saves it)
B4  live session, path mode, proposed fix   -> /acme/overview | token=NEWJWT
B5  stale token,  path mode, proposed fix   -> /acme/overview | token=NEWJWT
B6  fresh session, proposed fix             -> /acme/overview | token=NEWJWT
```

B3 versus B1 is the whole story: subdomain mode has `Login.vue` as a safety net, path mode has nothing.

Suggested fix — drop the precondition and let the hash win unconditionally:

```js
if (to.hash && to.hash.includes('token=')) {
  const token = new URLSearchParams(to.hash.replace(/^#/, '')).get('token')
  if (token) {
    setApiToken(token)
    sessionValidated = false   // new identity — force revalidation
    return { path: to.path.replace(/\/login$/, '/overview') }
  }
}
```

The `sessionValidated = false` matters once the gate is removed: overwriting the token for a different identity while a session was already validated in this page load would otherwise skip the `getMe()` check. B4/B5/B6 above run this exact variant. Cross-org is safe too — `useSession.loadUserData` already swaps the token via `/auth/switch-org` when the path slug and the JWT's org disagree (`useSession.js:73-82`).

On the B2 line above, for completeness: the expired-token variant is the weaker of the two, since `api.js` `fetchRaw` clears on every 401 and `Login.vue`'s `onMounted` calls `getMe()` whenever a token exists, so a stale token is normally gone before the OIDC hop. The live-session case in B1 needs no such precondition and carries the finding on its own.

---

### 2. [should-fix] `web/src/router.js` — routes table (not in this diff)

The guard now works around the fact that `/<org>/login` isn't a registered route in path mode, rather than fixing it. That gap has other consequences: any org-scoped path that doesn't match — including `/<org>/login` reached any other way, e.g. a bookmark or a stale link — renders the app shell around an empty content area instead of redirecting, since `routes` has no catch-all.

Verified with the same harness: `/acme/login` resolves to 0 matched records, `/acme/typo` resolves to 0 matched, and with a valid session the guard returns `true` on that unmatched route so nothing takes over. `/login` matches in both modes.

I tested the suggested remedy rather than guessing at it (9 assertions):

```js
// web/src/router.js — add to `routes`, path mode only for the first entry
{ path: '/:org/login', component: Login, meta: { public: true } },
{ path: '/:pathMatch(.*)*', redirect: '/login' },   // no more dead-end routes
```

Results, with the guard left exactly as this PR has it:

```
E1  live session, /acme/login#token=  -> /acme/overview via Login.vue's onMounted
E2  /acme/typo                        -> /login (instead of a dead end)
E3  /acme/risks/5                     -> Risks (catch-all does not shadow)
E4  /acme/overview                    -> Dashboard
E5  fresh session hash login          -> guard's fast path still wins
E6  routes + finding #1's guard fix   -> coexist cleanly
```

`/organizations` and `/signup` still resolve to their own components, so the catch-all does not shadow the top-level public routes either.

Please still fix finding #1 on this PR — it's the smaller change and it covers the hash shapes that don't end in `/login`. Treat this section as the follow-up that removes the dead-end-route class of failure generally.

---

### 3. [nit] `web/src/router.js` line 123

#100's CF handling also persisted `isms_user_email` / `isms_user_name` alongside the token; this path stores only the token (the PoC asserts both keys stay unset). So after any SSO login `getCurrentUser()` — which reads `localStorage.isms_user_email` — returns `''`, and roughly 15 call sites pass it as `author` / `resolved_by` / `created_by` / `requested_by`.

Not a regression (`Login.vue`'s OIDC path never stored them either) and not user-visible: the server overrides it — `api_collab.go:1706` has `comment.Author = getUserEmail(c) // always use authenticated user`, and `server.go:1448-1450` falls back to the JWT actor when the field is blank. The header reads `useSession`'s reactive `user`, which `loadUserData` fills from `/me`.

If you want the localStorage copy to be right for SSO too, the clean place is `useSession.loadUserData` (`useSession.js:85-87`) — persist `me.email` / `me.name` there once, and every login path benefits without the token round-tripping identity through the URL fragment.

Nothing reads a stored role (`grep -rn isms_user_role web/src` has no hits), so discarding `role=` from the hash is not an issue.

---

### 4. [nit] `web/src/router.js` line 120

Minor surface widening: any URL with `#token=…` on *any* path now sets the app's session token, where previously only `/login` (via `Login.vue`) did. The session-fixation-by-link pattern is pre-existing, but it's now reachable everywhere. The PoC confirms the delta — on master `/acme/risks#token=X` leaves the token unset; with this PR the same URL stores it.

Noting it for the record only, no change requested: a token planted this way is a token the attacker already holds, and narrowing the guard to `/login`-suffixed paths would conflict with finding 1's "hash always wins".

---

### Verification

`tests/poc_190_router_guard.mjs` — run with `node tests/poc_190_router_guard.mjs`, 30 assertions, all passing.

The repo has no JS test runner, so the harness runs on bare Node against `web/node_modules/vue-router` (4.6.4) with a memory history. The `routes` table and the whole `beforeEach` body are transplanted verbatim from `web/src/router.js` at `e57f8ca`; only the Vue components are inert markers and `api.js`/`localStorage` are injected fakes. Three guard variants run side by side — master, this PR, and finding #1's change — plus a `Login.vue` stub mirroring its `onMounted` hash handling, so the path-mode versus subdomain-mode asymmetry is directly observable. Section A covers the fresh-session flows for all three redirect shapes, B covers finding 1, C covers finding 2, D covers findings 3 and 4, and E covers finding 2's proposed routes.

Caveats, stated plainly: the fake `getMe` treats the freshly-minted token as valid and any other token as expired (mirroring `api.js` `fetchRaw`'s 401 then `clearApiToken`); the `Login.vue` stub is a faithful reduction of `onMounted`, not the real component; `createMemoryHistory` stands in for `createWebHistory`, which does not affect guard ordering or route matching; and the harness has no `App.vue`, so it reports an unmatched route as "nothing rendered" — what the real app does with that route was established by reading `App.vue:4-9,567-574` and `useSession.loadUserData`, not by the PoC.

Where each finding landed:

- Finding 1 — confirmed (B1 token dropped and the destination never loads; B2 the #187 loop), and the proposed fix validated (B4, B5, B6).
- Finding 2 — confirmed (0 matched records, guard passes through), and the suggested routes validated (E1 through E6, no shadowing).
- Finding 3 — confirmed but benign; the keys are unset, but the server overrides the author (`api_collab.go:1706`, `server.go:1448`).
- Finding 4 — confirmed as a real delta against master.

---

### **tests/poc_190_router_guard.mjs**

```
// PoC harness for PR #190 (web/src/router.js SSO hash-token guard).
//
// The routes table and the beforeEach body below are TRANSPLANTED VERBATIM from
// web/src/router.js @ e57f8ca — only the Vue components are replaced by inert
// markers and the api.js/localStorage deps are injected, so the control flow
// under test is the real one. Run with:
//   node poc190.mjs
import {
  createRouter,
  createMemoryHistory,
} from '../web/node_modules/vue-router/vue-router.node.mjs'

// ---------------------------------------------------------------- fake api.js
let store = {}
const getApiToken = () => store['isms_api_token'] || ''
const setApiToken = t => { store['isms_api_token'] = t }
const clearApiToken = () => {
  delete store['isms_api_token']
  delete store['isms_user_email']
  delete store['isms_user_name']
}
// getMe(): 'ok' → any token validates. '401' → only the freshly-minted SSO token
// (NEWJWT) validates; the pre-existing local token is expired, so the server 401s
// and api.js fetchRaw's 401 branch clears it (api.js:46-52).
let getMeMode = 'ok'
const api = {
  async getMe() {
    if (getMeMode === '401' && getApiToken() !== 'NEWJWT') {
      clearApiToken()
      const e = new Error('Authentication required'); e.status = 401; throw e
    }
    return { email: 'user@example.com', organization_slug: 'acme' }
  },
  async cfSession() { return null }, // not behind CF Access
}

// -------------------------------------------------- routes (verbatim, ll.35-100)
const C = name => ({ __marker: name, render: () => null })
const Login = C('Login')
const Dashboard = C('Dashboard')
const orgScopedRoutes = [
  { path: '/overview', component: Dashboard },
  { path: '/documents', component: C('Documents') },
  { path: '/documents/:docId', component: C('Documents') },
  { path: '/inbox', component: C('Inbox') },
  { path: '/inbox/:tab?', component: C('Inbox') },
  { path: '/risks', component: C('Risks') },
  { path: '/risks/:id', component: C('Risks') },
  { path: '/settings', component: C('Settings') },
  { path: '/admin', component: C('Admin') },
  { path: '/admin/:tab', component: C('Admin') },
]
function buildOrgRoutes(subdomainMode) {
  return orgScopedRoutes.map(r => ({
    path: subdomainMode ? r.path : '/:org' + r.path,
    component: r.component,
    meta: { orgScoped: true },
  }))
}
function buildRoutes(subdomainMode, opts = {}) {
  const extra = []
  // Finding #2's proposed additions, under test in section E.
  if (opts.orgLoginRoute && !subdomainMode) {
    extra.push({ path: '/:org/login', component: Login, meta: { public: true } })
  }
  if (opts.catchAll) {
    extra.push({ path: '/:pathMatch(.*)*', redirect: '/login' })
  }
  return [
    ...extra,
    { path: '/', component: C('Landing'), meta: { public: true } },
    { path: '/login', component: Login, meta: { public: true } },
    { path: '/signup', component: C('Signup'), meta: { public: true } },
    { path: '/forgot-password', component: C('ForgotPassword'), meta: { public: true } },
    { path: '/verify-email', component: C('VerifyEmail'), meta: { public: true } },
    { path: '/verify-email-change', component: C('VerifyEmailChange'), meta: { public: true } },
    { path: '/organizations', component: C('Organizations') },
    ...buildOrgRoutes(subdomainMode),
  ]
}

// ------------------------------------------- guard (verbatim, ll.113-184)
// `variant`: 'head'  = PR #190 as merged  (gated on !getApiToken())
//            'fixed' = review finding #1's proposed change (hash always wins)
//            'base'  = master, before PR #190 (no hash block at all)
function makeRouter({ subdomainMode, orgFromSubdomain, variant, routeOpts }) {
  const router = createRouter({ history: createMemoryHistory(), routes: buildRoutes(subdomainMode, routeOpts) })
  let sessionValidated = false
  let cfTried = false

  router.beforeEach(async (to, from) => {
    if (variant === 'head') {
      // ---- PR #190, web/src/router.js:114-127 (verbatim) ----
      if (!getApiToken() && to.hash && to.hash.includes('token=')) {
        const token = new URLSearchParams(to.hash.replace(/^#/, '')).get('token')
        if (token) {
          setApiToken(token)
          return { path: to.path.replace(/\/login$/, '/overview') }
        }
      }
    } else if (variant === 'fixed') {
      // ---- proposed: drop the !getApiToken() gate, reset sessionValidated ----
      if (to.hash && to.hash.includes('token=')) {
        const token = new URLSearchParams(to.hash.replace(/^#/, '')).get('token')
        if (token) {
          setApiToken(token)
          sessionValidated = false
          return { path: to.path.replace(/\/login$/, '/overview') }
        }
      }
    }

    if (to.path === '/organizations' && orgFromSubdomain()) {
      return getApiToken() ? { path: '/overview' } : { path: '/login' }
    }
    if (to.meta.public) {
      if (to.path === '/' && getApiToken()) {
        const slug = orgFromSubdomain()
        if (slug) return { path: '/overview' }
        return { path: '/organizations' }
      }
      return true
    }
    if (!getApiToken()) {
      if (!cfTried) {
        cfTried = true
        const cf = await api.cfSession()
        if (cf?.token) {
          setApiToken(cf.token)
          if (cf.email) store['isms_user_email'] = cf.email
          if (cf.name) store['isms_user_name'] = cf.name
        }
      }
      if (!getApiToken()) {
        return { path: '/login', query: to.fullPath !== '/' ? { redirect: to.fullPath } : undefined }
      }
    }
    if (!sessionValidated) {
      try {
        const me = await api.getMe()
        if (me?.email) { sessionValidated = true; return true }
      } catch (e) {
        if (e?.status !== 401 && e?.isNetwork) return true
      }
      clearApiToken()
      sessionValidated = false
      return { path: '/login', query: to.fullPath !== '/' ? { redirect: to.fullPath } : undefined }
    }
    return true
  })
  return router
}

// -------------------------------------------------------------- Login.vue stub
// Mirrors Login.vue onMounted (ll.230-248): if the mounted route IS Login and a
// hash token is present, it stores the token and routes to the org overview.
// This is the safety net that exists in subdomain mode and NOT in path mode.
async function mountIfLogin(router) {
  const cur = router.currentRoute.value
  const isLogin = cur.matched.some(m => m.components?.default?.__marker === 'Login')
  if (!isLogin) return { mounted: cur.matched.length ? 'other' : 'NOTHING (blank page)' }
  // Login.vue reads window.location.hash — i.e. the hash of where we LANDED, not
  // of the original SSO URL. A token folded into ?redirect= is invisible to it.
  const hash = cur.hash
  const tok = hash ? new URLSearchParams(hash.replace(/^#/, '')).get('token') : null
  if (tok) {
    setApiToken(tok)
    await router.push(cur.params.org ? `/${cur.params.org}/overview` : '/overview')
    return { mounted: 'Login → handled hash' }
  }
  return { mounted: 'Login (form)' }
}

// ------------------------------------------------------------------- scenarios
let pass = 0, fail = 0
function check(label, got, want) {
  const ok = got === want
  ok ? pass++ : fail++
  console.log(`   ${ok ? 'PASS' : 'FAIL'}  ${label}\n         got:  ${got}\n         want: ${want}`)
}

async function run({ title, subdomainMode, url, preToken, getMe = 'ok', variant, routeOpts, expect }) {
  store = {}
  if (preToken) store['isms_api_token'] = preToken
  getMeMode = getMe
  const router = makeRouter({
    subdomainMode,
    orgFromSubdomain: () => (subdomainMode ? 'acme' : null),
    variant,
    routeOpts,
  })
  // The SSO callback is a full browser navigation: this IS the initial nav.
  try { await router.push(url) } catch (e) { /* navigation failure */ }
  const view = await mountIfLogin(router)   // then the destination component mounts
  const landed = router.currentRoute.value  // read the URL AFTER the view has run
  const outcome = `${landed.fullPath} | view=${view.mounted} | token=${getApiToken() || '(none)'}`
  const ro = routeOpts ? `, routes+=${Object.keys(routeOpts).join('+')}` : ''
  console.log(`\n${title}\n   push: ${url}  [${variant}, ${subdomainMode ? 'subdomain' : 'path'} mode, preToken=${preToken || 'none'}, getMe=${getMe}${ro}]`)
  check(expect.label, outcome, expect.outcome)
  return outcome
}

const JWT = 'NEWJWT'

console.log('='.repeat(78))
console.log('A. Fresh session — the flow PR #190 targets (must work)')
console.log('='.repeat(78))
await run({
  title: 'A1 path mode, /<org>/login#token= — BASE (master, pre-PR)', variant: 'base',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`,
  expect: { label: '#187 reproduces on master', outcome: '/login?redirect=/acme/login%23token=NEWJWT%26role=reader | view=Login (form) | token=(none)' },
})
await run({
  title: 'A2 path mode, /<org>/login#token= — HEAD (PR #190)', variant: 'head',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`,
  expect: { label: 'PR fixes the reported flow', outcome: '/acme/overview | view=other | token=NEWJWT' },
})
await run({
  title: 'A3 subdomain mode, /login#token= — HEAD', variant: 'head',
  subdomainMode: true, url: `/login#token=${JWT}&role=reader`,
  expect: { label: 'subdomain mode works', outcome: '/overview | view=other | token=NEWJWT' },
})
await run({
  title: 'A4 apex fallback shape /#token= (api_oidc.go:351) — HEAD', variant: 'head',
  subdomainMode: false, url: `/#token=${JWT}&role=reader`,
  expect: { label: 'lands on org picker, not a dashboard', outcome: '/organizations | view=other | token=NEWJWT' },
})

console.log('\n' + '='.repeat(78))
console.log('B. Finding #1 — a token already in localStorage')
console.log('='.repeat(78))
await run({
  title: 'B1 path mode, live session, /<org>/login#token= — HEAD', variant: 'head',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, preToken: 'OLDJWT', getMe: 'ok',
  expect: { label: 'blank page, fresh token dropped', outcome: '/acme/login#token=NEWJWT&role=reader | view=NOTHING (blank page) | token=OLDJWT' },
})
await run({
  title: 'B2 path mode, stale/expired token (getMe 401) — HEAD', variant: 'head',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, preToken: 'STALEJWT', getMe: '401',
  expect: { label: '#187 loop returns, token trapped in ?redirect=', outcome: '/login?redirect=/acme/login%23token=NEWJWT%26role=reader | view=Login (form) | token=(none)' },
})
await run({
  title: 'B3 subdomain mode, live session — HEAD (Login.vue is the safety net)', variant: 'head',
  subdomainMode: true, url: `/login#token=${JWT}&role=reader`, preToken: 'OLDJWT', getMe: 'ok',
  expect: { label: 'subdomain mode recovers via Login.vue', outcome: '/overview | view=Login → handled hash | token=NEWJWT' },
})
await run({
  title: 'B4 path mode, live session — FIXED variant', variant: 'fixed',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, preToken: 'OLDJWT', getMe: 'ok',
  expect: { label: 'proposed fix resolves it', outcome: '/acme/overview | view=other | token=NEWJWT' },
})
await run({
  title: 'B5 path mode, stale token — FIXED variant', variant: 'fixed',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, preToken: 'STALEJWT', getMe: '401',
  expect: { label: 'fix survives an expired local token too', outcome: '/acme/overview | view=other | token=NEWJWT' },
})
await run({
  title: 'B6 fresh session, FIXED variant — no regression vs HEAD', variant: 'fixed',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`,
  expect: { label: 'fix keeps the happy path', outcome: '/acme/overview | view=other | token=NEWJWT' },
})

console.log('\n' + '='.repeat(78))
console.log('C. Finding #2 — unmatched org-scoped paths in path mode')
console.log('='.repeat(78))
{
  const r = makeRouter({ subdomainMode: false, orgFromSubdomain: () => null, variant: 'head' })
  check('/acme/login matches no route in path mode', String(r.resolve('/acme/login').matched.length), '0')
  check('/login matches Login at top level', String(r.resolve('/login').matched.length), '1')
  check('/acme/typo (any unmatched org path) matches nothing', String(r.resolve('/acme/typo').matched.length), '0')
  const rs = makeRouter({ subdomainMode: true, orgFromSubdomain: () => 'acme', variant: 'head' })
  check('/login matches Login in subdomain mode too', String(rs.resolve('/login').matched.length), '1')
}
await run({
  title: 'C1 path mode, unmatched org path with a valid session — HEAD', variant: 'head',
  subdomainMode: false, url: '/acme/typo', preToken: 'OLDJWT', getMe: 'ok',
  expect: { label: 'no catch-all → blank page', outcome: '/acme/typo | view=NOTHING (blank page) | token=OLDJWT' },
})

console.log('\n' + '='.repeat(78))
console.log('D. Findings #3 / #4 — identity keys, and hash accepted on any path')
console.log('='.repeat(78))
await run({
  title: 'D1 identity keys after a hash login — HEAD', variant: 'head',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`,
  expect: { label: 'only the token is stored', outcome: '/acme/overview | view=other | token=NEWJWT' },
})
check('finding #3: isms_user_email set by the hash path?', String(store['isms_user_email']), 'undefined')
check('finding #3: isms_user_name set by the hash path?', String(store['isms_user_name']), 'undefined')
check('finding #3: role= from the hash persisted anywhere?',
  String(Object.keys(store).filter(k => k !== 'isms_api_token').length), '0')

await run({
  title: 'D2 #token= on a non-login route — HEAD', variant: 'head',
  subdomainMode: false, url: `/acme/risks#token=${JWT}`,
  expect: { label: 'token accepted on any path (was /login only)', outcome: '/acme/risks | view=other | token=NEWJWT' },
})
{
  const r = makeRouter({ subdomainMode: false, orgFromSubdomain: () => null, variant: 'base' })
  store = {}
  await r.push('/'); store = {}
  try { await r.push(`/acme/risks#token=${JWT}`) } catch {}
  check('finding #4 baseline: master ignores #token= off /login', getApiToken() || '(none)', '(none)')
}

console.log('\n' + '='.repeat(78))
console.log("E. Finding #2's SUGGESTED fix under test — /:org/login route + catch-all")
console.log('='.repeat(78))
const R2 = { orgLoginRoute: true, catchAll: true }
await run({
  title: 'E1 live session + /<org>/login registered — guard UNCHANGED (head)', variant: 'head',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, preToken: 'OLDJWT', getMe: 'ok', routeOpts: R2,
  expect: { label: 'route alone is a sufficient safety net for finding #1', outcome: '/acme/overview | view=Login → handled hash | token=NEWJWT' },
})
await run({
  title: 'E2 unmatched org path with catch-all — head', variant: 'head',
  subdomainMode: false, url: '/acme/typo', preToken: 'OLDJWT', getMe: 'ok', routeOpts: R2,
  expect: { label: 'catch-all replaces the blank page', outcome: '/login | view=Login (form) | token=OLDJWT' },
})
await run({
  title: 'E3 existing org route still matches (no shadowing) — head', variant: 'head',
  subdomainMode: false, url: '/acme/risks/5', preToken: 'OLDJWT', getMe: 'ok', routeOpts: R2,
  expect: { label: 'catch-all does not shadow /:org/risks/:id', outcome: '/acme/risks/5 | view=other | token=OLDJWT' },
})
await run({
  title: 'E4 overview still matches — head', variant: 'head',
  subdomainMode: false, url: '/acme/overview', preToken: 'OLDJWT', getMe: 'ok', routeOpts: R2,
  expect: { label: 'no regression on the main landing route', outcome: '/acme/overview | view=other | token=OLDJWT' },
})
await run({
  title: 'E5 fresh session, guard fast path still wins — head', variant: 'head',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, routeOpts: R2,
  expect: { label: 'guard handles it before Login.vue mounts', outcome: '/acme/overview | view=other | token=NEWJWT' },
})
await run({
  title: 'E6 route + catch-all combined with the FIXED guard', variant: 'fixed',
  subdomainMode: false, url: `/acme/login#token=${JWT}&role=reader`, preToken: 'OLDJWT', getMe: 'ok', routeOpts: R2,
  expect: { label: 'belt-and-braces: both fixes coexist', outcome: '/acme/overview | view=other | token=NEWJWT' },
})
{
  const r = makeRouter({ subdomainMode: false, orgFromSubdomain: () => null, variant: 'head', routeOpts: R2 })
  check('E: /acme/login now matches Login', String(r.resolve('/acme/login').matched.length), '1')
  check('E: /organizations still matches Organizations',
    String(r.resolve('/organizations').matched[0]?.components?.default?.__marker), 'Organizations')
  check('E: /signup still matches Signup',
    String(r.resolve('/signup').matched[0]?.components?.default?.__marker), 'Signup')
}

console.log('\n' + '='.repeat(78))
console.log(`RESULT: ${pass} pass, ${fail} fail`)
console.log('='.repeat(78))
process.exit(fail ? 1 : 0)
```

unidoc-ahall:
Thanks — genuinely thorough, and the PoC made verification easy. I reconstructed and ran `tests/poc_190_router_guard.mjs` here (30/30), so every call below is against your harness. I folded all four findings into this PR rather than splitting #2/#3 into a follow-up — they're small and you'd already tested them, so a separate issue + review cycle was more overhead than the change.

**1. must-fix — applied.** You're right the `!getApiToken()` gate dropped the fresh token for anyone already logged in (B1: blank page, `OLDJWT` kept) and re-opened #187 on a stale token (B2). Dropped the precondition so the hash wins unconditionally, plus `sessionValidated = false` so the new identity revalidates via `getMe`. That's your `fixed` variant — B4/B5/B6 green, A2/A3/A4 unchanged.

**2. should-fix — applied (not deferred).** Registered `/:org/login` as a public route in path mode and a `/:pathMatch(.*)*` catch-all → `/login`. Your E-section is the acceptance test: E2 `/<org>/typo` → `/login` instead of a blank page, E3/E4 confirm no shadowing of `/:org/risks/:id` or `/:org/overview`, and `/organizations` + `/signup` still resolve. E6 confirms the catch-all + `/:org/login` coexist with the finding-1 guard fix.

**3. nit — applied.** `useSession.loadUserData` now persists `isms_user_email`/`name` via an exported `setCurrentUser`, so every login path (SSO included) fills the synchronous `getCurrentUser()` reads once, without round-tripping identity through the URL fragment. It was benign anyway (server overrides `author` at `api_collab.go:1706` / `server.go:1448`), but now it's correct.

**4. nit — acknowledged, no change.** A `#token=` planted on any path is a token the attacker already holds, and narrowing to `/login`-suffixed paths would fight finding #1's "hash always wins".

Two corrections taken:
- Right that this isn't a full mirror of #100 — the CF probe is idempotent/re-runnable, a hash token is one-shot, which is exactly why inheriting the no-token gate was the bug. Comment reworded.
- Correct that `build-web` isn't CI-verified — CI runs only `go-unit` + `integration`; that was a local build. The PoC is committed now so there's a runnable regression check for the guard.

unidoc-alip (dismissed):
looks good to me

unidoc-alip (approved):
LGTM

PR #193, opened 2026-08-05
fix(security): authorize comment writes on reviews and register entities

unidoc-alip:
Fixes the comment-authorization issues reported privately to `security@isms.sh` per `SECURITY.md` (no public issue exists, hence no `(#n)` in the commit subject).

## What was wrong

Three comment write paths authorized nothing beyond "is authenticated":

| Route | Handler | Gap |
|---|---|---|
| `POST /comments` | `handleAddCommentDB` | binds a client-supplied `review_id`, no participant check, no review-status check |
| `POST /comments/:id/resolve` | `handleResolveCommentDB` | no check of any kind |
| `POST /entity-comments` | `handleCreateEntityComment` | no role gate |

`POST /reviews/:id/comment` has always checked requester-or-assignee — but `POST /comments` accepts the same write with the review named in the body, so the check guarded one door in a two-door room. A `reader` refused by the first route got `201` from the second, and the injected comment appears on the review timeline indistinguishable from a legitimate one. Neither route looked at review status, so comments also landed on **merged** reviews.

`RoleMiddleware`'s reader carve-out (`middleware.go`) asserted `// Allow readers to add comments (assignment check in handler)` — an invariant no handler implemented — and matched a bare `/api/v1/comments` prefix, so it also exempted `/accept` and `/reject` and would have exempted any future route beneath the prefix.

## The fix

- **One rule, both routes.** `authorizeReviewComment` + `isReviewParticipant` extracted from `handleAddReviewComment`; `handleAddCommentDB` now calls it whenever `review_id` is set. The rule is **assignment-based, not role-based** — per `CLAUDE.md`, "Review assignment grants approve/comment rights to any role", so an assigned reader still comments. A `role !== 'reader'` check would have been the wrong fix.
- **Review status is enforced** in the shared helper: `merged`/`closed` → `400 review is merged`. Checked before the participant rule so the message names the real reason.
- **`document_id` now comes from the review**, not the client, so it can't disagree with the review it's attached to.
- **Resolve is authorized**: admin/manager, the comment's author, or a participant in the comment's review.
- **`POST /entity-comments` gated to contributor and above**, mirroring `handleCreateEntitySuggestion` and its #23 rationale — the two are the contributor's write surface and had drifted apart.
- **Middleware exemption narrowed** to `POST /api/v1/comments` and the concrete `/comments/:id/resolve` path, anchored on prefix *and* suffix together so `/entity-comments/:id/resolve` cannot match it. The handler checks are the real fix; this is defence in depth.
- **Frontend**: hid the four affordances this change would otherwise turn into dead buttons — the `CommentsPanel` composer for readers (it's on every register entity page), the Inbox **Resolve** button for users who can't resolve, and in `DocumentViewer` the per-comment **Reply** and **Resolve** controls (new `can-comment` / `can-resolve-comments` props, with `Reviews.vue` passing the participant rule from `authorizeReviewComment` / `handleResolveCommentDB`). Both props default to `true`, so a review-less document viewer is unchanged: readers keep document commenting, which is what the middleware exemption exists for.

## Decision needing your confirmation

The **resolve rule** is a product call I made rather than found: *admin/manager, the comment's author, or a participant in the comment's review*. The reasoning is that the document author resolving reviewer feedback is the common case (covered by "participant"), and an author closing their own comment is uncontroversial. If you'd rather restrict it to admin/manager plus the author, it's a two-line change — say so and I'll adjust.

## Verification

- 16 new integration tests in `tests/test_comment_authorization.py`. The load-bearing ones are in `TestAssignedReaderKeepsAccess`: an assigned reader must keep access through **both** routes. A blunt role-based fix passes every refusal test in this file and fails those.
- Also covered: document comments without `review_id` still `201` for a reader; `document_id` taken from the review; `400` on a merged review even for an admin; author-resolve and participant-resolve allowed, non-participant reader refused with the comment verified still open; reader refused and contributor allowed on `entity-comments`.
- **Idempotent on a persistent stack**: the file was run twice against the same database with no reset between runs — 16 passed, then 16 passed. `handleReviewSend` reuses an already-open review, `AddReviewAssignmentTx` is `ON CONFLICT DO NOTHING`, and each test posts fresh comments.
- Full suite: **673 passed**, 1 failure — `test_brute_force_protection`, which asserts a `429` from the login limiter that the documented test env (`ISMS_RATE_LIMIT=0`) disables. Pre-existing and unrelated to this change. The test already carries a `skipif` for exactly this configuration; it did not fire because the guard reads `ISMS_RATE_LIMIT` from the **pytest process's** environment while that run exported it only to the server. Exporting it to pytest as well — which CI does job-wide — skips the test as intended, so there is nothing to fix. An earlier revision of this description misattributed the failure to `RoleMiddleware`; that was wrong, `RoleMiddleware` does correctly skip `/api/v1/auth/` (`middleware.go:363`).
- `go build ./...`, `go vet ./...`, `npm run build` all clean. No schema or migration changes.

## Related

- #192 — the reader-visible UI affordances on the Reviews and Inbox pages that already `403` **before** this change, and so are deliberately left there. The inline paragraph composer in `DocumentViewer` is one of them: it posts to the always-gated `/reviews/:id/comment`, so this change doesn't alter its behaviour and hiding it stays #192's job. Only the controls this PR would *newly* break are fixed here — the four listed above, whose paths (`POST /comments` with a `review_id`, and `POST /comments/:id/resolve`) returned `201`/`200` to anyone before this change.
- #23 — readers are read-only; the rule `entity-comments` was missing.

### DIFF
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -843,6 +843,49 @@ func (s *Server) handleReviewDiff(c echo.Context) error {
 	})
 }
 
+// isReviewParticipant reports whether actor is the review's author or one of its
+// assigned reviewers. Assignment — not role — is what grants comment rights on a
+// review, so a reader assigned as a reviewer participates fully (see CLAUDE.md,
+// "Review assignment grants approve/comment rights to any role").
+func (s *Server) isReviewParticipant(ctx context.Context, orgID int, review *db.Review, actor string) bool {
+	if review == nil || actor == "" {
+		return false
+	}
+	if review.RequestedBy == actor {
+		return true
+	}
+	assignments, _ := s.db.ListAssignmentsForReview(ctx, orgID, review.ID)
+	for _, a := range assignments {
+		if a.Reviewer == actor {
+			return true
+		}
+	}
+	return false
+}
+
+// authorizeReviewComment is the single gate for writing a comment onto a review.
+// Both comment routes must call it: POST /reviews/:id/comment and POST /comments
+// with a client-supplied review_id. Without it on the latter, the check here is
+// decorative — one guarded door in a two-door room.
+func (s *Server) authorizeReviewComment(ctx context.Context, orgID int, reviewID int, actor, role string) (*db.Review, error) {
+	review, err := s.db.GetReview(ctx, orgID, reviewID)
+	if err != nil {
+		return nil, echo.NewHTTPError(http.StatusNotFound, "review not found")
+	}
+	// A published or abandoned review is a closed record — nobody appends to it.
+	// Checked before the participant rule so the message names the real reason.
+	if review.Status == "merged" || review.Status == "closed" {
+		return nil, echo.NewHTTPError(http.StatusBadRequest, "review is "+review.Status)
+	}
+	if role == "admin" || role == "manager" {
+		return review, nil
+	}
+	if !s.isReviewParticipant(ctx, orgID, review, actor) {
+		return nil, echo.NewHTTPError(http.StatusForbidden, "not authorized to comment on this review")
+	}
+	return review, nil
+}
+
 // handleAddReviewComment adds a comment tied to a specific review.
 func (s *Server) handleAddReviewComment(c echo.Context) error {
 	orgID := getOrgID(c)
@@ -852,27 +895,11 @@ func (s *Server) handleAddReviewComment(c echo.Context) error {
 	}
 	ctx := c.Request().Context()
 
-	// Verify review exists.
-	review, err := s.db.GetReview(ctx, orgID, id)
-	if err != nil {
-		return echo.NewHTTPError(http.StatusNotFound, "review not found")
-	}
-
-	// Authorization: admin/manager always, others must be assigned
 	actor := getUserEmail(c)
 	role, _ := c.Get("user_role").(string)
-	if role != "admin" && role != "manager" {
-		assignments, _ := s.db.ListAssignmentsForReview(ctx, orgID, id)
-		isAssigned := false
-		for _, a := range assignments {
-			if a.Reviewer == actor {
-				isAssigned = true
-				break
-			}
-		}
-		if !isAssigned && review.RequestedBy != actor {
-			return echo.NewHTTPError(http.StatusForbidden, "not authorized to comment on this review")
-		}
+	review, err := s.authorizeReviewComment(ctx, orgID, id, actor, role)
+	if err != nil {
+		return err
 	}
 
 	var req struct {
@@ -1704,6 +1731,20 @@ func (s *Server) handleAddCommentDB(c echo.Context) error {
 		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
 	}
 	comment.Author = getUserEmail(c) // always use authenticated user
+	// review_id arrives from the client, so a comment aimed at a review must pass
+	// the same gate as POST /reviews/:id/comment — otherwise this route is a way
+	// around it. Plain document comments (no review_id) stay open to every role,
+	// which is what the reader exemption in RoleMiddleware exists for.
+	if comment.ReviewID != nil {
+		role, _ := c.Get("user_role").(string)
+		review, err := s.authorizeReviewComment(c.Request().Context(), orgID, *comment.ReviewID, comment.Author, role)
+		if err != nil {
+			return err
+		}
+		// Trust the review, not the client, for which document this lands on.
+		comment.DocumentID = review.DocumentID
+	}
+	comment.Status = "open" // the row is inserted open; echo what was stored
 	// If suggestion_body is provided, set suggestion_status to pending
 	if comment.SuggestionBody != nil && *comment.SuggestionBody != "" && comment.SuggestionStatus == nil {
 		pending := "pending"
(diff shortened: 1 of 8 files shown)


unidoc-ahall (dismissed):
Traced every comment-write path in the tree to confirm this closes the reported gaps without leaving a side door. Solid fix — the assignment-based (not role-based) rule is exactly right per the CLAUDE.md invariant, and `TestAssignedReaderKeepsAccess` is the test that actually matters.

**Completeness — checked all comment-write surfaces, all covered:**
- `POST /reviews/:id/comment` and `POST /comments` (with `review_id`) → both now go through `authorizeReviewComment`. The two-door-room framing was accurate; both doors are locked.
- `POST /comments` without `review_id` (document comment) → stays open to readers, the intended exemption.
- MCP `comment_on_review` → posts to `/reviews/:id/comment` via the REST client, so it inherits the same gate. No separate DB path to worry about.
- `POST /comments/:id/accept` / `/reject` → already gated to admin/manager or the review author, and the middleware narrowing now also stops readers reaching them (they were reader-exempt under the old bare `/api/v1/comments` prefix — a real incidental fix).
- `POST /entity-comments` → now contributor-and-above; `/entity-comments/:id/resolve` and `DELETE` were already admin/manager.

No remaining bypass.

**Resolve rule — confirming your call, keep it as-is.** admin/manager + comment author + review participant is right. A document author resolving reviewer feedback is the legitimate common case and it's covered by "participant"; restricting to author-only would break it for no security gain.

**Non-blocking:**
1. **Idempotency on the persistent stack** — I traced it: `handleReviewSend` reuses an already-open review (doesn't 409), `AddReviewAssignmentTx` is `ON CONFLICT DO NOTHING`, and each test posts fresh comments, so a second run is safe. Worth a quick double-run to confirm per our usual convention, but the design holds.
2. **Agent flows** — since the generic `/comments` bypass is now closed, worth a sanity check that no MCP/automation was leaning on it to comment on a review it isn't assigned to. That path now 403s (correctly) — but if some flow depended on it, behavior changes. The legitimate agent path (`create_suggestion`, and `comment_on_review` when assigned) is unaffected.
3. **Inbox `canResolveComment`** hides Resolve from a review participant the server *would* let resolve (the inbox row lacks review/participant data). You already flag this — fine to leave; a later tweak to carry `review_id`/participant in the inbox payload would let the button match the server rule exactly.
4. Agree the `test_brute_force_protection` failure is pre-existing and unrelated. `RoleMiddleware` not skipping `/api/v1/auth/` in the `ISMS_RATE_LIMIT=0` env is a real, separate small bug — worth its own issue so it doesn't rot rather than folding it in here.

Approving — complete fix, right invariant, real tests.

unidoc-alip:
Thanks — worked through the four non-blocking items. Two of them turned up corrections to things I'd written in the description, so flagging both rather than quietly fixing them. Pushed as c1ff01e; description updated.

**Correction 1 — item 4 is not a bug, and the component I named was wrong.** My description said "`RoleMiddleware` skips `/api/v1/auth/` entirely. Its own docstring says it should skip in this configuration but the skip isn't implemented", which you reasonably restated as `RoleMiddleware` not skipping `/api/v1/auth/`. That sentence was wrong twice over and I'd rather it didn't become an issue.

`RoleMiddleware` does correctly skip `/api/v1/auth/` (`middleware.go:363`). The skip that matters is a `@pytest.mark.skipif` on `tests/test_security.py:81`, keyed on `ISMS_RATE_LIMIT == "0"` — it is implemented, and it is pre-existing on `master`, untouched by this branch. It didn't fire because it reads `ISMS_RATE_LIMIT` from the **pytest process's** environment and my run exported it only to the server. Proven both ways against the same server:

- without `ISMS_RATE_LIMIT` in the pytest env: `FAILED (assert 401 == 429)`
- with `ISMS_RATE_LIMIT=0` in the pytest env: `SKIPPED`

CI sets it job-wide (`tests.yml:82`), which is why it passes there. So: no product bug, nothing to file. Sorry for pointing you at a phantom.

**Correction 2 — item 2 found a real gap, in the frontend rather than in agent flows.** Your MCP trace holds: `comment_on_review` posts to `/reviews/:id/comment` (`mcp/server.go:615`) and inherits the gate, and `authorizeReviewComment` short-circuits admin/manager, so no privileged automation is newly refused. The one thing not knowable from the repo: an external script posting to `/comments` with a `review_id` on a non-admin/manager token that isn't a participant now correctly 403s.

What the trace didn't reach was the web client. `api.addComment` → `POST /comments`, and its caller that sets `review_id` is `DocumentViewer.vue`'s per-comment **Reply** composer, rendered from the review page. That component wasn't in my frontend pass, and its Reply button was guarded only by `comment.status !== 'resolved'`. The per-comment **Resolve** button next to it had no guard either, and resolve previously checked nothing at all.

Both were **newly** dead — Reply returned 201 before this branch (I confirmed against `origin/master`: the old `handleAddCommentDB` binds and inserts with zero authorization), and Resolve succeeded for everyone. That distinction matters because it makes my "#192 — those 403 today, before this change" justification inaccurate for these two. They belong here, not in #192. The inline paragraph composer in the same component does post to the always-gated review route, so it genuinely is unaffected and stays #192's.

Fixed with `can-comment` / `can-resolve-comments` props, `Reviews.vue` passing the assignment-based rule (admin/manager, review author, or assigned reviewer) plus the merged/closed refusal from `authorizeReviewComment`; Resolve additionally admits the comment's own author per comment, mirroring `handleResolveCommentDB`. Both default to `true` so a review-less viewer is unchanged and readers keep document commenting.

Checked the three ways this could have been a guard that compiles but never fires: `getCurrentUser()` returns the email that comments carry as `author`; `review.status` uses the same `merged`/`closed` literals as four other spots in that view; and `CommentsForDocument` filters strictly with `AND review_id = $3`, so a review-scoped list never mixes in `review_id IS NULL` comments and the participant rule applies to every comment shown.

**Items 1 and 3.** Idempotency confirmed as you predicted — `test_comment_authorization.py` twice against the same database with no reset between runs, 16 passed then 16 passed. Left `canResolveComment` in the inbox as-is per your call.

unidoc-ahall (approved):
Re-reviewed c1ff01e — verified the delta rather than re-litigating the base.

**Item 2 (the frontend gap) — fix is correct and correctly scoped.** Confirmed in the code: `submitReply` posts via `api.addComment` → `POST /comments` with a `review_id` (the newly-gated bypass), now guarded by `props.canComment`; the inline paragraph composer posts via `api.addReviewComment` → `/reviews/:id/comment` (always gated), so it genuinely is unaffected and stays #192's. The distinction holds in the source. `canCommentOnReview` faithfully mirrors `authorizeReviewComment` — `canWrite` is admin/manager, OR participant, minus merged/closed — and `canResolveComment` mirrors `handleResolveCommentDB` (plus the comment's own author, per row). Both props default `true` and `DocumentViewer` is only mounted from `Reviews.vue`, so nothing else shifts. Good catch that these two were *newly* dead rather than pre-existing #192 controls — that reclassification is right.

**Item 4 — your correction stands; I'm withdrawing the issue rec.** Verified both: `RoleMiddleware` does skip `/api/v1/auth/` (`middleware.go:363`), and the `skipif` on `ISMS_RATE_LIMIT == "0"` exists (`test_security.py:81`). The failure was only the env not reaching the pytest process, which CI sets job-wide — no product bug, nothing to file. My item 4 restated the description's error; glad you chased it to ground rather than letting it become a phantom issue.

Items 1 (idempotency — confirmed twice against the same DB) and 3 (inbox left as-is) are settled.

Re-approving. The backend gate is the security boundary and was solid on the first pass; this is a clean, correctly-scoped frontend follow-up.

PR #209, opened 2026-08-15
feat(web): render Mermaid diagrams in Markdown

### LINKED ISSUE #208: [Feature]: Render Mermaid diagrams in Markdown content
### Problem / motivation

ISMS stores documents and register free-text fields as Markdown, but fenced
Mermaid blocks are currently displayed as ordinary source code.

Diagrams are useful for documenting review workflows, processes, system
relationships, incident timelines, and threat models. Today, authors must either
use an external rendering tool or commit diagram screenshots.

Screenshots are difficult to review and maintain: their source is not
diff-friendly, searchable, or editable alongside the surrounding Markdown.
External rendering also breaks the self-contained document workflow.

It would be useful if ISMS could render Mermaid fences in read-only views while
keeping the original fenced Markdown as the canonical, editable source.

### Proposed solution

Recognize fenced code blocks whose first info-string token is `mermaid` and
render them as diagrams in read-only Markdown surfaces.

Expected behaviour:

- Render Mermaid blocks in documents, register Markdown fields, reviews, and
  document diff/change views.
- Keep Mermaid source as a normal editable code block in the TipTap editor.
- Preserve the complete fence info string and source formatting through
  Markdown-to-editor-to-Markdown round-trips.
- Keep existing syntax highlighting and code-block behaviour unchanged for
  non-Mermaid fences.
- Render diagrams client-side and lazy-load Mermaid only when a page contains a
  diagram.
- Isolate invalid syntax to an accessible inline error state without breaking
  the rest of the document.

The rendering boundary should remain explicit:

- Mermaid runs with `securityLevel: "strict"`.
- HTML labels are disabled.
- Source enters the renderer as inert text rather than trusted HTML.
- Generated SVG is sanitized with DOMPurify before insertion.
- Rendered SVG is display-only and is never written back to stored Markdown.

Comment and mention rendering should remain outside this feature because those
surfaces use a separate, more constrained rich-text security model.

Live diagram editing, diagram-specific toolbar controls, image export, and
server-side rendering are out of scope.

I have a working, tested implementation and browser screenshots available, but
wanted to confirm that this scope and approach fit the project before opening a
pull request.

### Alternatives considered

**Keep Mermaid as plain code.** This preserves the source but does not provide
the visualization authors are trying to communicate.

**Commit rendered images alongside Markdown.** Images work everywhere, but the
diagram and its source can drift apart. Image changes are also difficult to
review as text diffs.

**Render Mermaid server-side.** This could support non-browser consumers, but it
adds rendering, caching, and binary-generation complexity to the backend. The
current browser UI can render diagrams without changing the storage or API
model.

**Allow authors to embed raw SVG.** This weakens the Markdown security boundary
and makes stored content harder to edit and review safely.

SbstnErhrdt:
## Summary

- render fenced `mermaid` blocks across document, register, review, and diff Markdown surfaces
- preserve Mermaid source, complete fence info strings, wrapping metadata, and trailing blank lines through editor round-trips
- lazy-load Mermaid in read-only views while keeping the TipTap editor source-only
- add focused rendering, lifecycle, sanitization, and round-trip tests plus Markdown pipeline documentation

## Security

- Mermaid uses `securityLevel: "strict"` with HTML labels disabled
- diagram source is transported as inert text rather than trusted HTML
- generated SVG is sanitized with DOMPurify's SVG profiles before insertion
- invalid diagrams are isolated in an accessible inline error state
- rendered SVG remains display-only and is never written back to stored Markdown

### Rendered diagram

Rendered Mermaid diagram in the document read view

### Editable source

Mermaid source preserved in the document editor

### Invalid syntax

Controlled inline error state for invalid Mermaid syntax

## Validation

- `npx --yes --package=node@24 node --test` - 20/20 tests pass
- Node 24 production Vite build succeeds
- `gofmt -l .` reports no unformatted Go files
- `CGO_ENABLED=0 go build ./...`
- `CGO_ENABLED=0 go test ./internal/... -count=1 -timeout 120s`
- `go vet ./...`
- browser verification covered read view, editor source preservation, and invalid-syntax handling

A local macOS integration run also exposed two review E2E failures that reproduce
unchanged on upstream `master` at `7636eed`:

- `TestReviewTwo::test_both_approve`
- `TestReviewRoundUX::test_round_ui`

They are unrelated to the Mermaid changes; the pull request CI remains the
authoritative Linux integration run.

## AI assistance disclosure

This contribution was developed with assistance from OpenAI Codex. The resulting
code was reviewed, tested, and manually verified before submission.

Closes #208


unidoc-alip (changes requested):
> **Verdict: REQUEST_CHANGES** — 1 must-fix, 3 should-fix, 1 nit

Solid, well-scoped feature. The display/editor pipeline split is the right call, and the placeholder-then-sanitize boundary holds up under testing: mermaid runs with `securityLevel: strict`, the SVG-only DOMPurify profile drops `<script>`, and root-level `htmlLabels: false` is the correct key for mermaid 11.x (verified against `config.type.d.ts` — the deprecated `flowchart.htmlLabels` is not needed). Comment and mention surfaces are correctly excluded, since they never run Markdown at all. All 16 new tests pass and the Vite build is clean on my machine too.

Every finding below was reproduced with a runnable proof of concept on this branch, and each suggested fix was applied locally and re-verified against the full test suite before being reverted.

Findings are keyed `F1`, `F2`, … rather than `#1`, `#2`, because GitHub autolinks `#N` to unrelated issues.

---

### F1 [must-fix] `web/src/composables/useMarkdownConvert.js` line 34

`escapeHtml` (`web/src/utils/html.js`) escapes only `&`, `<`, `>` — not `"` — so a fence info string containing a double quote breaks out of the new `data-info-string` attribute and corrupts the round-trip this PR is built to guarantee.

Proof of concept, run against this branch:

````js
const source = '```mermaid" data-wrapped="true\nflowchart LR\n  A --> B\n```'
const html = markdownToHtml(source)
const roundTripped = htmlToMarkdown(html).trim()
````

Output:

````
EDITOR HTML:
<pre data-info-string="mermaid" data-wrapped="true"><code class="language-mermaid"">…

ROUND-TRIPPED MARKDOWN:
```mermaid wrap

fence info preserved?      false
wrap flag silently added?  true
custom token survived?     false
````

Saving that document rewrites the user's opening fence from `mermaid" data-wrapped="true` to `mermaid wrap` — the custom token is dropped and word-wrap is switched on that the user never asked for. The same string is also an arbitrary-attribute injection into the HTML that TipTap parses. Nothing executes today, since `markdownToHtml` output only reaches `setContent()` and ProseMirror discards attributes outside the node schema — but the primitive shouldn't exist on a change whose description leads with its security boundary.

The fix needs two lines, not one:

```diff
 // web/src/utils/html.js
 export function escapeHtml(s) {
   return s
     .replace(/&/g, '&amp;')
     .replace(/</g, '&lt;')
     .replace(/>/g, '&gt;')
+    .replace(/"/g, '&quot;')
+    .replace(/'/g, '&#39;')
 }
```

```diff
 // web/src/composables/useMarkdownConvert.js line 32
-      const langClass = lang ? ` class="language-${lang}"` : ''
+      const langClass = lang ? ` class="language-${escapeHtml(lang)}"` : ''
```

The `escapeHtml` change alone stops the breakout but still loses the `"` from the first token, because turndown reads the language back out of the equally-unescaped `class` attribute. With both changes the round-trip is byte-exact — `fence info preserved? true / wrap flag silently added? false / custom token survived? true` — and `node --test` stays 16/16. The `escapeHtml` docblock already anticipates exactly this: *"if attribute-safety escaping (e.g. `"` → `&quot;`) is ever needed, it's added here once."*

The `language-${lang}` interpolation is pre-existing and not a regression from this PR, but the round-trip doesn't actually close without it.

---

### F2 [should-fix] `web/src/composables/useMermaid.js` line 22

Under `theme: 'dark'`, most of this `themeVariables` block never reaches the SVG. Mermaid's dark theme hardcodes its palette *after* deriving from `primaryColor` (`mermaid/dist/chunks/mermaid.core/chunk-I66GZJ75.mjs:462` → `this.mainBkg = "#1f2020"`, `:473` → `this.textColor = "#ccc"`), so the overrides are discarded. Rendering `flowchart LR / A[Draft] --> B[Review]` with this exact config and inspecting the emitted `<style>`:

```
background         #020617  -> IGNORED (dark-theme default wins)
primaryColor       #1e293b  -> IGNORED
primaryTextColor   #e2e8f0  -> IGNORED
primaryBorderColor #475569  -> IGNORED
lineColor          #64748b  -> APPLIED
secondaryColor     #0f172a  -> IGNORED
tertiaryColor      #172554  -> APPLIED

node fill emitted : #1f2020   (warm grey — not the intended slate #1e293b)
node text emitted : #ccc
container bg      : #0f172a   (slate-900, from .mermaid-diagram in style.css)
```

Diagrams therefore render as warm-grey boxes inside a slate container — visibly off-palette against the rest of the app — and the config reads as if it were theming when it mostly isn't.

This also lands on the new print rule. `@media print` forces `.mermaid-diagram { background: transparent !important }`, putting the diagram on a white page while it is still drawn for a dark canvas. Text with a backing shape survives — node labels are `#ccc` on `#1f2020`, 10.17:1 — but text drawn straight onto the container does not: a diagram title renders `#ccc` on white at **1.61:1**, invisible on paper and in PDF export.

Switching to the `base` theme is mermaid's documented way to let `themeVariables` drive the palette. Verified by rendering the same diagram under both:

```
theme: 'dark'  -> themeVariables applied: 2/7   node fill: #1f2020
theme: 'base'  -> themeVariables applied: 6/7   node fill: #1e293b   (intended slate)
```

```diff
-  theme: 'dark',
+  // 'base' is the only theme that lets themeVariables drive the palette;
+  // 'dark' hardcodes mainBkg/textColor and discards most overrides.
+  theme: 'base',
```

With `base` applied, also set `textColor` and `titleColor` explicitly — they're the ones with no backing shape — and either re-render light for print or add a print override such as `.mermaid-diagram svg text { fill: #0f172a !important }` alongside the existing `.mermaid-diagram` print rule.

---

### F3 [should-fix] `web/src/components/codeLanguages.js` line 22

The header comment says Mermaid "deliberately stays plain source in the editor", but that isn't what happens. `@tiptap/extension-code-block-lowlight` falls through to `lowlight.highlightAuto()` for any language it can't resolve (`dist/index.js:39`), and lowlight's `common` set has no `mermaid` grammar.

Replaying that exact branch:

```
lowlight.registered("mermaid") : false
hljs.getLanguage("mermaid")    : false
lowlight.highlight("mermaid")  -> Unknown language: `mermaid` is not registered

tiptap picks highlight(language) ?  false
=> falls through to lowlight.highlightAuto(), which guesses:
   detected language : rust
```

So picking **Mermaid** in the DocumentEditor dropdown hands `flowchart LR / Draft --> Review` to auto-detection, which colourises it as Rust. Not a crash, but the opposite of the documented intent — and `codeLanguages.test.js` only asserts the entry exists in the array, so the suite can't catch it.

Registering a no-op grammar next to the existing `createLowlight(common)` call in `DocumentEditor.vue` fixes it:

```js
const lowlight = createLowlight(common)
// Mermaid has no grammar in `common`; register an empty one so the editor keeps the
// source plain instead of falling through to highlightAuto's best guess.
lowlight.register('mermaid', function mermaid () { return { name: 'mermaid', contains: [] } })
```

Verified: `registered: true`, zero `hljs-*` spans emitted, source text byte-identical.

---

### F4 [should-fix] `web/src/components/TrackChanges.vue` line 45

For a block of `type: 'change'`, the HTML comes from `wordDiffHtml()` (`TrackChanges.vue:399`), which injects `<del>`/`<ins>` markers into the raw text *before* `parseMd()`. Inside a fence those markers are literal text, so an edited diagram reaches the placeholder with diff markup baked into its source.

Reproducing the `change` path verbatim — `wordDiffHtml` → `parseMd` → `renderMermaidDiagrams` with real mermaid — while editing `Draft --> Review` into `Draft --> Review --> Approved`:

```
Mermaid source it will parse:
"flowchart LR\n  Draft --> Review<ins class=\"tc-word-ins\"> --&gt; Approved</ins>"

resulting state : error
role            : alert
visible text    : "Diagram could not be rendered"
source still on the page for the reviewer?  false
```

The outcome is worse than simply not rendering: `showError()` calls `replaceChildren()`, so the reviewer gets a bare error box with no way to read what changed — exactly where the source matters most.

Keeping the source visible on failure fixes this and also improves the plain "user typed invalid syntax" case. Thread the already-claimed `job.source` into `showError`:

```diff
-function showError(target) {
+function showError(target, source) {
   target.replaceChildren()
   target.removeAttribute('data-mermaid-pending')
   target.dataset.mermaidState = 'error'
   target.setAttribute('role', 'alert')

   const message = document.createElement('span')
   message.className = 'mermaid-error-message'
   message.textContent = 'Diagram could not be rendered'
-  target.append(message)
+  const pre = document.createElement('pre')
+  pre.className = 'mermaid-error-source'
+  pre.textContent = source          // textContent — inert, adds no HTML sink
+  target.append(message, pre)
 }
```

Verified: the error state keeps `role="alert"`, the source is recoverable, and `textContent` assignment injects no markup. `.mermaid-diagram[data-mermaid-state="error"]` will need its `color` reset so the source is actually readable. Alternatively, drop `v-mermaid` from the changed-block path in `TrackChanges`/`DocumentDiff` entirely — diff-marked diagram source can never render meaningfully.

---

### F5 [nit] `web/package.json` line 33

Bundle impact, measured by building both branches with a clean `npm ci` on each:

```
master (7636eed) total JS in dist/assets : 2,083 kB
this branch                              : 5,441 kB   (+3,358 kB, +161%)
```

Largest additions: `chunk-KEIR6QF5` 663 kB, `cytoscape.esm` 436 kB, `katex` 259 kB, `architectureDiagram` 149 kB, `sequenceDiagram` 116 kB, `swimlanes` 114 kB, plus roughly 35 smaller per-diagram-type chunks.

All correctly code-split behind the dynamic `import('mermaid')`, so readers who never open a document containing a diagram pay nothing. No change requested — just worth recording that the repo's total JS surface roughly triples. If it ever matters, mermaid supports registering only the diagram types you use to drop the long tail.

unidoc-alip (approved):
> **Verdict: APPROVE** — all four actionable findings addressed and verified at `3addb7c`

Re-ran each original proof of concept against the new head. Every one passes, the suite is 20/20 with four new tests, and the build is clean.

**F1** — both halves landed (`"`/`'` escaping in `utils/html.js` *and* `escapeHtml(lang)` on the class attribute). Round-trip is byte-exact: `fence info preserved? true / wrap flag silently added? false / custom token survived? true`. No attribute breakout.

**F2** — `theme: 'base'` plus explicit `textColor`/`titleColor` takes the palette from 2/9 applied to 8/9 (only `background` remains inert, which is expected — the base theme doesn't use it for flowcharts). Node fill is the intended `#1e293b`.

The print fix is better than what I suggested. Computing the `invert(1) hue-rotate(180deg)` chain per the CSS filter spec against the colours mermaid actually emits:

```
title text  #e2e8f0 -> #121e0d   vs white page : 17.25:1  PASS  (was 1.61:1, invisible)
node text   #e2e8f0 -> #121e0d   vs node fill  : 15.68:1  PASS
node fill   #1e293b -> #cdfdff   vs white page :  1.10:1  (light box, minimal ink)
```

No second render path, and boxes print light instead of solid dark. Nice.

**F3** — `registered('mermaid')` is now true, so tiptap takes the `highlight()` path instead of `highlightAuto()`; zero `hljs-*` spans, source byte-identical. Regression-checked that Go still highlights and all 38 languages remain registered.

**F4** — `showError(target, source)` threads the source through both call sites; the diff-marked source is visible again instead of a bare error box. I probed the new path with `` inside a fence: `pre.innerHTML` is fully escaped and zero `img` elements are created. Good call adding the `<script>` regression test.

**F5** was informational only — no change expected.

Two non-blocking notes for whenever you next touch this:

- `editorLowlight.test.js` still wraps its import in a bare `try`/`catch` from the TDD phase. Now that the module exists, an unrelated import error gets swallowed and resurfaces as a confusing "expected function, got undefined" — a plain top-level import would fail more honestly.
- The print fix relies on the print engine honoring CSS `filter`. Chrome and Safari do; anything that drops it degrades to the previous behaviour rather than something worse.

Thanks for the thorough turnaround — the added tests pin every fix at the right level.

Issue #212, opened 2026-08-18
[Feature]: Add i18n support and Brazilian Portuguese (pt-BR) localization


leonardo-ods:
### Problem / motivation

I’m evaluating ISMS for self-hosted production use in a Brazilian company.

We are already ISO 9001 certified and I’m interested in using ISMS to manage our QMS, as well as using the ISO 27001 templates to structure our IT/security processes.

The main issue for us is that the application is currently English-only. Browser translation is not a good solution for production use, especially when users are creating and editing content in Portuguese.

### Proposed solution

Would you be open to adding proper i18n/localization support to the application? I would be happy to contribute the Brazilian Portuguese (pt-BR) translation.

Is this something you would be interested in? If so, I’d be happy to discuss the preferred approach before starting the translation.

unidoc-alip:
Hi @leonardo-ods thank for the offering to do the pt-BR translation, and yes we want to add the feature for the i18n support in the application side.

Currently we are looking into approach for adding the i18n support.

leonardo-ods:
Hi @unidoc-alip !
Thanks for the update! Great to hear that. I’ll wait for the i18n approach to be defined, and I’d be happy to contribute the pt-BR translation once the structure is ready.

unidoc-alip:
Hi @leonardo-ods, we are done with the support for the multi-language.

You can follow this PR https://github.com/unidoc/isms/pull/280 regarding the translation or multi-language support.
That PR is example I am adding the "Indonesian" as second language.

unidoc-alip:
Closing this out — i18n/multi-language support has landed in the app (see PR #280, which added Indonesian as a worked example of adding a new locale).

We don't currently have anyone on the team who's a native Brazilian Portuguese speaker to do the pt-BR translation itself, so that part hasn't happened yet. @leonardo-ods if you're still up for contributing the pt-BR translation, please open a new issue specifically for that and we can help you get it structured using the same approach as the Indonesian PR.

leonardo-ods:
Hi @unidoc-alip!
Thanks for implementing the i18n support and for pointing me to PR #280.

Yes, I'm still interested in contributing the Brazilian Portuguese (pt-BR) translation. I'll take a look at the Indonesian PR first to understand the approach, and then I'll open a new issue specifically for the pt-BR localization.

Thanks again for the support!

PR #218, opened 2026-08-24
feat(i18n): ship the id-ID message bundle

unidoc-alip:
Part of #212. Rebased onto master now that the foundation has landed as #217 — this diff is only the locale, and the branch is a single commit on top of master.

Adds `web/src/locales/id-ID/`, mirroring `en/` file for file and key for key, and registers its lazy loader.

## Why id-ID and not pt-BR

The foundation needs a second locale that is not the fallback to prove the resolution chain, and it has to be one we can author and review ourselves.

pt-BR remains the goal of #212 and gets contributed later, against the frozen keyset — at which point adding it is one entry in the loader map plus a locale directory. Nothing here is id-ID-specific.

## Scope

The keyset is deliberately small. Nothing is extracted from components yet, so `common.json` is all there is to mirror; area files appear here as the matching `en/` ones land with their extraction PRs. What this proves is the mechanism, not coverage.

Two tests are added:

- the load-and-render path for a non-fallback locale;
- the fallback path — a key removed from the id-ID bundle at runtime renders English rather than a raw key, which is what lets CI warn instead of fail on a lagging translation.

## Two of #217's tests had to be retargeted

Both were written while id-ID was the stock example of a tag the build *cannot* render, so shipping it invalidated their premise rather than their intent:

- `loadable locales always include the bundled fallback` now uses `fr-FR` as the unrenderable tag, and gains an assertion that a shipped tag *is* listed;
- `a superseded locale load does not overwrite the newer one` stashes and restores the real static `id-ID` loader instead of `delete`-ing the entry, which would otherwise have left every later test with no bundle to load.

## Verification

`npm test` 59/59, `npm run build` clean — id-ID is a separate 0.39 kB chunk (0.29 kB gzipped), so it costs nothing to users who do not select it. `localeKeyset.test.js` from #217 now covers id-ID; the keyset diff against `en` is clean in both directions, and the three identical values are acronyms (Global, EEA, APAC).

Exercised in a browser against a local server: with `isms_locale=id-ID` the app loads the id-ID chunk on demand, `<html lang>` becomes `id-ID`, a temporarily-bound placeholder rendered "Cari", and switching back to `en` restored "Search".

## The notification-keying blocker is cleared

This PR opens the window where a second locale is reachable through `navigator.languages` alone, with no picker required — which is why it could not merge while `notifications.title` and `body` were plain pre-rendered English `TEXT`. #220 has since landed, so that is no longer a reason to hold this back.

### DIFF
--- a/web/src/i18n.js
+++ b/web/src/i18n.js
@@ -149,10 +149,20 @@ let applySeq = 0
 // value (browser order, org default, fallback) never hardens into stored state
 // that would then outrank the org default on the next visit.
 //
+// One case suppresses the write even with `persist: true`: degrading to FALLBACK
+// because the chunk failed to load. That degrade is a transient failure, not a
+// choice of English, and `stored` outranks navigator.languages in the resolution
+// chain — so writing it would turn one bad network moment or one stale
+// index.html into a permanent silent downgrade that survives the chunk becoming
+// loadable again. Leaving storage untouched lets the next load retry the tag.
+//
 // Returns the applied tag, or null when a newer call superseded this one.
 export async function setLocale(tag, { persist = false } = {}) {
   const seq = ++applySeq
   let locale = negotiate(tag) ?? FALLBACK
+  // Distinguishes "ended up on English because the chunk broke" from every other
+  // route to English, which stay persistable.
+  let loadFailed = false
   if (locale !== FALLBACK && !i18n.global.availableLocales.includes(locale)) {
     try {
       const m = await loaders[locale]()
@@ -161,11 +171,12 @@ export async function setLocale(tag, { persist = false } = {}) {
     } catch {
       if (seq !== applySeq) return null
       locale = FALLBACK // chunk failed to load (offline, bad deploy)
+      loadFailed = true
     }
   }
   i18n.global.locale.value = locale
   if (typeof document !== 'undefined') document.documentElement.lang = locale
-  if (persist) {
+  if (persist && !loadFailed) {
     try {
       if (typeof localStorage !== 'undefined') localStorage.setItem(STORAGE_KEY, locale)
     } catch {
--- a/web/src/i18n.js
+++ b/web/src/i18n.js
@@ -24,7 +24,7 @@ export const STORAGE_KEY = 'isms_locale'
 // Exported so tests can register a controllable loader; production code adds a
 // static line here rather than mutating it at runtime.
 export const loaders = {
-  // 'id-ID': () => import('./locales/id-ID/index.js'),
+  'id-ID': () => import('./locales/id-ID/index.js'),
 }
 
 // Locales the server says it supports, as {tag, name} — name is the endonym, for
--- a/web/src/locales/id-ID/common.json
+++ b/web/src/locales/id-ID/common.json
@@ -1,0 +1,30 @@
+{
+  "action": {
+    "save": "Simpan",
+    "cancel": "Batal",
+    "confirm": "Konfirmasi",
+    "close": "Tutup",
+    "create": "Buat",
+    "edit": "Ubah",
+    "delete": "Hapus",
+    "search": "Cari"
+  },
+  "state": {
+    "loading": "Memuat…",
+    "empty": "Belum ada apa pun di sini",
+    "error": "Terjadi kesalahan"
+  },
+  "validation": {
+    "required": "Kolom ini wajib diisi"
+  },
+  "entity": {},
+  "field": {},
+  "enum": {},
+  "error": {},
+  "region": {
+    "global": "Global",
+    "eu": "UE",
+    "eea": "EEA",
+    "apac": "APAC"
+  }
+}
--- a/web/src/locales/id-ID/index.js
+++ b/web/src/locales/id-ID/index.js
@@ -1,0 +1,18 @@
+// The `id-ID` message bundle: one file per area, merged into a single nested
+// keyspace. Area = the top-level key, and it is also the filename — so
+// `risks.table.header.likelihood` will live in `risks.json` and nowhere else.
+//
+// This directory mirrors `en/` file for file and key for key. Only the values
+// are translated. A key missing here is not an error: `fallbackLocale` renders
+// it in English, so a lagging translation degrades gracefully rather than
+// showing a raw key — which is also why an area file only appears here once the
+// matching `en/` one exists.
+//
+// Import attributes (`with { type: 'json' }`) are required by Node for JSON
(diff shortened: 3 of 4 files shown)


unidoc-ahall (changes requested):
Independently verified — build, tests, and the keyset mirror are all clean. One real bug from Copilot's inline comment (`i18n.js:27`) is still open and needs a fix before this ships.

**Verified clean, reproduced locally on c0d62f1:**
- `npm test` — 59/59, `npm run build` — `id-ID` is its own 0.39 kB / 0.29 kB gzip chunk, exactly as described.
- `id-ID/common.json` is a byte-for-byte key mirror of `en/common.json` — walked both by hand, same 4 groups (`action`, `state`, `validation`, `region`) plus the three empty reserved groups (`entity`, `field`, `enum`, `error`), same 8/3/1/4 keys respectively. `global`/`eea`/`apac` are identical to English as claimed (Global stays as-is, EEA/APAC are the same acronym in Indonesian); `eu` is translated (`UE`).
- `localeKeyset.test.js` discovers locale directories from the filesystem, so it picks up `id-ID` automatically — confirmed it's exercising it (not just present in the diff) by checking `npm test`'s full run, not just this PR's own new test file.
- The `resolveInitialLocale`/`negotiate` test additions match the actual precedence chain and `SIMPLE_TAG` behavior in `i18n.js` — traced through by hand, not just read.

**Not addressed — Copilot's `i18n.js:27` finding is real, and it's live for the first time because of this PR:**

Checked what `loaders` looked like on `master` before this PR: `{ // 'id-ID': ... }` — commented out, so `loaders` was an **empty object** in production. In `setLocale()`, the branch `locale !== FALLBACK && !i18n.global.availableLocales.includes(locale)` can only be entered when `negotiate()` returns something other than `FALLBACK`, which requires a real entry in `loaders` — with `loaders` empty, that was structurally impossible, so the chunk-load-failure catch path (`locale = FALLBACK`) was dead code in production, only reachable through the existing tests' mock loaders. This PR is what makes it reachable for real for the first time.

The concrete scenario: a user explicitly picks `id-ID` (`persist: true`) and the dynamic `import()` fails — a transient network blip, or the classic SPA case of a stale `index.html` referencing a chunk hash a subsequent deploy rotated out from under it. `setLocale` sets `locale = FALLBACK` in the catch, then falls through to `if (persist) localStorage.setItem(STORAGE_KEY, locale)` — persisting `'en'`, not the tag that was actually requested. Since `resolveInitialLocale`'s precedence puts `stored` above `navigator.languages`, every later page load reads `stored = 'en'` and never renegotiates `id-ID` again, even after the chunk becomes loadable — a transient failure becomes a silent, permanent downgrade the user has to notice and manually undo.

Suggest not persisting when the fallback was reached via a load failure (vs. an actual choice of English) — e.g. track whether the catch fired and skip the `persist` block in that case, so the next load's `stored` value still points at the originally requested tag and the chunk gets retried. Worth a regression test alongside it: `persist: true` with a rejecting mock loader, assert `localStorage` still holds (or is absent) the pre-failure value, not `'en'`.

Everything else here is solid — requesting changes for this one before merge.

unidoc-alip:
Thanks — the finding is correct, and the reachability argument is the part that makes it worth fixing now rather than later. Fixed in 57e3c18.

I traced it rather than taking it on faith, and each step holds. `setLocale`'s catch reassigns `locale = FALLBACK` and then falls through to `if (persist) localStorage.setItem(STORAGE_KEY, locale)`, so the failure writes `'en'`. `resolveInitialLocale`'s candidate order is `[userLocale, stored, ...navigatorLocales, orgLocale]`, so `stored` does outrank `navigator.languages` — the downgrade survives every later load, not just the session. And `loaders` was `{}` on master, so the branch guarding the import could not be entered and the catch was unreachable outside tests; this branch is what arms it.

One correction to the framing, and it makes the bug less urgent than it reads: there is currently no production caller passing `persist: true`. `main.js:16` is the only `setLocale` caller and it passes nothing, so the path is latent until the locale picker lands. Fixing it now anyway — the guard is three lines, and the picker is exactly what turns it live.

## The fix

A `loadFailed` flag set only inside the catch, gating the persist write.

It is deliberately scoped to the catch rather than to `locale === FALLBACK`. Reaching English any other way is a real outcome that should still persist: an outright `setLocale('en', { persist: true })` is a genuine preference, and so is a tag that negotiates to nothing. Only "ended up here because the chunk broke" is suppressed, which leaves whatever was stored intact so the next load retries the originally requested tag.

## Test

Added the regression test you asked for, with the two boundary assertions alongside it — a successful load and an outright choice of English both still persist, so the guard has not quietly disabled persistence at large.

One note on how it is written, since it is not the obvious shape: it mocks `fr-FR` (rejecting) and `de-DE` (resolving) rather than using the real `id-ID` loader. My first attempt did use `id-ID` and it broke `a superseded locale load does not overwrite the newer one` — loading the shipped bundle leaves it resident in `availableLocales`, so that test's `setLocale` skips the import branch entirely, never awaits its gate, and returns the tag instead of `null`. Mock tags keep the two tests order-independent.

I also checked the test actually pins the fix rather than passing incidentally: reverting the guard fails that test and nothing else.

`npm test` 60/60, `npm run build` clean, `id-ID` still its own 0.39 kB / 0.29 kB gzip chunk.

unidoc-ahall (approved):
Verified independently, not just read — this closes the finding cleanly.

**Reproduced on 57e3c18:** `npm test` — 60/60, `npm run build` — `id-ID` still its own 0.39 kB / 0.29 kB gzip chunk.

**Confirmed the fix itself is correct** by tracing `setLocale`: `loadFailed` is set only inside the catch, and `if (persist && !loadFailed)` is the only gate on the storage write — so an outright `setLocale('en', {persist:true})` and a genuinely successful load both still persist (the two boundary assertions in the new test), and only the "reached English because the chunk broke" path is suppressed. Matches exactly what was asked for.

**Mutation-checked myself** rather than trusting the "reverting the guard fails that test and nothing else" claim: reverted `if (persist && !loadFailed)` back to `if (persist)` locally and reran `npm test` — exactly one test failed (`a failed chunk load does not persist the fallback over the stored tag`), the rest of the 60 stayed green. Restored the fix, back to 60/60.

**Checked the "not urgent" correction against the actual call sites**, not just accepted it: `main.js:16` is the only production caller of `setLocale`, and it's `setLocale(resolveInitialLocale())` — no options object, so `persist` defaults to `false`. Confirmed the bug was indeed latent until this fix, not live — fixing it now was still the right call given how cheap it was and that the picker will arm it.

Approving.

PR #223, opened 2026-08-25
fix: Cloudflare Access login (RS256 + probe ordering) + FreeBSD build

### LINKED ISSUE #221: Cloudflare Access login fails when the team's JWKS signs with RS256 (verifier parses EC keys only)
## Problem
The Cloudflare Access JWT verifier (`cfKeyCache` in `internal/isms/api`) parses **EC (ES256) keys only** — it drops every non-EC key at JWKS parse (`if k.Kty != "EC" { continue }`). But CF Access signs its Access-application JWT with **RS256 (RSA)** *or* **ES256 (EC-P256)** depending on the team and on key rotations, and a team's JWKS at `/cdn-cgi/access/certs` can carry both. When a team's token is RS256 the RSA key is discarded, so the token's `kid` is never found and `VerifyJWT` returns "unknown key ID"; the CF Access login path 401s and SSO login is impossible.

It fails **closed** (rejects a valid token; never accepts an invalid one), so it's a functionality bug, not a security exposure — but it makes CF Access SSO unusable on any team whose JWKS signs with RSA.

## Fix
Parse RSA keys (`kty=RSA`, `n`/`e`) as well as EC, and dispatch signature verification on the JWT `alg` (RS256 → RSA, ES256 → EC), cross-checking `alg` against the resolved key type so there is no alg-confusion opening. Also: on an unknown `kid`, refresh the JWKS once past the 5-minute throttle (handles key rotation); bound the JWKS fetch with an HTTP timeout; reject an unexpected `alg` explicitly.

## Verification
Go unit tests mint RS256 and ES256 tokens against an httptest JWKS that carries both key types: RS256 verifies (the regression), ES256 still verifies, and wrong-audience / wrong-issuer / expired / bad-signature / alg-key-mismatch / unsupported-alg all reject.

### LINKED ISSUE #222: Cloudflare Access login never attempted: the SPA guard's CF probe never runs on public routes
## Problem
`web/src/router.js`'s global navigation guard tries to mint a session from a Cloudflare Access identity — `if (!getApiToken()) { ...; const cf = await api.cfSession(); ... }` — but that block sits AFTER the `meta.public` early return (`if (to.meta.public) { ...; return true }`).

Every route a fresh, not-yet-locally-authenticated visitor actually lands on is `meta.public`: `/`, `/login`, and the org-scoped `/<org>/login`. All three return `true` before the guard ever reaches the CF-probe branch. A Cloudflare-Access-authenticated visitor with no local ISMS token can land on the app, see the marketing page or a login form, and never once have their CF identity checked — the only way in is to already know to type a non-`meta.public` URL (e.g. `/organizations`) by hand, which forces the guard past the early return.

Confirmed live against a self-hosted Cloudflare Access deployment: repeated navigation to `/` and to `/login` produced zero `GET /api/v1/auth/cf-session` requests in the server log; navigating directly to `/organizations` produced one immediately, and login succeeded from there.

The same bug also affects subdomain mode's `/organizations` special case, which sat before the old probe location too.

## Fix
Hoist the CF-probe block to run unconditionally near the top of the guard (after the SSO hash-token handling, before the subdomain `/organizations` special case and the `meta.public` check), gated only on `!getApiToken() && !cfTried` as before — so `getApiToken()` is accurate by the time any route-type decision is made. No change needed in `Login.vue`: its own `onMounted` already redirects an already-authenticated visitor away from the form once a token exists.

## Verification
`web/test/router.test.js` — the guard body transplanted into `vue-router` under `createMemoryHistory`, with a fake `cfSession()`. Covers: minting on `/`, on `/login`, on `/<org>/login`, on subdomain-mode `/organizations`; the probe still fires at most once per page load; and a deployment not behind Cloudflare Access degrades to normal (no token, public route renders).

unidoc-ahall:
Three fixes, one PR — the last of them (router ordering) is the one that was actively blocking login on a live deployment.

## 1. router.js: CF Access probe never ran on the routes a visitor actually lands on (Closes #222)
Every entry point a fresh, unauthenticated visitor hits — `/`, `/login`, `/<org>/login` — is `meta.public`, and the guard's CF-session probe sat *after* the `meta.public` early return. A Cloudflare-Access-authenticated visitor with no local token could never get logged in through normal navigation; the only way in was to already know to type a non-public URL (`/organizations`) by hand. Confirmed live: zero `/api/v1/auth/cf-session` requests navigating `/`/`/login` repeatedly; one immediately on `/organizations`, which then logged in successfully.

Fix hoists the probe above every route-type decision. `Login.vue` needs no change — it already redirects an authenticated visitor off the form on mount. Regression test transplants the real guard body into `vue-router` (`web/test/router.test.js`, 6 cases, including the subdomain-mode `/organizations` variant of the same bug).

## 2. Cloudflare Access: verify RS256, not just ES256 (Closes #221)
The verifier parsed EC keys only, so a team whose JWKS signs with RS256 had every CF Access token rejected as "unknown key ID." Now parses both and dispatches on the JWT `alg`, cross-checked against the resolved key type. Unit tests mint RS256 + ES256 against an httptest JWKS with both key types.

## 3. FreeBSD: rc.d rewrite + `just build-freebsd`
Follow-up to #188. rc.d reworked for the real deploy layout (`/opt/isms` binary + env file), `REQUIRE: cleanvar DAEMON NETWORKING`, refuses to start on an unreadable env file, logs to `/var/log/isms.log`. `just build-freebsd [arch]` builds the frontend then cross-compiles a static FreeBSD binary.

## Verification
`just build-go`, `just test-go`, `just build-web`, `just build-freebsd`, `npm test` (web) all clean. `sh -n` on the rc.d script passes. No schema or migration changes.


unidoc-alip (changes requested):
> **Verdict: REQUEST_CHANGES** — 2 must-fix, 3 should-fix, 2 nits

Three independent fixes in one PR: the SPA router guard's Cloudflare Access probe hoisted above the `meta.public` early return (#222), the CF Access JWT verifier taught to handle RS256 as well as ES256 (#221), and a FreeBSD rc.d rewrite plus a `just build-freebsd` target.

The two CF Access fixes are correct and well-targeted. The router hoist is the right shape — the probe genuinely could not run on any entry point a fresh visitor lands on, and the transplanted guard body confirms the new ordering resolves all four cases. The RS256 verifier is careful: it dispatches on the JWT `alg` and cross-checks it against the resolved key type, so widening to RSA does not open an alg-confusion hole.

What the review found sits mostly around the edges of those fixes. The FreeBSD commit changes the deploy layout without touching the install guide that ships in the same release tarball, leaving an operator following the shipped docs with a service that cannot start and a documented `sysrc` knob that is now a silent no-op. The verifier's new unknown-`kid` force-refresh is unthrottled and runs under the exclusive write lock on an unauthenticated code path. The verifier's new unknown-`kid` force-refresh is unthrottled and runs under the exclusive write lock on an unauthenticated code path — a PoC measures a legitimate CF Access verification going from 8ms to 601ms behind eight attacker requests. And the new web regression test never executes in CI.

Findings are keyed `F1`, `F2`, … rather than `#1`, `#2` — GitHub would autolink the latter to unrelated issues. F2 ships with a runnable PoC, reproduced in full in an appendix at the end so you can drop it in and verify it yourself; F1, F3, F5 and F7 are source-read only and say so.

---

### F1 [must-fix] `contrib/freebsd/README.md` — install guide not updated for the `/opt/isms` layout

The rc.d rewrite changes every path and drops a documented rc.conf knob, but `contrib/freebsd/README.md` is untouched by this PR. That file is not incidental documentation: `.goreleaser.yaml` lines 53–54 ship it inside every release tarball as `FREEBSD.md`, so it is the install guide an operator actually follows.

After this PR the guide is wrong in four places:

- `install -m 755 isms /usr/local/bin/isms` — the script now hardcodes `command="/opt/isms/bin/isms"`. `service isms start` fails on a missing binary.
- `install -m 600 -o root -g wheel isms.env.sample /usr/local/etc/isms.env` — the default is now `/opt/isms/etc/server.env`, so the new `if [ ! -r "$isms_env_file" ]; then err 1` guard fires and the service refuses to start. Following the guide verbatim produces a service that cannot come up at all.
- The rc.conf table documents `isms_args` (default `server serve`), and the "bind to loopback behind a reverse proxy" recipe tells the operator to run `sysrc isms_args="server serve --addr 127.0.0.1:8080"`. `isms_args` no longer exists — `isms_start` hardcodes `"$command" server serve`. That is the one failure here that is silent rather than loud: the operator sets the variable, restarts, sees the service running, and believes they are bound to loopback while the listener is still on all interfaces. On a jail host without an external firewall that exposes the app directly.
- "Logs go to syslog (tag `isms`) via `daemon(8) -S`" — output now goes to `/var/log/isms.log` via `-o`, and the new `isms_logfile` knob is undocumented.

There is also a security note that got orphaned. The guide's env-file step carries an explicit warning — install `root:wheel`, "NOT owned by the isms user: the rc.d script sources this file as root (before dropping privileges), so an env file writable by isms would be a local root-escalation path." `isms_start` still sources the env file as root (privileges are dropped only later, by `daemon -u`), so the hazard is unchanged — but the mitigation is now attached to a path the script no longer reads, and `/opt/isms/etc/server.env` has no documented ownership at all.

Either update the README to the new layout (paths, `isms_logfile`, the ownership warning moved to `/opt/isms/etc/server.env`), or keep `isms_args` and the `/usr/local` defaults as overridable rc.conf variables so both layouts work. If `isms_args` is dropped deliberately, drop it from the README table and replace the loopback recipe with whatever the supported mechanism now is — an operator silently failing to restrict a bind address is the worst outcome of the three.

Confirmed by source read plus `.goreleaser.yaml` archive contents; not reproduced on a live FreeBSD host.

---

### F2 [must-fix] `internal/isms/api/api_cf_jwt.go` line 186 — unknown `kid` forces an unthrottled JWKS fetch under the exclusive write lock

The new key-rotation handling is the right idea, but it has no cooldown of its own:

```go
if !ok {
    if err := c.fetchKeys(true); err != nil {
```

`force` bypasses the 5-minute throttle unconditionally, and `fetchKeys` holds `c.mu.Lock()` — the exclusive write lock — for the entire outbound HTTP round-trip to Cloudflare (bounded by the new 10s client timeout, which is itself an improvement over `master`'s unbounded `http.Get`).

The path is reachable by an unauthenticated remote. `AuthMiddleware` is installed globally (`server.go` line 386), and its Cloudflare branch (`middleware.go` line 230) verifies `Cf-Access-Jwt-Assertion` on any `/api/v1` request that carries a `Cf-Access-Authenticated-User-Email` header and no bearer token. `authLimit` is a per-route middleware attached only to the `/auth/*` handlers (`server.go` lines 464–472), so that branch has no rate limiter at all. Each request bearing a JWT with a random `kid` therefore converts into one outbound fetch to Cloudflare, serialized behind the same mutex every legitimate `VerifyJWT` needs.

`master` returned `unknown key ID` immediately on a `kid` miss, so this lever is new in this PR.

**PoC — reproduced in full in the appendix at the end of this review.** Save it as `internal/isms/api/poc_223_cf_kid_flood_test.go` (it is in package `api` and reuses the helpers your `api_cf_jwt_test.go` already defines — `b64`, `baseClaims`, `signRS256`, `rsaKid`, `ecKid`, `testTeamDomain`, `testAudience`), then run:

```
go test ./internal/isms/api -run TestPoC223 -v
```

It drives the real `cfKeyCache` against an `httptest` JWKS with a 200ms per-response delay, warms the cache, then fires 8 concurrent unknown-`kid` lookups while timing one legitimate RS256 verification against a known, cached `kid`. Three variants share the harness: `base` (master's kid-miss path), `head` (lines 183–192 transplanted verbatim), and `fixed` (head plus a 30s cooldown on the forced refresh).

```
base  | forced JWKS fetches from 8 unknown-kid requests: 0 | legit verify latency: 8.14ms
head  | forced JWKS fetches from 8 unknown-kid requests: 8 | legit verify latency: 600.74ms
fixed | forced JWKS fetches from 8 unknown-kid requests: 1 | legit verify latency: 185.41ms
```

Attacker requests map 1:1 onto outbound fetches, and a legitimate verification against an already-cached key — which should touch no network at all — pays a 74x latency penalty queueing behind them. Eight concurrent requests is nothing; the effect is bounded only by how fast an attacker can send, and it lands squarely on the login path this PR exists to repair.

Suggested fix — give the forced refresh its own cooldown:

```go
c.mu.RLock()
lastForced := c.lastForced
c.mu.RUnlock()
if !ok && time.Since(lastForced) > 30*time.Second {
    c.mu.Lock()
    c.lastForced = time.Now()
    c.mu.Unlock()
    if err := c.fetchKeys(true); err != nil { ... }
}
```

A `singleflight.Group` around the fetch would additionally collapse the concurrent case down to one in-flight request. The throttle check itself can move under `RLock` so the warm path never takes the write lock at all — that part is pre-existing (`master` had the same always-write-lock shape, `middleware.go`'s old `fetchKeys`), but the same change covers both.

Caveats on the PoC: the 200ms delay stands in for the CF round-trip, so the absolute latencies are synthetic — the real magnitude scales with the actual RTT to `cloudflareaccess.com` plus its 10s timeout ceiling, not with this number. The PoC calls `VerifyJWT` directly rather than through Echo, so the middleware reachability above is source-read, not exercised end to end.

---

### F3 [should-fix] `contrib/freebsd/rc.d/isms` line 33 — lost `daemon -r`, so a crashed process is no longer restarted

`master`'s invocation was `daemon -f -r -S -T ${name} ...`. The rewrite drops `-r`:

```sh
/usr/sbin/daemon \
        -p "$pidfile" \
        -o "$isms_logfile" \
        -u "$isms_user" \
        "$command" \
        server serve
```

`-r` is what made `daemon(8)` supervise and restart the child. Without it, any `isms server serve` exit — a panic, an OOM kill, a transient DB failure at startup — leaves the service permanently down until someone notices. The commit message describes the rc.d changes as a rework for the deploy layout and does not mention giving up supervision, so this reads as unintentional. Adding `-r` back is a one-line change and composes fine with `-o`.

Source-read only; not exercised on FreeBSD.

Related, lower confidence: the rewrite also drops `master`'s `install -o "${isms_user}" -g wheel -m 644 /dev/null "${pidfile}"`, whose comment said "daemon writes the pidfile as isms_user — make sure it can." If that comment was accurate, `/var/run/isms.pid` is no longer creatable and start fails. I could not confirm the privilege ordering inside `daemon(8)` from here — worth a check on a real host before merge, since it would be a hard start failure.

---

### F4 [should-fix] `web/test/router.test.js` — the regression test never runs in CI, and cannot catch the regression it was written for

Two things combine to make this test approximately decorative:

1. `.github/workflows/tests.yml` has exactly two jobs, `go-unit` and `integration`. The integration job runs `cd web && npm ci && npm run build` — it builds the frontend and never runs `npm test`. `web/package.json` does define `"test": "node --test"`, and it works: `cd web && npm test` passes 36 tests locally on Node 25, including the 6 new ones. There is already a pre-existing web test suite, so this is not a "no runner exists" situation — the runner exists and CI has simply never been wired to it.

2. The guard body in the test is a *copy* of `router.js`'s `beforeEach`, not an import of it. The header comment is upfront about this, and the constraint is real (the real router calls `createWebHistory()` and imports `.vue` components at module scope). But it means a future edit that reorders the real guard leaves this test green — the copy stays correct on its own terms. The test documents the intended ordering; it does not enforce it.

Wiring `npm test` into CI fixes the first half and is worth doing regardless of this PR:

```yaml
      - name: Web unit tests
        run: cd web && npm test
```

For the second half, the usual escape is to extract the guard body into an exported factory in `web/src/` that both `router.js` and the test consume — `export function makeAuthGuard({ api, getApiToken, setApiToken, orgFromSubdomain, storage })` — so the test exercises the shipped code path. That is a larger change and reasonable to defer, but the copy's limits are worth stating in the header comment alongside the rationale that is already there.

---

### F5 [should-fix] `web/src/router.js` line 163 — every public route now blocks first paint on an un-timed `cf-session` round-trip

Hoisting the probe is correct, but it widens what the probe can hold up. Previously the guard returned from the `meta.public` branch without any network I/O, so `/`, `/login`, `/signup`, `/forgot-password`, `/verify-email` and `/verify-email-change` rendered without waiting on anything. Now the first navigation of every page load awaits `api.cfSession()` before any of them resolve.

`cfSession` itself is well-behaved — it swallows every failure and returns `null` (`web/src/api.js` lines 188–196), so there is no risk of an unhandled rejection aborting the navigation. But `safeFetch` sets no timeout (`api.js` lines 72–80), so the await is bounded only by the browser's own connection timeout. On a deployment where `/api/v1/auth/cf-session` is slow or hanging — an overloaded origin, a proxy black-holing the request — the marketing landing page and the login form now show nothing until it resolves, where before they rendered immediately.

The mechanism is pre-existing; what this PR changes is the blast radius, extending it from authenticated routes to every public entry point. That is an acceptable trade for the fix, but it is not mentioned in the PR description and it is cheap to bound:

```js
if (!getApiToken() && !cfTried) {
  cfTried = true
  const cf = await Promise.race([
    api.cfSession(),
    new Promise(r => setTimeout(() => r(null), 3000)),
  ])
  ...
```

Source-read only — the harness has no notion of paint, so no user-visible timing claim here comes from running anything.

---

### F6 [nit] `internal/isms/api/api_cf_jwt.go` line 100 — `fetchKeys` ignores the HTTP status code

```go
resp, err := c.httpClient.Get(c.certsURL)
if err != nil {
    return err
}
defer resp.Body.Close()

var jwks cfJWKS
if err := json.NewDecoder(resp.Body).Decode(&jwks); err != nil {
```

Any response whose body happens to be valid JSON is accepted regardless of status. A Cloudflare 5xx or a captive-portal-style error page returning `{"error":"..."}` decodes cleanly into a `cfJWKS` with zero keys, and the function then commits `c.keys = keys` — wiping a previously good cache and stamping `lastFetch`. It self-heals (the new `len(c.keys) > 0` throttle condition means an empty cache always refetches), but until CF recovers, every request also does an unthrottled fetch under the write lock, which feeds straight back into F2. Cheap guard:

```go
if resp.StatusCode != http.StatusOK {
    return fmt.Errorf("CF Access certs endpoint returned %s", resp.Status)
}
```

Not worth wiping the cache on a failed refresh either — assigning into a local and only swapping on success would keep the last-known-good keys serving.

---

### F7 [nit] `contrib/freebsd/rc.d/isms` lines 22–37 — indentation switches to spaces mid-file, and the usage header was deleted

The new `isms_start` body is indented with 8 spaces while the rest of the file (and the `: ${isms_enable:="NO"}` block right above it) uses tabs. Cosmetic, but it makes the diff noisier than the change.

Separately, the rewrite deletes the whole header comment block — the copy-to-`/usr/local/etc/rc.d`, `sysrc isms_enable=YES` usage note and the rc.conf variable table. Given F1, that header was the one place the new paths and the new `isms_logfile` knob could have been recorded next to the code that defines them. Worth restoring in updated form.

---

- **"the CF-session probe sat after the `meta.public` early return"** — **confirmed.** `git show master:web/src/router.js` has the probe inside the `if (!getApiToken())` block at line 171, well below the `if (to.meta.public)` return. All of `/`, `/login`, `/signup`, `/forgot-password`, `/verify-email`, `/verify-email-change` and `/:org/login` carry `meta: { public: true }`.
- **"`Login.vue` needs no change — it already redirects an authenticated visitor off the form on mount"** — **confirmed.** `web/src/views/Login.vue` lines 465–472: `if (getApiToken()) { await api.getMe(); await redirectAfterLogin(); return }`, after the callback-token branch and after `orgSlug` is resolved, so a CF-minted token on `/login` routes into the org rather than sitting on the form.
- **"the subdomain-mode `/organizations` variant of the same bug"** — **confirmed.** That special case sits at line 174, above the `meta.public` check but below the old probe location, so it was equally unreachable by the probe.
- **"Regression test transplants the real guard body into `vue-router` (`web/test/router.test.js`, 6 cases)"** — **partially confirmed.** Six cases, and the transplant is faithful to the shipped guard. But "transplants" is doing real work here: it is a copy, not an import, so it cannot fail on a future reordering of `router.js`, and CI never executes it. See F4.
- **"The verifier parsed EC keys only […] Now parses both and dispatches on the JWT `alg`, cross-checked against the resolved key type"** — **confirmed.** `master`'s `fetchKeys` had `if k.Kty != "EC" { continue }`; the new `switch k.Kty` handles `RSA` and `EC`, and the `switch header.Alg` arms each assert the matching key pointer is non-nil before verifying.
- **"Unit tests mint RS256 + ES256 against an httptest JWKS with both key types"** — **confirmed** by file inspection and by the passing `go-unit` CI job.
- **"Follow-up to #188"** — **confirmed** as a reference; note #188 is an issue (closed), not a PR, so there is no diff to compare against. The rc.d script it produced landed in `97ddfb6` (#189), which is what the rewrite replaces.
- **"`just build-go`, `just test-go`, `just build-web`, `just build-freebsd`, `npm test` (web) all clean"** — **partially confirmed.** `npm test` reproduced clean locally (36 pass, Node 25). `go-unit` and `integration` both pass in CI. The Go and web builds are exercised by CI. `just build-freebsd` is not — no CI job cross-compiles for FreeBSD, so that one rests on the author's local run. The description correctly presents these as local verification and makes no false CI claim.
- **"No schema or migration changes"** — **confirmed.** No files under `migrations/` in the diff.
- **"refuses to start on an unreadable env file"** — **confirmed** (`err 1` guard), and `sh -n contrib/freebsd/rc.d/isms` passes. But see F1: with the README unchanged, that guard is what an operator following the shipped install steps will hit.

---

## Appendix — the F2 proof of concept

Save as `internal/isms/api/poc_223_cf_kid_flood_test.go` and run `go test ./internal/isms/api -run TestPoC223 -v`. It compiles against the branch as-is: it lives in package `api` and reuses the helpers already in `api_cf_jwt_test.go` (`b64`, `baseClaims`, `signRS256`, and the `rsaKid` / `ecKid` / `testTeamDomain` / `testAudience` constants), so nothing else needs changing. The rest of the package's tests still pass alongside it.

```go
package api

// PoC for PR #223 review finding F2.
//
// api_cf_jwt.go's new unknown-kid handling calls fetchKeys(true), which bypasses
// the 5-minute throttle entirely and holds cfKeyCache.mu (the EXCLUSIVE write
// lock) for the whole outbound JWKS round-trip. VerifyJWT is reachable by an
// unauthenticated remote via middleware.go:230 (any /api/v1 path carrying a
// Cf-Access-Authenticated-User-Email header and no bearer token; that middleware
// has no rate limiter). So each attacker request with a random kid = one outbound
// fetch, serialized against every legitimate CF Access verification.
//
// Three variants side by side, all against the SAME slow httptest JWKS:
//
//	base  — master's behaviour: kid miss returns immediately, no refetch.
//	head  — this PR: kid miss calls fetchKeys(true), unthrottled.
//	fixed — head plus a cooldown on the forced refresh (the suggested fix).
//
// Measured: JWKS fetches triggered by a flood of unknown-kid tokens, and the
// wall-clock latency a legitimate RS256 verification pays while that flood runs.
//
// Run: go test ./internal/isms/api -run TestPoC223 -v

import (
	"crypto/ecdsa"
	"crypto/elliptic"
	"crypto/rand"
	"crypto/rsa"
	"encoding/json"
	"fmt"
	"math/big"
	"net/http"
	"net/http/httptest"
	"sync"
	"sync/atomic"
	"testing"
	"time"
)

const (
	pocJWKSDelay = 200 * time.Millisecond // stand-in for the CF round-trip
	pocFloodSize = 8                      // attacker requests with unknown kids
)

// slowJWKSServer is jwksServer with a per-response delay and a request counter,
// so the PoC can both count forced fetches and make the lock-hold observable.
func slowJWKSServer(t *testing.T, rsaPub *rsa.PublicKey, ecPub *ecdsa.PublicKey, hits *atomic.Int64) *httptest.Server {
	t.Helper()
	jwks := cfJWKS{Keys: []cfJWK{
		{Kid: rsaKid, Kty: "RSA", Alg: "RS256",
			N: b64(rsaPub.N.Bytes()),
			E: b64(big.NewInt(int64(rsaPub.E)).Bytes())},
		{Kid: ecKid, Kty: "EC", Alg: "ES256", Crv: "P-256",
			X: b64(ecPub.X.FillBytes(make([]byte, 32))),
			Y: b64(ecPub.Y.FillBytes(make([]byte, 32)))},
	}}
	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		hits.Add(1)
		time.Sleep(pocJWKSDelay)
		_ = json.NewEncoder(w).Encode(jwks)
	}))
}

func pocCache(t *testing.T) (*cfKeyCache, *rsa.PrivateKey, *atomic.Int64) {
	t.Helper()
	rsaPriv, err := rsa.GenerateKey(rand.Reader, 2048)
	if err != nil {
		t.Fatal(err)
	}
	ecPriv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
	if err != nil {
		t.Fatal(err)
	}
	hits := &atomic.Int64{}
	srv := slowJWKSServer(t, &rsaPriv.PublicKey, &ecPriv.PublicKey, hits)
	t.Cleanup(srv.Close)
	c := newCFKeyCache(testTeamDomain, testAudience)
	c.certsURL = srv.URL
	return c, rsaPriv, hits
}

// --- the three kid-miss policies -------------------------------------------

// baseKidMiss reproduces master: look the kid up, give up if it's absent.
// (git show master:internal/isms/api/middleware.go — VerifyJWT returned
// "unknown key ID: %s" straight after the RLock miss.)
func baseKidMiss(c *cfKeyCache, kid string) {
	c.mu.RLock()
	_, ok := c.keys[kid]
	c.mu.RUnlock()
	_ = ok
}

// headKidMiss is api_cf_jwt.go lines 183-192, transplanted verbatim.
func headKidMiss(c *cfKeyCache, kid string) {
	c.mu.RLock()
	_, ok := c.keys[kid]
	c.mu.RUnlock()
	if !ok {
		_ = c.fetchKeys(true)
		c.mu.RLock()
		_, ok = c.keys[kid]
		c.mu.RUnlock()
	}
	_ = ok
}

// fixedKidMiss is headKidMiss with the suggested cooldown on the forced refresh.
type pocCooldown struct {
	mu         sync.Mutex
	lastForced time.Time
}

func (p *pocCooldown) kidMiss(c *cfKeyCache, kid string) {
	c.mu.RLock()
	_, ok := c.keys[kid]
	c.mu.RUnlock()
	if !ok {
		p.mu.Lock()
		allowed := time.Since(p.lastForced) > 30*time.Second
		if allowed {
			p.lastForced = time.Now()
		}
		p.mu.Unlock()
		if allowed {
			_ = c.fetchKeys(true)
			c.mu.RLock()
			_, ok = c.keys[kid]
			c.mu.RUnlock()
		}
	}
	_ = ok
}

// --- the measurement --------------------------------------------------------

type pocResult struct {
	forcedFetches int64
	legitLatency  time.Duration
}

// runFlood warms the cache, then fires pocFloodSize concurrent unknown-kid
// lookups while timing one legitimate RS256 verification against a known kid.
func runFlood(t *testing.T, kidMiss func(*cfKeyCache, string)) pocResult {
	t.Helper()
	c, rsaPriv, hits := pocCache(t)

	// Warm the cache the way a first real request would, then discount that fetch.
	if _, err := c.VerifyJWT(signRS256(t, rsaPriv, rsaKid, baseClaims())); err != nil {
		t.Fatalf("warm-up verify should succeed: %v", err)
	}
	warm := hits.Load()

	var wg sync.WaitGroup
	start := make(chan struct{})
	for i := 0; i < pocFloodSize; i++ {
		wg.Add(1)
		go func(i int) {
			defer wg.Done()
			<-start
			kidMiss(c, fmt.Sprintf("attacker-kid-%d", i))
		}(i)
	}

	legit := make(chan time.Duration, 1)
	wg.Add(1)
	go func() {
		defer wg.Done()
		<-start
		// Give the flood a head start so the legit request arrives while the
		// attacker traffic is already holding the lock.
		time.Sleep(20 * time.Millisecond)
		t0 := time.Now()
		if _, err := c.VerifyJWT(signRS256(t, rsaPriv, rsaKid, baseClaims())); err != nil {
			t.Errorf("legitimate RS256 verify failed: %v", err)
		}
		legit <- time.Since(t0)
	}()

	close(start)
	wg.Wait()
	return pocResult{forcedFetches: hits.Load() - warm, legitLatency: <-legit}
}

func TestPoC223_UnknownKidFloodStallsLegitimateVerification(t *testing.T) {
	base := runFlood(t, baseKidMiss)
	head := runFlood(t, headKidMiss)
	cd := &pocCooldown{}
	fixed := runFlood(t, cd.kidMiss)

	for _, r := range []struct {
		name string
		res  pocResult
	}{{"base", base}, {"head", head}, {"fixed", fixed}} {
		t.Logf("%-5s | forced JWKS fetches from %d unknown-kid requests: %d | legit verify latency: %v",
			r.name, pocFloodSize, r.res.forcedFetches, r.res.legitLatency)
	}

	// One joined outcome string per variant, so a mismatch prints in full.
	got := fmt.Sprintf("base=%d head=%d fixed=%d", base.forcedFetches, head.forcedFetches, fixed.forcedFetches)
	want := fmt.Sprintf("base=0 head=%d fixed=1", pocFloodSize)
	if got != want {
		t.Errorf("forced-fetch counts: got %q, want %q", got, want)
	}

	// The lock is held across each forced fetch, so the legit request queues.
	if head.legitLatency < pocJWKSDelay {
		t.Errorf("head legit latency %v: expected the legit verify to queue behind at least one %v fetch",
			head.legitLatency, pocJWKSDelay)
	}
	if base.legitLatency > pocJWKSDelay {
		t.Errorf("base legit latency %v: master should not stall at all (cache is warm)", base.legitLatency)
	}
	if fixed.legitLatency >= head.legitLatency {
		t.Errorf("fixed legit latency %v should be well under head's %v", fixed.legitLatency, head.legitLatency)
	}
}
```

unidoc-ahall:
Thanks — this is an unusually thorough review, and F2's PoC made verification straightforward rather than a matter of trust. All seven confirmed real, all fixed.

**F1 (must-fix).** Confirmed all four stale points and fixed the README (paths, the restored `isms_args` recipe, `isms_logfile` replacing the syslog claim, the ownership warning moved to `/opt/isms/etc/server.env`). `isms.env.sample`'s header updated too.

**F2 (must-fix).** Added `tryForceRefresh` with its own 30s cooldown on a dedicated mutex (so checking it never contends with a normal verification's `RLock`) — at most one forced fetch per window regardless of flood size. I adapted your PoC to call the actual shipped `tryForceRefresh` rather than reimplementing the suggested fix, so it's testing the real code path. One honest note: in my adapted harness the exact latency multiplier didn't reproduce as cleanly as your original numbers — `sync.RWMutex`'s exact queueing order turned out to be sensitive to goroutine-launch timing in ways I couldn't fully pin down, and a few runs showed the "vulnerable" and "fixed" cases landing closer together than expected. The **forced-fetch count** (N → 1, deterministic across every run I did) is what the test asserts on; I logged the latencies for context instead of asserting a specific ratio, rather than tune the harness until a number happened to pass. The fix's actual guarantee — bounding outbound Cloudflare calls and exclusive-lock holds to at most one per 30s under flood — holds regardless.

**F6.** Fixed alongside F2 since both touch `fetchKeys`: non-200 responses no longer wipe the cache, and a failed refresh leaves the last-known-good keys serving.

**F3.** `daemon -r` and the pidfile-ownership install step are both back.

**F4.** Wired `npm test` into the integration job. Took your suggestion on the copy-vs-import limitation — documented it explicitly in the test header rather than fixing it now; the factory-extraction refactor is deliberately deferred.

**F5.** Bounded the probe with `Promise.race` at 3s, plus a test with an injected short timeout so the hang case doesn't cost the suite 3 real seconds.

**F7.** rc.d body back to tabs; the header comment restored in updated form.

One more thing I caught while fixing F4: the test file's header comment named a real deployment while describing where this was reported live. Scrubbed it to a neutral description and grepped the tree to confirm nothing else leaked — that should never have been committed in the first place.

`gofmt`, `go build`, `go vet`, `go test ./internal/isms/api/...` (including the adapted PoC), `sh -n` on the rc.d script, `just build-freebsd`, `just build-web`, and `npm test` (37 pass, 2 new) are all clean.

unidoc-alip (approved):
> **Verdict: APPROVE** — all 7 findings verified fixed at `e8a2290`

Re-reviewed at `e8a229065d9d13b8d21e54dd86965f391837ab94`. Every fix was checked against the shipped code rather than the commit message, and the PoC was re-run against the real `tryForceRefresh`. All seven land. Thanks especially for adapting the PoC to call the actual shipped function instead of a copy of my suggestion — that is the right instinct, and it is what let me verify F2 in one command instead of by inspection.

**F2 — verified, and your caveat was the correct call.** I re-ran your adapted PoC five times on the merge head:

```
run1  vulnerable: 8 fetches / 207.44ms | fixed: 1 fetch / 183.26ms
run2  vulnerable: 8 fetches / 611.69ms | fixed: 1 fetch / 185.89ms
run3  vulnerable: 8 fetches / 386.38ms | fixed: 1 fetch / 183.31ms
run4  vulnerable: 8 fetches / 999.42ms | fixed: 1 fetch / 183.57ms
run5  vulnerable: 8 fetches / 388.55ms | fixed: 1 fetch / 182.21ms
```

Your read is right and worth recording: the forced-fetch count is dead stable at 0/8/1 across every run, while the vulnerable variant's latency swings between 207ms and 999ms — run 1 lands close enough to `fixed` that a ratio assertion would flake. Asserting on the count and logging the latency is the honest choice, and better than tuning the harness until a number passes. `fixed` is pinned at ~183ms, which is exactly the one permitted forced fetch the legitimate verify queues behind; under a sustained flood that is one per 30s window rather than one per request, which is the property that actually matters.

`tryForceRefresh` itself is correct: the cooldown is claimed and stamped under `forceMu` *before* `fetchKeys` is entered, so concurrent callers can't both slip through, and `forceMu` never contends with `c.mu`. Gating the re-lookup on `ran` is a nice detail — a cooling-down caller now falls through to the accurate "unknown key id" error with the `iss` vs `team_domain` hint rather than silently re-reading an unchanged map.

**F1 — verified.** All four stale points fixed, `install -d /opt/isms/{bin,etc}` added, and the root-escalation warning moved intact to `/opt/isms/etc/server.env`. `isms_args` is genuinely restored in the script (`command_args` interpolates `${isms_args}`), so the README's loopback-bind recipe is live again rather than a documented no-op — that was the failure mode I cared most about. `isms.env.sample`'s header now carries the ownership requirement too, which it didn't before.

**F3 — verified.** `-r` is back, and the `install -o ${isms_user} ... ${pidfile}` step with it. Reverting to master's `command=/usr/sbin/daemon` + explicit `procname` shape is better than the custom `start_cmd` in more than one way: rc.subr's default `stop`/`status` now resolve against `procname` the way the rest of the file expects, and the env-file readability guard survives as a `start_precmd`.

**F4 — verified.** `npm test` is wired into the integration job and ran green on this push (both CI jobs pass at `e8a2290`), so the regression test is now actually executed rather than merely present. `npm test` locally: 37 pass, 0 fail. The copy-vs-import limitation is documented plainly in the header, which is what I asked for; deferring the factory extraction is fine.

**F5 — verified.** `Promise.race` at 3s, and the new hang test injects a 20ms bound so the suite doesn't pay for it. It asserts the two things that matter — a hang is not mistaken for a successful login, and the public route still renders.

**F6 — verified.** Non-200 rejected explicitly, and the success-only cache swap means a transient CF outage no longer wipes a good key set.

**F7 — verified.** Tabs throughout, header restored with the new defaults including `isms_logfile` and `isms_args`.

**On the scrubbed deployment name** — good catch, and agreed it shouldn't have been committed. One thing to handle at merge time, though: scrubbing it in `e8a2290` doesn't remove it from the branch, only from the tip. It is still in `c9e4b81`'s copy of `web/test/router.test.js`:

```
$ git grep -l nanitor c9e4b81
c9e4b81:web/test/router.test.js
$ git grep -l nanitor e8a2290    # clean
```

A merge commit or a rebase that preserves the four commits carries that blob into `master`'s history permanently. Squash-merging this PR keeps it out. Worth deciding deliberately rather than by whichever button gets clicked.

---

## Non-blocking observations

None of these hold up the merge; noting them so they're not lost.

- `contrib/freebsd/README.md` line 52 still runs the migrate step as bare `isms server migrate`. That worked when the binary was at `/usr/local/bin`; at `/opt/isms/bin` it isn't in root's default PATH, so the command as written will not resolve. `/opt/isms/bin/isms server migrate` fixes it.
- The rc.d `command_args` now passes both `-f` and `-o`. That mirrors master's established `-f -S` pattern and your comment explains the intent correctly, but I can't exercise `daemon(8)` from here — worth one `service isms start` smoke test on a real host before cutting a release, together with the pidfile-ownership behaviour I flagged as low-confidence last round.
- Pre-existing, unchanged by this PR, and out of scope — but adjacent to F6: `VerifyJWT` returns `fetching CF Access keys: ...` whenever `fetchKeys(false)` errors, before it ever consults the cache. So once the 5-minute throttle lapses during a CF certs outage, every verification fails even though good keys are still cached, and each request also pays a fetch under the write lock. F6 preserved the cache; letting the warm path serve from it on a failed refresh would be the natural follow-up.

---

## Verification performed

At `e8a2290`, in a detached worktree so the review didn't touch the branch:

- `go vet ./internal/isms/api/` — clean; `gofmt -l` — no output.
- `go test ./internal/isms/api/` — pass, including the adapted PoC.
- `go test ./internal/isms/api -run TestPoC223 -v` — 5 repeat runs, numbers above.
- `cd web && npm test` — 37 pass, 0 fail (Node 25).
- `sh -n contrib/freebsd/rc.d/isms` — pass.
- `gh pr checks 223` — `go-unit` and `integration` both pass at `e8a2290`; the integration job now includes the web unit tests.
- Read every changed file at head rather than relying on the diff, and confirmed the PoC's `fixedKidMiss` calls the shipped `c.tryForceRefresh()`.

Not verified: FreeBSD runtime behaviour (no host available) and `just build-freebsd`, which no CI job exercises.

PR #224, opened 2026-08-25
fix: single-tenant UI fixes + normalize example placeholder values

unidoc-ahall:
Two independent fixes, one PR.

## 1. Single-tenant deployment UI fixes
Three small issues on a self-hosted single-org deployment (`ISMS_SUBDOMAIN_ROUTING=0`, signup disabled):

- `subdomainRoutingEnabled` promoted from a plain module variable to a `ref` (exported via a new `subdomainRouting()` getter), so `App.vue` can gate "Switch organization" / "Create new organization" on it — previously reachable with nothing to switch to and no way to create another org.
- `Landing.vue` seeded `brandName` from `window.location.hostname` instead of the literal `"isms.sh"`, so a self-hosted deployment doesn't flash the wrong brand name before `/config` resolves.
- `.doc-prose li` spacing tightened, and `Documents.vue`'s scoped styles get their own loose-list collapse rule (`:deep(li > p) { margin: 0 }`) — its scoped `:deep(p)` rule was higher-specificity than the global collapse rule, so markdown lists with blank lines between items never got it applied there specifically.

## 2. Normalize example placeholder values
Comments, docs, and test fixtures had drifted onto several different example domains and org names instead of one consistent fictional convention. Normalized everything onto `acme.isms.sh` / `isms.example.com` / `audit.example.com` / `Acme Corp` — the convention several of these same files already use in their other cases. No behavior change: string literals in comments, docs prose, and test fixture values only.

## Verification
`gofmt`, `go build`, `go vet`, `go test ./...`, `just build-web`, `npm test` (web) all clean.


unidoc-ahall:
Pushed one more commit: the "Switch organization" link was gated on `subdomainRoutingEnabled` (an infra flag — can this apex serve wildcard subdomains) instead of `otherOrgs.length > 0` (the actual fact — does this user have another org to go to).

Caught this discussing a real near-term case: a self-hosted single-tenant apex can pick up a second org path-based (`/org1`, `/org2`) without ever needing subdomain routing — `org_resolver.go`'s path-based branch runs unconditionally, subdomain resolution is the only part gated by that flag. With the old condition the switcher would have stayed hidden forever on that deployment even after a real second org existed. `otherOrgs.length > 0` already updates reactively with membership and needs no config change when that day comes.

`subdomainBound` (unrelated axis — am I on a dedicated tenant subdomain like `unidoc.isms.sh`) is untouched; that one's already correct. `subdomainRoutingEnabled` still gates "Create new organization" — self-serve org creation is a SaaS-only affordance regardless of org count.

unidoc-alip (changes requested):
> **Verdict: REQUEST_CHANGES** — 0 must-fix, 2 should-fix, 1 nit (reviewed at 95390df)

Two independent changes in one PR: three single-tenant web UI fixes, and a
placeholder-value normalization sweep across comments, docs and test fixtures.

The normalization half is clean — I re-grepped the whole tree for every old
placeholder (`STS`, `CommandVector`, `stsplatform`, `sts.is`, `verkis`) and
found zero remaining occurrences, the Go changes are comments and fixture
values only, and `mail_test.go`'s leak assertion still tests what it claims
(the new operator brand `Operator Inc` is not a substring of the new tenant
brand `Acme Audit ehf`). The `:deep(p)` specificity diagnosis in the
description is also correct, and the fix is complete — `Documents.vue` is the
only component in the tree with a scoped `.doc-prose :deep(p)` override.

The UI half originally gated both the org switcher and the create-org link on
`subdomain_routing`, which is an infrastructure flag, not an org-count or
authorization one. **95390df fixes the switcher half** — see the note below;
that was the one blocking issue and it is resolved. What is left is the
create-org link, still on that same flag, and two smaller items. Findings are
keyed F1, F2, F3 (not `#1`, `#2` — GitHub would autolink those to unrelated
issues).

---

## Resolved by 95390df

The header "Switch organization" link now reads
`v-if="!subdomainBound && otherOrgs.length > 0"` — membership, which is the
right signal and needs no server change (my earlier draft of this review
wanted a new `single_org` field in `/config`; `otherOrgs.length` is better,
it is already computed and already reactive). The commit message's reasoning
matches the repo's own model: `docs/multi-tenant-routing.md` documents path
mode as a multi-org mode ("the `/organizations` picker is reachable", "the
header org-switcher shows other org memberships", "the only mode on
deployments without wildcard DNS / TLS"), and `server.go`'s comment on the
field says it is "false for demo / dev".

I re-ran the gate PoC (below) against 95390df: the switcher regression is
gone in every scenario, including localhost dev and the first-paint case.

```
                                                   base            f8791e3         95390df
path mode, MANY orgs (no wildcard DNS)             create|switch   - | -           - | switch
path mode, ONE org (the targeted deployment)       create|switch   - | -           - | -
localhost dev, MANY orgs                           create|switch   - | -           - | switch
subdomain mode apex, MANY orgs                     create|switch   create|switch   create|switch
subdomain mode, on a tenant subdomain              create| -       create| -       create| -
/config not yet resolved (first paint), MANY orgs  create|switch   - | -           - | switch
```

The harness transplants both `v-if` expressions verbatim from `App.vue` and
drives the *real* `useCurrentOrg.js` ref through the *real*
`setSubdomainRouting()` — the same call `useSession.js:181` makes from
`/config`. Caveats: it renders the gate expressions, not the app, so it says
nothing about the link *target* (`/organizations` stays URL-reachable in every
row — what these gates control is the affordance only). Harness available on
request; it has to sit under `web/` for `vue` to resolve, then
`node web/poc_224_single_org_gate.mjs`.

---

### F1 [should-fix] `web/src/App.vue` line 79 — the create-org link still rides an infra flag, and its comment now contradicts the code

95390df's rationale for leaving this one on `subdomainRoutingEnabled` is a
product decision I have no objection to — self-serve org creation as a SaaS
affordance, self-hosted operators use the CLI. Two things about how it is
expressed, though:

The comment above the line still says
`<!-- Create new org — hidden on single-org deployments -->`. That was the
original (incorrect) intent; per 95390df's own reasoning the link is hidden on
*non-subdomain-routing* deployments regardless of how many orgs they have. The
setup block below it was updated with the new rationale — this line was
missed, and it is now the misleading half of a two-place explanation.

Underneath that, `ISMS_SUBDOMAIN_ROUTING` is being used as a proxy for "this
is the hosted product". It happens to correlate today, but the two can be
decoupled by an operator who turns on wildcard DNS for a private install, and
nothing in the flag's name or its `server.go` doc comment suggests it carries
that meaning. If the policy is "self-serve org creation is SaaS-only", an
explicit capability in `/config` (`self_serve_orgs`, alongside the existing
`signup_enabled`, which already encodes an adjacent policy) says what is meant
and cannot drift. Source read only — this is a naming/semantics point, not a
behavioural one.

One pre-existing wrinkle in the same line, worth folding in while it is being
touched: it has no `!subdomainBound` guard (its header counterpart does), so
on a tenant subdomain it renders a link the `router.js` guard bounces straight
back. True on master too; visible in the PoC's fifth row.

---

### F2 [should-fix] `web/src/composables/useCurrentOrg.js` line 151 — the flag is `null` until `/config` resolves, and stays `null` if it fails

`_subdomainRoutingEnabled` starts at `ref(null)`, and the only writer is
`loadBranding()` in `useSession.js`, whose `catch` swallows a failed
`/api/v1/config` entirely. Now that a template gates on it:

- Pop-in: on a subdomain-routing deployment the create-org link is absent on
  first paint and appears when `/config` resolves, every load.
- Fail-closed on error: if `/config` 500s or is blocked, it never appears at
  all. `loadAppData()` also returns early for public routes and
  `/organizations`, so the flag is never set on those routes.

Smaller than it was before 95390df, since the switcher no longer depends on
it, and fail-closed on a hidden link is the safe direction — so not a
must-fix. But the tri-state is currently incidental rather than intended;
either seed it from the synchronous `<meta>` channel the apex host already
uses, or decide explicitly what `null` should render.

---

### F3 [nit] `web/src/views/Landing.vue` line 179 — raw hostname as the seed

`window.location.hostname` shows exactly what is in the address bar, so the
pre-`/config` flash becomes `www.isms.sh`, `localhost`, or a bare IP depending
on how the user arrived. It does fix the stated problem (no more foreign brand
name), but a neutral `'ISMS'` — the same fallback `App.vue`'s sidebar already
uses for `orgName` — would read better than a hostname, and needs no
`typeof window` guard.

---

The description still describes the f8791e3 state of the switcher gate
("`App.vue` can gate 'Switch organization' / 'Create new organization' on
it"); worth a refresh so it matches 95390df, where only the create link is
gated on that flag.

- "`subdomainRoutingEnabled` promoted to a `ref`, exported via a new
  `subdomainRouting()` getter" — **confirmed.** The mechanism works: a
  top-level `ref` binding returned from `setup` auto-unwraps in the template.
- "`orgEntryURL()`'s one internal read updated to `.value`; `useSession.js`
  (the only writer) untouched" — **confirmed.** One read at line 199,
  `setSubdomainRouting` is the sole writer, called only from
  `useSession.js:181`.
- "`Landing.vue` … doesn't flash the wrong brand name" — **confirmed**, with
  the caveat in F3 about what it flashes instead.
- "`Documents.vue`'s scoped `:deep(p)` rule was higher-specificity than the
  global collapse rule" — **confirmed.** `.doc-prose :deep(p)` compiles to
  `.doc-prose[data-v-…] p` at (0,2,1), which beats the global
  `.doc-prose li > p` at (0,1,2). `Documents.vue` is also the only component
  in the tree carrying such an override, so the fix needs no counterpart
  elsewhere.
- "`.doc-prose li` spacing tightened" — **confirmed but understated.** The
  `style.css` change (`margin-bottom` 0.375rem → 0.125rem, plus a new
  `line-height: 1.55`) is global: 16 components render `.doc-prose`, and none
  of them override `li`, so every list in the app changes rhythm, not just the
  document viewer. Presumably intended — worth saying so in the description.
- "Normalized everything onto `acme.isms.sh` / `isms.example.com` /
  `audit.example.com` / `Acme Corp`" — **confirmed.** A tree-wide
  case-insensitive grep for `STS`, `CommandVector`, `stsplatform`, `sts.is`
  and `verkis` returns nothing.
- "No behavior change: string literals in comments, docs prose, and test
  fixture values only" — **confirmed.** Go changes are comments plus table
  fixtures and their matching assertions; no test logic or code path changed.
- "`gofmt`, `go build`, `go vet`, `go test ./...`, `just build-web`,
  `npm test` (web) all clean" — **confirmed.** Reproduced locally: `gofmt -l`
  empty, `go build ./...`, `go vet ./...`, `go test ./internal/...` clean, and
  `cd web && npm test` 57 pass / 0 fail (re-run at 95390df). CI agrees —
  `go-unit` and `integration` both green on 95390df, and the `integration` job
  is the one that runs `npm test`.

unidoc-ahall:
Pushed a fix commit for F1a, F1c, F2 and F3.

**F1c** — added the missing `!subdomainBound` guard to the create-org link (confirmed the dead-link before fixing: `router.js:188` bounces `/organizations` back to `/overview`/`/login` whenever `orgFromSubdomain()` is truthy, and the header's own switcher link already had this guard, just not this one). **F1a** — comment above it now describes the current gate instead of the pre-89b78dd intent.

**F2** — you're right this is more than cosmetic: `loadAppData()` returns before calling `loadBranding()` on `/organizations` specifically (`App.vue:677`), so any session that lands there first — a plausible actual entry point, not just a config-fetch-failure edge case — never sees the create-org link for the rest of that session. Seeded `_subdomainRoutingEnabled` synchronously from a new `<meta name="isms-subdomain-routing">` tag, injected by the same `serveIndex` handler that already injects the apex host for the identical race. Added `test_index_html_injects_subdomain_routing_flag` in `test_routing.py` next to the existing apex-injection regression tests.

**F3** — `brandName` now seeds from the neutral `'ISMS'` fallback (same one `App.vue`'s `orgName` already uses) instead of the raw hostname.

**F1's other half** (subdomain_routing as a proxy for an org-creation policy) — pulled that into its own issue rather than deciding it inline: https://github.com/unidoc/isms/issues/225. Turns out the policy itself was never actually decided anywhere — the backend has no deployment-level gate at all today, any org-admin can already create more orgs via the API or by navigating to /organizations directly, so hiding this one link was discoverability-only, not access control. Filed to work out whether that should change before picking what gates the link.

unidoc-alip (approved):
> **Verdict: APPROVE** — all four findings verified addressed (re-reviewed at 1ba4b1f)

Re-verified F1a, F1c, F2 and F3 against the current head. All genuinely
addressed, and F2 is addressed in the better of the two ways I offered.

**F1c — missing `!subdomainBound` on the create-org link.** `App.vue` line 81
now reads `v-if="!subdomainBound && subdomainRoutingEnabled"`, the same guard
the header switcher carries at line 410, both fed by the single
`const subdomainBound = isSubdomainMode()` at line 544. The fifth row of my
PoC table — tenant subdomain renders a link `router.js` bounces straight back
— is closed.

**F1a — stale comment.** Now states both halves of the actual gate (no
subdomain routing, since self-serve creation is a SaaS affordance; and tenant
subdomain, since `/organizations` bounces there). No longer the misleading half
of a two-place explanation.

**F2 — the `null` tri-state.** Fixed via the synchronous `<meta>` channel
rather than by deciding what `null` renders, which is the stronger of the two.
Checked the details that could have gone wrong and none did:
`_subdomainRoutingEnabled` is now declared at `useCurrentOrg.js` line 33,
*before* the module-scope seed block that assigns it at line 53 — the
declaration move is load-bearing, not cosmetic. Unlike `apex_host` the tag is
injected unconditionally (`%t`, both `true` and `false`), so a served page
never leaves the flag unknown. And `serveIndex` handles both `/` and every
SPA-route fallback, so a direct landing on `/organizations` — the case that
made this more than a cosmetic pop-in — gets the seed too.
`setSubdomainRouting` is still the only async writer, so `/config` continues to
win later without conflicting with the seed.

This one is runtime-proven, not source-only:
`test_index_html_injects_subdomain_routing_flag` asserts the marker with no
skip branch, and it ran against a live server in CI — `integration` green on
1ba4b1f at 768 passed / 1 skipped, the single skip being the pre-existing apex
test's `apex_host`-empty guard, not this one.

**F3 — raw hostname seed.** `Landing.vue` line 179 is now `ref('ISMS')`,
matching `App.vue`'s own `orgName` fallback, and the `typeof window` guard went
away with it.

Locally on head: `gofmt -l` empty, `go build ./...` and
`go vet ./internal/isms/api` clean, `cd web && npm test` 57 pass / 0 fail.

Two observations, neither a finding. Under the Vite dev server there is no
injected meta, so the flag stays `null` and the create-org link is hidden in
dev — fail-closed, and dev is path-based regardless, so that reads as intended.
And filing F1's other half as #225 rather than settling it inline was the right
call; the issue also surfaces a fact my review did not have — there is no
backend deployment-level gate at all today, so this link was discoverability
only, never access control. That reframes the whole question usefully.

One thing left over, non-blocking and not a keyed finding: the PR description
still says "`App.vue` can gate 'Switch organization' / 'Create new
organization' on it", which describes the f8791e3 state. Since 95390df the
switcher gates on `otherOrgs.length > 0`, and as of ab2bd01 the create link
also carries `!subdomainBound`. Worth a refresh before merge so the description
matches what shipped, but not something to hold the approval on.

PR #227, opened 2026-08-25
fix(api): version-downgrade guard now floors on the approved milestone, not the current draft

### LINKED ISSUE #226: Version-downgrade guard compares against the wrong floor
## Bug

`handleUpdateDocumentContent` (`internal/isms/api/server.go`) blocked lowering a document's `version` field by comparing the requested value against the document's **current draft frontmatter value**, not the last **approved** milestone. While a document is in `draft` status, nothing depends on its frontmatter version staying monotonic against its own prior draft saves — `document_versions` milestones are only written on approve/merge/confirm. Comparing against the current draft value locked in whatever was last saved, mistake or not, as a new floor.

## Concrete case

A document goes: `0.1` (approved) → edit auto-bumps to `0.2` (draft) → user manually (mis)types `0.3` and saves → user tries to correct back to `0.2`. Blocked with "cannot lower version from 0.3 to 0.2" — even though `0.2` is still above the last accepted version (`0.1`), the only floor that should actually matter.

Live on a production self-hosted deployment: hit exactly this sequence editing `documents/agreements/data-processing-agreement.md`, had to bypass the API by editing the frontmatter directly via git.

## Fix (PR #227)

Compares against `db.LatestVersion(ctx, orgID, docID)` — the last approved milestone — via a shared `versionFloorError()` helper called from **both** `PUT .../content` and `PUT .../metadata` (the metadata endpoint, which the document header's own version field uses, was found unguarded during review and closed in the same PR). A database error other than "no milestone" is a 500, not a silent skip.

**Correction:** a document that has never been approved (no milestone yet) is unconstrained on either endpoint — this is a real widening from the old behavior, not equivalent to it. The old guard compared against the draft frontmatter value, which exists on every document regardless of approval history, so it also constrained never-approved documents; the new one has nothing to compare against until a milestone exists. (An earlier version of this issue description said this case was "same as before" — it is not; caught in review.)

unidoc-ahall:
Fixes https://github.com/unidoc/isms/issues/226.

The version-downgrade guard on `PUT /documents/:docId/content` compared a requested version against the document's current draft frontmatter value instead of the last **approved** milestone (`document_versions`, written only on merge/confirm). A draft's version field is working state until re-approved, so locking against its own prior draft saves — rather than against what was actually accepted — blocked a legitimate correction: a document approved at 0.1, auto-bumped to 0.2 on edit, manually over-bumped to 0.3 by mistake, could never be corrected back to 0.2 even though 0.2 is still above the 0.1 floor that actually matters.

Now compares against `db.LatestVersion()` via a shared `versionFloorError()` helper, called from **both** `PUT .../content` and `PUT .../metadata` — the latter is what the document header's own version field uses, and was left completely unguarded until this update (review finding F1). A database error other than "no milestone" is now a 500, not a silent skip (F2).

**A document with no approval history yet (no milestone) is unconstrained — this is a real widening from prior behavior, not "same as before"** (the earlier draft said otherwise; corrected per review finding F3). The old guard compared against the draft frontmatter, which every document has regardless of approval history, so it also constrained never-approved documents; the new one has nothing to compare against until a milestone exists.

The rejection message names the actual floor instead of just restating the two numbers.

New regression tests in `tests/test_version_downgrade_guard.py` walk: the full correction sequence (create → pin 0.1 → approve → merge → auto-bump → over-bump → correct downward → at-or-below-floor rejected), the metadata-endpoint enforcement (mutation-checked — fails without the fix), and the never-approved case being unconstrained on both writers by design.

**Verified locally**, not just claimed: `go build`/`vet`/`test` clean; full non-e2e pytest suite (691 passed, 1 skipped) and `test_e2e_browser.py` (69 passed) both green against a fresh database, run repeatedly for idempotency on a persistent stack — which also surfaced and fixed a latent idempotency gap in `test_review_branch.py::TestDocumentContentSave` unrelated to this PR's own scope but exposed by the new, real floor.

Surfaced live on a production self-hosted deployment — a user hit this exact sequence and had to bypass the API by editing the frontmatter directly via git to get unblocked.

### DIFF
--- a/internal/isms/api/server.go
+++ b/internal/isms/api/server.go
@@ -1635,9 +1635,20 @@ func (s *Server) handleUpdateDocumentContent(c echo.Context) error {
 
 	// Update metadata fields if provided
 	if req.Version != "" {
-		// Prevent version downgrade
-		if compareVersions(req.Version, pf.Frontmatter.Version) < 0 {
-			return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf("cannot lower version from %s to %s", pf.Frontmatter.Version, req.Version))
+		// The floor is the last APPROVED milestone, not the document's current
+		// draft value: document_versions rows are written only on approve/
+		// merge/confirm, so while a document sits in draft its frontmatter
+		// version is working state, not a milestone — a user must be free to
+		// correct it either way (e.g. undo a typo'd bump) as long as it never
+		// regresses past what was actually accepted. Comparing against the
+		// current draft value instead locks in whatever was last saved,
+		// mistake or not, and blocks exactly that correction.
+		if latest, err := s.db.LatestVersion(c.Request().Context(), orgID, docID); err == nil {
+			if compareVersions(req.Version, latest.Version) <= 0 {
+				return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf(
+					"cannot set version to %s: must be greater than %s, the last approved version (draft version numbers may otherwise move freely)",
+					req.Version, latest.Version))
+			}
 		}
 		pf.Frontmatter.Version = req.Version
 	}
--- a/internal/isms/api/server.go
+++ b/internal/isms/api/server.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	gowebauthn "github.com/go-webauthn/webauthn/webauthn"
+	"github.com/jackc/pgx/v5"
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
 	"gopkg.in/yaml.v3"
@@ -1365,6 +1366,17 @@ func (s *Server) handleUpdateDocumentMetadata(c echo.Context) error {
 		}
 	}
 
+	// This is the second reachable writer of Frontmatter.Version (the other is
+	// PUT .../content) — the floor must be enforced here too, or it is
+	// advisory rather than real: whichever version-edit affordance a user
+	// happens to touch (the editor's save payload vs. the document header's
+	// version field) would otherwise get a different, inconsistent answer.
+	if v, ok := req.Fields["version"]; ok && v != "" {
+		if err := s.versionFloorError(c.Request().Context(), orgID, docID, v); err != nil {
+			return err
+		}
+	}
+
 	st, err := s.storeForOrg(c.Request().Context(), orgID)
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, err.Error())
@@ -1635,20 +1647,8 @@ func (s *Server) handleUpdateDocumentContent(c echo.Context) error {
 
 	// Update metadata fields if provided
 	if req.Version != "" {
-		// The floor is the last APPROVED milestone, not the document's current
-		// draft value: document_versions rows are written only on approve/
-		// merge/confirm, so while a document sits in draft its frontmatter
-		// version is working state, not a milestone — a user must be free to
-		// correct it either way (e.g. undo a typo'd bump) as long as it never
-		// regresses past what was actually accepted. Comparing against the
-		// current draft value instead locks in whatever was last saved,
-		// mistake or not, and blocks exactly that correction.
-		if latest, err := s.db.LatestVersion(c.Request().Context(), orgID, docID); err == nil {
-			if compareVersions(req.Version, latest.Version) <= 0 {
-				return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf(
-					"cannot set version to %s: must be greater than %s, the last approved version (draft version numbers may otherwise move freely)",
-					req.Version, latest.Version))
-			}
+		if err := s.versionFloorError(c.Request().Context(), orgID, docID, req.Version); err != nil {
+			return err
 		}
 		pf.Frontmatter.Version = req.Version
 	}
@@ -3165,6 +3165,44 @@ func (s *Server) handleCreateOverdueTasks(c echo.Context) error {
 		Detail: fmt.Sprintf("Created %d review tasks (%d skipped, already existed)", len(result.Created), result.Skipped),
 	})
 	return c.JSON(http.StatusOK, result)
+}
+
+// versionFloorError returns a 400 if v is not strictly above the last
+// approved milestone for docID (a document_versions row, written only on
+// merge/confirm — see CLAUDE.md). A document with no milestone yet (never
+// approved) is unconstrained: nothing has been accepted for its version to
+// regress past.
+//
+// Called from every writer of Frontmatter.Version reachable over the API —
+// PUT .../content and PUT .../metadata both go through this, so the floor
+// cannot be bypassed by picking the other endpoint.
+//
+// A database error other than "no milestone" (pgx.ErrNoRows) is a 500, not a
+// silent skip: this guard exists to enforce a governance invariant, and
+// "the check didn't run because the database hiccuped" must not read as
+// "the check passed".
+//
+// compareVersions reads major.minor only, so a value exactly equal to the
+// milestone (rejected by this <= 0 comparison, not just < 0) also catches a
+// same-major.minor patch component it cannot see — "1.0.1" reads as equal to
+// "1.0.0". No caller in this repo produces a 3-component version today
+// (incrementVersion only ever emits major.minor), so this is unreachable in
+// practice, but it is reachable the moment an operator types one by hand,
+// since the field is a free-form frontmatter string.
+func (s *Server) versionFloorError(ctx context.Context, orgID int, docID, v string) error {
+	latest, err := s.db.LatestVersion(ctx, orgID, docID)
+	if err != nil {
+		if errors.Is(err, pgx.ErrNoRows) {
+			return nil
+		}
+		return echo.NewHTTPError(http.StatusInternalServerError, "checking version floor: "+err.Error())
+	}
+	if compareVersions(v, latest.Version) <= 0 {
+		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf(
+			"cannot set version to %s: must be greater than %s, the last approved version (draft version numbers may otherwise move freely)",
+			v, latest.Version))
+	}
+	return nil
 }
 
 // stripFrontmatter removes YAML frontmatter from markdown content.
--- a/tests/test_review_branch.py
+++ b/tests/test_review_branch.py
@@ -259,23 +259,56 @@ class TestStaleMergeConflict:
 
 
 class TestDocumentContentSave:
-    """Test that document edit saves content+metadata in one commit."""
+    """Test that document edit saves content+metadata in one commit.
+
+    iso27001-4-1 is a shared scaffold document touched by many test files, so
+    its version only ever climbs across a persistent stack's repeated runs.
+    These tests only care that a version+author save works at all, not which
+    exact number — so they read the document's current version and bump the
+    MAJOR component, which is always strictly above both it and any approved
+    milestone (an approval can never exceed the draft version it was approved
+    from). A hardcoded literal ("1.0", "1.1") would eventually fall behind
+    that climb and start colliding with the approved-version floor.
+    """
+
+    def _next_major(self, api_url, admin_headers):
+        # The floor this PUT must clear is the last APPROVED milestone
+        # (document_versions), which on a persistent stack can outrun the
+        # document's current draft frontmatter — some other test elsewhere in
+        # the suite may have re-scaffolded or otherwise reset iso27001-4-1's
+        # git content to a lower version after an earlier run already
+        # recorded a higher milestone for it. Reading only the draft value
+        # was not enough; take whichever of the two is higher.
+        majors = [0]
+        r = requests.get(f"{api_url}/documents/iso27001-4-1/body", headers=admin_headers)
+        assert r.status_code == 200, f"reading current version: {r.text}"
+        draft = r.json().get("version") or ""
+        if draft:
+            majors.append(int(draft.split(".")[0]))
+        r = requests.get(f"{api_url}/documents/iso27001-4-1/versions", headers=admin_headers)
+        assert r.status_code == 200, f"reading version history: {r.text}"
+        history = r.json().get("data") or []
+        if history:
+            majors.append(int(history[0]["version"].split(".")[0]))
+        return f"{max(majors) + 1}.0"
 
     def test_save_content_with_version_and_author(self, api_url, admin_headers):
         """PUT /documents/:id/content with version+author should work."""
+        v = self._next_major(api_url, admin_headers)
         r = requests.put(f"{api_url}/documents/iso27001-4-1/content",
                          headers=admin_headers,
                          json={"content": "# Test content\n\nUpdated.",
-                               "version": "1.0",
+                               "version": v,
                                "author": "testadmin@isms-test.local"})
         assert r.status_code == 200, f"Save failed: {r.text}"
         assert "commit" in r.json()
 
     def test_metadata_multi_field_update(self, api_url, admin_headers):
         """PUT /documents/:id/metadata with multiple fields in one commit."""
+        v = self._next_major(api_url, admin_headers)
         r = requests.put(f"{api_url}/documents/iso27001-4-1/metadata",
                          headers=admin_headers,
-                         json={"fields": {"version": "1.1", "status": "draft"}})
+                         json={"fields": {"version": v, "status": "draft"}})
         assert r.status_code == 200, f"Metadata update failed: {r.text}"
         assert "commit" in r.json()
 
--- a/tests/test_version_downgrade_guard.py
+++ b/tests/test_version_downgrade_guard.py
@@ -1,0 +1,98 @@
+"""Regression: the version-downgrade guard on PUT /documents/:docId/content must
+compare a requested version against the last APPROVED milestone
+(document_versions, written only on merge/confirm), not against the document's
+current draft frontmatter value.
+
+A draft's version field is working state until it is re-approved — nothing
+depends on it staying monotonic against its own prior unapproved saves. Before
+this fix, comparing against the current draft value locked in whatever was
+last saved (including a typo) as a new floor, so a user correcting an
+over-bumped draft version (e.g. 0.3 -> 0.2, still well above the 0.1 that was
+actually approved) was permanently blocked.
+"""
+import uuid
+import requests
+from conftest import READER_EMAIL
+
+BODY = "# Heading\n\nBody content for the version-downgrade-guard regression test.\n"
+
+
+def _approve_and_merge(api_url, admin_headers, reader_headers, doc_id):
+    r = requests.post(f"{api_url}/documents/{doc_id}/reviews", headers=admin_headers,
+                      json={"reviewers": [READER_EMAIL], "message": "review"})
+    assert r.status_code in (200, 201), f"send for review: {r.text}"
+    rid = r.json()["review_id"]
+    r = requests.post(f"{api_url}/reviews/{rid}/approve", headers=reader_headers,
+                      json={"decision": "approved", "comment": "LGTM"})
+    assert r.status_code == 200, f"approve: {r.text}"
+    r = requests.post(f"{api_url}/reviews/{rid}/merge", headers=admin_headers, json={})
+    assert r.status_code == 200, f"merge: {r.text}"
+
+
+def _current_version(api_url, admin_headers, doc_id):
+    # PUT .../content returns only {"commit", "status"} — the applied version
+    # has to be read back separately.
+    r = requests.get(f"{api_url}/documents/{doc_id}/body", headers=admin_headers)
+    assert r.status_code == 200, f"reading back version: {r.text}"
+    return r.json()["version"]
+
+
+class TestVersionDowngradeGuardComparesAgainstApprovedFloor:
+    def test_draft_version_may_correct_downward_above_the_approved_floor(self, api_url, admin_headers, reader_headers):
+        # Unique per run — the assertions below pin absolute version numbers
+        # ("0.2", "0.3"), which only hold for a document that has never been
+        # approved before this test creates it. A fixed id would keep
+        # climbing on a persistent test stack across repeated runs.
+        DOC = f"version-guard-test-{uuid.uuid4().hex[:8]}"
+
+        # 1. Create a document, pin its starting version, approve + merge.
+        #    No milestone exists yet at this point, so the floor check is
+        #    inert here regardless of the fix.
+        r = requests.post(f"{api_url}/documents", headers=admin_headers, json={
+            "folder": "iso27001", "filename": DOC + ".md",
+            "document_id": DOC, "title": "Version Guard Test", "content": BODY,
+        })
+        assert r.status_code in (200, 201), r.text
+        r = requests.put(f"{api_url}/documents/{DOC}/content", headers=admin_headers,
+                         json={"content": BODY, "version": "0.1"})
+        assert r.status_code == 200, f"pin initial version: {r.text}"
+
+        _approve_and_merge(api_url, admin_headers, reader_headers, DOC)
+        # The approved milestone is now 0.1 — the floor every later check in
+        # this test is measured against.
+
+        # 2. Edit the approved doc with no explicit version: auto-bumps to
+        #    0.2 and drops back to draft, exactly like a normal edit in the UI.
+        r = requests.put(f"{api_url}/documents/{DOC}/content", headers=admin_headers,
+                         json={"content": BODY + "\nAn edit.\n"})
+        assert r.status_code == 200, f"auto-bump edit: {r.text}"
+        assert _current_version(api_url, admin_headers, DOC) == "0.2", \
+            f"expected auto-bump to 0.2, got {_current_version(api_url, admin_headers, DOC)}"
+
+        # 3. User (mis)types 0.3 and saves — still draft, still above the 0.1
+        #    floor, so this must succeed both before and after the fix.
+        r = requests.put(f"{api_url}/documents/{DOC}/content", headers=admin_headers,
+                         json={"content": BODY + "\nAn edit.\n", "version": "0.3"})
+        assert r.status_code == 200, f"manual bump to 0.3: {r.text}"
+
+        # 4. THE FIX: correcting back to 0.2 must now succeed — 0.2 is still
+        #    above the 0.1 that was actually approved, even though it is
+        #    below the document's current (mistaken) draft value of 0.3.
+        r = requests.put(f"{api_url}/documents/{DOC}/content", headers=admin_headers,
+                         json={"content": BODY + "\nAn edit.\n", "version": "0.2"})
+        assert r.status_code == 200, \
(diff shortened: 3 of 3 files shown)


unidoc-alip (changes requested):
> **Verdict: REQUEST_CHANGES** — 1 must-fix, 2 should-fix, 1 nit

This PR changes the version-downgrade guard on `PUT /documents/:docId/content` so that a
requested `version` is compared against the last **approved** milestone (`db.LatestVersion`,
i.e. a `document_versions` row, written only on merge/confirm) rather than against the
document's current draft frontmatter value. The reasoning is sound and matches the
documented versioning model in `CLAUDE.md`: `document_versions` holds milestones, draft
frontmatter is working state, so a draft's own prior save should not become a floor. The
new pytest regression test walks the real sequence from the issue against a live server,
and CI (`go-unit`, `integration`) is green — the new test does run in CI, since the
`integration` job invokes `pytest tests/` wholesale.

What the review found: the invariant the PR sets out to establish ("a version may never
regress past what was actually accepted") is not actually enforced, because a sibling
endpoint writes the same frontmatter field with no check at all — and the web UI's own
"set version" control uses that endpoint, not the one this PR hardens (F1). The new
guard also fails open on any database error (F2), and both the PR body and issue #226
describe the never-approved case as "same as before" when it is in fact a real widening
(F3).

On the severity of F1: the unguarded sibling endpoint is a **pre-existing** gap, proved
present on `master` as well as on this branch, so it is not a regression introduced here.
It is filed as must-fix all the same, because it is the finding that decides whether this
PR achieves what it says it achieves. The change defines a governance floor and states
that a version "never regresses past what was actually accepted"; with the metadata
endpoint left as it is, that sentence is not true of the merged system, and the one route
a user is most likely to take — the version box in the document header — is the route with
no check on it. Closing it is a dozen lines and it reuses the very comparison this diff
already writes.

F2 *is* new in this diff (the old comparand could not fail, being already in memory) but
is only observable while the database is failing, so it stays should-fix. F3 is a
documentation correction, not a code defect.

Findings are keyed `F1`, `F2`, … rather than `#1`, `#2`, because GitHub would autolink
`#N` to an unrelated issue or PR.

---

### F1 [must-fix] `internal/isms/api/server.go` line 1645 — the new floor is bypassable through `PUT /documents/:docId/metadata`, which is the endpoint the UI's own version field uses

`handleUpdateDocumentMetadata` (`internal/isms/api/server.go:1340`) whitelists `version`
among its updatable frontmatter fields and applies it via
`store.UpdateDocumentMetadataMulti` (`internal/isms/store/git.go:1750`), which does a
plain textual frontmatter rewrite with **no version comparison of any kind** — neither
the old draft-value check nor the new milestone check. So after this PR there are two
writers of `Frontmatter.Version` reachable over the API, one guarded and one not:

- `PUT /documents/:docId/content` — guarded (this diff)
- `PUT /documents/:docId/metadata` with `{"fields": {"version": "0.0"}}` — unguarded

This is not a regression from this diff; the gap pre-dates it. It is raised as must-fix
because this PR is precisely the change that defines the floor, and the unguarded path is
not an obscure one. `web/src/views/Documents.vue:1804` is the document header's version
control:

```js
async function saveVersion() {
  const v = editVersion.value.trim()
  if (!v || !activeId.value || v === activeDoc.value?.version) return
  try {
    await api.updateDocumentMetadata(activeId.value, { version: v })   // <- no floor check server-side
```

The practical consequences:

- The floor is advisory. A manager/admin can set an approved document's version to `0.0`
  (or anything at all — the field is a free string) through the UI's own version box,
  which is a stronger version of the "bypass by editing frontmatter via git" workaround
  the issue describes, except it needs no git access.
- Enforcement depends on which UI affordance the user happens to touch. Editing the
  version inside the editor's save payload (`useDocumentEditor.js:124-126`) goes through
  the guarded endpoint and can be rejected; typing it in the header field
  (`Documents.vue:1804`) always succeeds. Same user intent, two outcomes, one error
  message that only appears sometimes.
- The same whitelist also accepts `status`, so the same endpoint can push a document to
  `approved` without a review — worth a separate look, out of scope for this PR.

Confirmed by source read plus a live PoC run against three servers (base, head, head +
the suggested fix — see the PoC section below). After approving and merging a document at
`0.1`, `PUT .../metadata` with `version: "0.0"` returns `200` on **both** base and head
and the frontmatter reads back as `0.0`, i.e. below the approved milestone, while the same
value through `PUT .../content` returns `400` on head. One extra detail the run surfaced:
on base, once the metadata write has walked the frontmatter down to `0.0`, the *content*
endpoint then accepts `0.0` too (it compares against the frontmatter, which is now `0.0`),
so the bypass used to be self-reinforcing. On head it stays rejected — this diff is a
real improvement on that path, which is exactly why the remaining hole is worth closing.

Suggested fix — lift the floor check into a helper and call it from both handlers:

```go
// versionFloorError returns a 400 if v is not strictly above the last approved
// milestone for docID. A document with no milestone yet is unconstrained.
func (s *Server) versionFloorError(ctx context.Context, orgID int, docID, v string) error {
	latest, err := s.db.LatestVersion(ctx, orgID, docID)
	if errors.Is(err, pgx.ErrNoRows) {
		return nil // never approved — no floor
	}
	if err != nil {
		return echo.NewHTTPError(http.StatusInternalServerError, "checking version floor: "+err.Error())
	}
	if compareVersions(v, latest.Version) <= 0 {
		return echo.NewHTTPError(http.StatusBadRequest, fmt.Sprintf(
			"cannot set version to %s: must be greater than %s, the last approved version (draft version numbers may otherwise move freely)",
			v, latest.Version))
	}
	return nil
}
```

then in `handleUpdateDocumentMetadata`, after the whitelist loop:

```go
if v, ok := req.Fields["version"]; ok && v != "" {
	if err := s.versionFloorError(c.Request().Context(), orgID, docID, v); err != nil {
		return err
	}
}
```

If enforcing it on the metadata endpoint is deliberately out of scope, that is a
defensible call — but then the PR should say so, because as merged the stated invariant
does not hold.

---

### F2 [should-fix] `internal/isms/api/server.go` line 1645 — the guard fails open on any database error, not just "no milestone"

```go
if latest, err := s.db.LatestVersion(c.Request().Context(), orgID, docID); err == nil {
	if compareVersions(req.Version, latest.Version) <= 0 {
```

`LatestVersion` (`internal/isms/db/versions.go:82`) returns a single wrapped error for
every failure mode:

```go
if err != nil {
	return nil, fmt.Errorf("no versions found for %s: %w", documentID, err)
}
```

so "this document has never been approved" (`pgx.ErrNoRows`) is indistinguishable at the
call site from a connection reset, a cancelled request context, a statement timeout, or
a schema problem. `err == nil` therefore means "check the floor only if the database
happened to answer" — under database trouble the endpoint silently accepts any version,
including one below an approved milestone, and reports `200`. The old code could not fail
this way: its comparand came from the already-loaded document.

For a guard whose whole purpose is a governance invariant, silence is the wrong default.
The fix is to separate the two cases (`errors.Is(err, pgx.ErrNoRows)` works because
`LatestVersion` wraps with `%w`) — see the `versionFloorError` helper in F1, which does
exactly this: no-rows skips the guard, any other error is a 500.

Source-read only. A PoC would require injecting a database fault mid-request; the code
path is unambiguous enough from reading it.

---

### F3 [should-fix] PR body and issue #226 — "no milestone means unconstrained, same as before" is not accurate; the never-approved case is genuinely widened

Both the PR description ("A document with no approval history yet (no milestone) is
unconstrained, same as before") and issue #226 ("if the document has never been approved
(no milestone exists), the check is inert, matching prior behavior") state that nothing
changes for a document with no approval history. That is the opposite of what happens:

- Before: the comparand was `pf.Frontmatter.Version`, which exists on every document
  regardless of approval history. A never-approved document sitting at draft `0.5` could
  not be set to `0.1` — the guard applied.
- After: `LatestVersion` errors, the whole check is skipped, and a never-approved
  document's version can be set to anything, in either direction.

So the change is not "same floor, better source" — it removes the guard entirely for
every document that has never been through approve+merge, which for a freshly scaffolded
org is all of them. That may well be the intended reading of "draft version is working
state" (it is consistent with the rest of the rationale), and I am not asking for the
old behavior back. But the claim as written would lead a reviewer or a future reader to
believe a protection is still in place where it is not, so it is worth correcting in the
PR body and in the issue.

PoC-confirmed, base vs head. Creating a document, pinning it to `0.5` without ever
approving it, then requesting `0.1`:

- base (`master` c6a45cd): `400 cannot lower version from 0.5 to 0.1`
- head (`cedff33`): `200`, and the version reads back as `0.1`

So the guard did apply before and does not now.

---

### F4 [nit] `internal/isms/api/server.go` line 1646 — tightening `< 0` to `<= 0` also rejects patch-level bumps, because `compareVersions` only reads major.minor

`compareVersions` (`internal/isms/api/server.go:3189`) splits on the *first* dot only and
`Sscanf`s the remainder as the minor, so a third component is silently absorbed:
`compareVersions("1.0.1", "1.0.0") == 0`. With the comparison loosened from `< 0` to
`<= 0`, that equality is now a rejection: a document approved at `1.0.0` cannot be moved
to `1.0.1`, `1.0.2`, or any other patch of the same major.minor, and the error message
would read "must be greater than 1.0.0" about a value that plainly is.

Low reachability, which is why this is a nit and not a should-fix: I found no 3-component
version anywhere in the repo, `incrementVersion` only ever produces `major.minor`, and
the templates and tests all use two components. It becomes reachable the moment an
operator types one, since `version` is a free-form frontmatter string. If patch versions
are considered out of scope, a one-line comment on `compareVersions` saying it compares
major.minor only would save the next reader the trip.

The related behavior change — a version *exactly equal* to the approved milestone is now
rejected where it used to be a no-op — looks intended (step 5 of the new test asserts it
explicitly), and the primary UI path omits an unchanged version from the payload
(`useDocumentEditor.js:124`), so I am not flagging it as a finding. Noting it only
because it is a semantic change the description does not mention.

---

## Proof of concept

Vehicle: the repo's own pytest integration suite against live servers (the layer the guard
lives at; the `integration` CI job runs the same suite). Three variants were run
side by side, each with its own throwaway Postgres database, its own data dir and its own
port, per the project's standard local test env — stale listeners checked for first, and
each server's log checked for a bind error rather than trusting `/healthz`:

- **base** — `master` at c6a45cd, the PR's actual parent, on `:9092`
- **head** — the PR at cedff33, on `:9091`
- **fixed** — head plus the `versionFloorError` helper from F1 wired into both handlers
  (including the `errors.Is(err, pgx.ErrNoRows)` split from F2), on `:9093`

Two PoC files, reproduced in full in the appendix at the end of this review so they can be
dropped straight into `tests/` and run. They are named `poc_*` rather than `test_*` so
pytest's default collection will not pick them up in CI.

`tests/poc_227_version_floor_bypass.py` — F1. Mirrors the PR's own test up to the merged
`0.1` milestone, then sends the identical requested version (`0.0`, below the milestone) to
both endpoints and asserts on one joined outcome string:

```
base:  milestone=0.1 | content_0.0=400 | metadata_0.0=200 | readback=0.0 | content_0.0_again=200
head:  milestone=0.1 | content_0.0=400 | metadata_0.0=200 | readback=0.0 | content_0.0_again=400
fixed: milestone=0.1 | content_0.0=400 | metadata_0.0=400 | readback=0.1 | content_0.0_again=400
```

Read across: the metadata bypass is present on base and head alike, so F1 is a
pre-existing gap rather than a regression from this diff; head hardens the content endpoint
so the bypass is no longer self-reinforcing; and the suggested fix closes it outright,
leaving the frontmatter at `0.1` where base and head both leave it at `0.0`.

The rejection message was captured live and matches the description's claim verbatim:
`cannot set version to 0.0: must be greater than 0.1, the last approved version (draft
version numbers may otherwise move freely)`.

`tests/poc_227_never_approved_floor.py` — F3. Never-approved document pinned to `0.5`, then
asked to move to `0.1`: base `400 cannot lower version from 0.5 to 0.1`, head `200`.

To run: save both appendix files into `tests/`, start a server the way
`.github/workflows/tests.yml`'s `integration` job does, then

```
ISMS_TEST_URL=http://localhost:9091 pytest tests/poc_227_version_floor_bypass.py -s
ISMS_TEST_URL=http://localhost:9091 pytest tests/poc_227_never_approved_floor.py -s
# for the base / fixed variants, pass the expected string:
#   POC227_WANT="milestone=0.1 | content_0.0=400 | metadata_0.0=200 | readback=0.0 | content_0.0_again=200"
```

Regression check on the suggested fix, not just on the bug: the full non-e2e suite was run
against the **fixed** server — `689 passed, 1 skipped`, i.e. the same result the PR claims
for head, including the PR's own new `test_version_downgrade_guard.py`. (Two
`test_routing.py` failures appeared on the first attempt and were a harness artifact, not
the fix: the extra worktree had a zero-byte placeholder `index.html`, so the apex `<meta>`
injection had no `</head>` to inject into. Both pass once the real `index.html` is in place.)

Caveats on what the PoC does not model: it runs as an admin in a single org, so it says
nothing about RBAC interaction; it does not drive the browser, so the claim that
`Documents.vue:1804` is a user-reachable route into the unguarded endpoint rests on the
source read rather than on clicking the control; and F2 (fail-open on database error) is not
covered at all, since it needs a fault injection the suite has no vehicle for — the `fixed`
variant demonstrates that the `pgx.ErrNoRows` split compiles and preserves both intended
behaviors, but not that it catches a real outage.

Also noted while setting this up, not a finding: the PR is based on c6a45cd while
`origin/master` has since moved to 58c13ee (#224, #218). Neither of those commits touches
`handleUpdateDocumentContent`, `compareVersions`, `LatestVersion` or the version control in
`Documents.vue`, so there is no interaction to worry about — a rebase is housekeeping only.

---

- **"compared a requested version against the document's current draft frontmatter value
  instead of the last approved milestone"** — **confirmed.** The removed line compared
  against `pf.Frontmatter.Version`; `document_versions` rows are written only from the
  two `RecordVersionTx` call sites in `api_collab.go` (merge and confirm), matching the
  model documented in `CLAUDE.md`.
- **"Now compares against `db.LatestVersion()`"** — **confirmed**, with the caveat in F2
  that the comparison is skipped on any error rather than only on no-rows, and the caveat
  in F1 that the sibling metadata endpoint still performs no comparison at all.
- **"A document with no approval history yet (no milestone) is unconstrained, same as
  before"** — **not confirmed**, and disproved live. See F3: on base the same sequence
  (never-approved document pinned to `0.5`, then asked for `0.1`) returns
  `400 cannot lower version from 0.5 to 0.1`; on head it returns `200`. The first clause
  is right; "same as before" is not.
- **"The rejection message now names the actual floor"** — **confirmed.** The message
  includes the milestone version and explains that draft numbers may otherwise move
  freely. Verified live in the PoC.
- **"New regression test walks the full sequence against a live server: create → pin 0.1
  → approve → merge → auto-bump edit → manual over-bump → correct downward → attempt
  at-or-below-floor"** — **confirmed.** `tests/test_version_downgrade_guard.py` does
  exactly that, and the randomized `document_id` plus the comment explaining why it is
  randomized correctly handles the repeated-run case on a persistent stack.
- **"`go build`/`vet`/`test` clean; full non-e2e pytest suite and the full
  `test_e2e_browser.py` suite green"** — **partially confirmed.** Not re-run in full
  here, but CI's `go-unit` (build, gofmt, `go test ./internal/...`) and `integration`
  (`pytest tests/`, which includes both the new test and `test_e2e_browser.py`) are both
  green on `cedff33`, and the non-e2e suite reproduced locally at `689 passed, 1 skipped`
  — the exact count claimed. That covers the substance of it. Worth adding that, unlike
  the CI claim on some earlier PRs here, the jobs named really do exist and really do run
  this test: `.github/workflows/tests.yml` invokes `pytest tests/` with no selector.
- **"Surfaced live on a production self-hosted deployment — a user had to bypass the API
  by editing frontmatter directly via git"** — **not verifiable** from the repo, and
  taken at face value. Worth noting alongside F1 that the API bypass the user needed did
  in fact exist: `PUT /documents/:docId/metadata` would have accepted the corrected
  version without any git access.

## Not findings, checked and cleared

- `LatestVersion` orders by `created_at DESC`, and `RecordVersion`'s upsert does not
  touch `created_at` on conflict, so "latest" is the most recently *inserted* milestone
  rather than the highest version. I could not construct a reachable sequence where those
  differ, since a milestone is only ever written at the version the document held when it
  was merged or confirmed. Noting it as read-and-dismissed rather than as a finding.
- The auto-bump branch above the guard (`status == "approved" && req.Version == ""`)
  bumps the frontmatter without consulting the floor, but it only ever increments, and
  its input is the frontmatter value that a guarded write already kept above the
  milestone — no way to land at or below the floor through it.
- `PUT /reviews/:id/content` (`api_collab.go:1579`) takes raw content and so looked like a
  possible third unguarded writer of `version`, but it explicitly re-reads the existing
  frontmatter block from the review branch and prepends it to the submitted body
  (`api_collab.go:1633-1658`), replacing only the body. It cannot change the version.
- The new test hardcodes the `iso27001` folder, which `CLAUDE.md` warns against for core
  code. Consistent with every other test in `tests/`, so not worth raising.

---

## Appendix: the PoC files

Save these into `tests/` to reproduce the runs above. Both use the existing `conftest.py`
fixtures (`api_url`, `admin_headers`, `reader_headers`), so they need nothing beyond a
running server.

### `tests/poc_227_version_floor_bypass.py` (F1)

```python
"""PoC for PR #227 review, finding F1: the version floor introduced on
PUT /documents/:docId/content is not enforced on PUT /documents/:docId/metadata,
which is the endpoint the web UI's own version field uses
(web/src/views/Documents.vue saveVersion -> api.updateDocumentMetadata).

Sequence mirrors tests/test_version_downgrade_guard.py up to the merged 0.1
milestone, then sends the SAME requested version (0.0, below the milestone) to
both endpoints and compares.
"""
import os
import uuid
import requests
from conftest import READER_EMAIL

BODY = "# Heading\n\nBody for the version-floor bypass PoC.\n"


def _approve_and_merge(api_url, admin_headers, reader_headers, doc_id):
    r = requests.post(f"{api_url}/documents/{doc_id}/reviews", headers=admin_headers,
                      json={"reviewers": [READER_EMAIL], "message": "review"})
    assert r.status_code in (200, 201), f"send for review: {r.text}"
    rid = r.json()["review_id"]
    r = requests.post(f"{api_url}/reviews/{rid}/approve", headers=reader_headers,
                      json={"decision": "approved", "comment": "LGTM"})
    assert r.status_code == 200, f"approve: {r.text}"
    r = requests.post(f"{api_url}/reviews/{rid}/merge", headers=admin_headers, json={})
    assert r.status_code == 200, f"merge: {r.text}"


def _version(api_url, admin_headers, doc_id):
    r = requests.get(f"{api_url}/documents/{doc_id}/body", headers=admin_headers)
    return r.json()["version"]


def test_metadata_endpoint_bypasses_the_approved_version_floor(api_url, admin_headers, reader_headers):
    DOC = f"poc227-{uuid.uuid4().hex[:8]}"
    outcomes = []

    r = requests.post(f"{api_url}/documents", headers=admin_headers, json={
        "folder": "iso27001", "filename": DOC + ".md",
        "document_id": DOC, "title": "PoC 227", "content": BODY,
    })
    assert r.status_code in (200, 201), r.text
    r = requests.put(f"{api_url}/documents/{DOC}/content", headers=admin_headers,
                     json={"content": BODY, "version": "0.1"})
    assert r.status_code == 200, f"pin 0.1: {r.text}"
    _approve_and_merge(api_url, admin_headers, reader_headers, DOC)
    milestone = _version(api_url, admin_headers, DOC)
    outcomes.append(f"milestone={milestone}")

    # A: guarded endpoint, version below the milestone -> expect 400
    r = requests.put(f"{api_url}/documents/{DOC}/content", headers=admin_headers,
                     json={"content": BODY, "version": "0.0"})
    outcomes.append(f"content_0.0={r.status_code}")
    content_msg = r.json().get("message", "") if r.status_code >= 400 else ""

    # B: unguarded sibling endpoint, SAME value -> observed 200
    r = requests.put(f"{api_url}/documents/{DOC}/metadata", headers=admin_headers,
                     json={"fields": {"version": "0.0"}})
    outcomes.append(f"metadata_0.0={r.status_code}")
    outcomes.append(f"readback={_version(api_url, admin_headers, DOC)}")

    # C: the DB milestone is untouched — content endpoint still rejects
    r = requests.put(f"{api_url}/documents/{DOC}/content", headers=admin_headers,
                     json={"content": BODY, "version": "0.0"})
    outcomes.append(f"content_0.0_again={r.status_code}")

    joined = " | ".join(outcomes)
    # Expected outcome per variant (POC227_WANT overrides):
    #   base  (master c6a45cd): ... metadata_0.0=200 | readback=0.0 | content_0.0_again=200
    #   head  (PR cedff33):     ... metadata_0.0=200 | readback=0.0 | content_0.0_again=400
    #   fixed (head + F1 fix):  ... metadata_0.0=400 | readback=0.1 | content_0.0_again=400
    want = os.environ.get(
        "POC227_WANT",
        "milestone=0.1 | content_0.0=400 | metadata_0.0=200 | readback=0.0 | content_0.0_again=400")
    print("\nPoC F1 outcome: " + joined)
    print("content endpoint message: " + content_msg)
    assert joined == want, f"\ngot:  {joined}\nwant: {want}"
    assert "approved" in content_msg.lower(), content_msg
```

### `tests/poc_227_never_approved_floor.py` (F3)

```python
"""PoC for PR #227 review, finding F3: for a document that has NEVER been
approved (no document_versions milestone), the old guard still constrained the
version against the draft frontmatter value; the new one skips entirely.

Run against a base(master) server and a head(PR) server: base should reject the
downward move, head should accept it. The PR body claims this case is
"unconstrained, same as before".
"""
import uuid
import requests

BODY = "# Heading\n\nBody for the never-approved floor PoC.\n"


def test_never_approved_draft_version_may_move_down(api_url, admin_headers):
    DOC = f"poc227b-{uuid.uuid4().hex[:8]}"
    r = requests.post(f"{api_url}/documents", headers=admin_headers, json={
        "folder": "iso27001", "filename": DOC + ".md",
        "document_id": DOC, "title": "PoC 227b", "content": BODY,
    })
    assert r.status_code in (200, 201), r.text

    # Never approved: no milestone will ever exist for this document.
    r = requests.put(f"{api_url}/documents/{DOC}/content", headers=admin_headers,
                     json={"content": BODY, "version": "0.5"})
    assert r.status_code == 200, f"pin 0.5: {r.text}"

    r = requests.put(f"{api_url}/documents/{DOC}/content", headers=admin_headers,
                     json={"content": BODY, "version": "0.1"})
    print(f"\nPoC F3 outcome: never_approved_0.5_to_0.1={r.status_code}"
          f" msg={r.json().get('message','') if r.status_code>=400 else '-'}")
```

unidoc-ahall:
Addressed all four findings, pushed as one commit.

**F1 (must-fix)** — you're right this was the finding that decided whether the PR means what it says. Lifted the check into `versionFloorError()` and call it from `handleUpdateDocumentMetadata` too, right after the field whitelist. Added a test that mutation-checks itself: reverted just the metadata-endpoint call locally and confirmed exactly `test_metadata_endpoint_enforces_the_same_floor_as_content` fails, nothing else — same discipline your PoC used.

**F2 (should-fix)** — folded into the same helper: `errors.Is(err, pgx.ErrNoRows)` skips the check (no milestone), anything else is a 500. Source-level fix only, as you noted a real fault-injection test isn't feasible with what this suite has.

**F3 (should-fix)** — corrected both the PR body and issue #226. You're right it's a genuine widening, not a documentation nit that changes nothing — added a test (`test_never_approved_document_has_no_floor_on_either_writer`) that pins it as the intended contract on both writers now, rather than leaving it as an unstated side effect.

**F4 (nit)** — added the caveat to `versionFloorError`'s doc comment. Left `compareVersions` itself alone since it already documents "major.minor" in its own comment; the risk is specifically in the `<= 0` comparison this helper adds, so that's where I put the note.

One more thing your review indirectly surfaced, found running the full suite twice against a persistent stack (not something your single-fresh-DB PoC run would hit): `test_review_branch.py::TestDocumentContentSave` hardcoded absolute version strings against the shared `iso27001-4-1` scaffold document. It only worked before because the old guard's floor was the mutable draft value, which those tests could always out-climb; the real milestone floor doesn't move that way. Fixed to derive a version from the higher of the document's current draft major and its latest milestone major.

Re-verified everything: `go build`/`vet`/`test` clean, full non-e2e suite 691 passed/1 skipped and `test_e2e_browser.py` 69 passed, both run three times against the same persistent database to confirm idempotency holds now.

unidoc-alip (approved):
> **Verdict: APPROVE** — all four findings verified fixed on 8e24ac0

Re-reviewed against 8e24ac0. Every finding is addressed, and I re-ran the evidence rather than reading the diff and taking the rest on trust.

**F1 (must-fix) — closed.** `versionFloorError` is called from both writers. The PoC from my earlier review, restored verbatim from that comment's appendix and run against the new head:

```
cedff33:  content_0.0=400 | metadata_0.0=200 | readback=0.0 | content_0.0_again=400
8e24ac0:  content_0.0=400 | metadata_0.0=400 | readback=0.1 | content_0.0_again=400
```

I also reproduced your mutation check independently instead of taking it on trust: removed only the `versionFloorError` call in `handleUpdateDocumentMetadata` from a copy of 8e24ac0 and ran the full suite — exactly `test_metadata_endpoint_enforces_the_same_floor_as_content` failed, nothing else. The claim holds precisely.

**F2 — closed, and confirmed live rather than only at source level.** The never-approved case returns `200`, not `500`, which proves `errors.Is(err, pgx.ErrNoRows)` actually penetrates `LatestVersion`'s `%w` wrapping. That was the one assumption in my suggested fix I had not verified, so this is now better established than when I proposed it.

**F3 — corrected** in both the PR body and issue #226, and pinning the widened contract with `test_never_approved_document_has_no_floor_on_either_writer` is more than I asked for. Making it an asserted contract rather than an unstated side effect is the right call.

**F4 — noted** on `versionFloorError`. Putting it there rather than on `compareVersions` is the better placement: the `<= 0` comparison is where the gap actually bites.

**On the `test_review_branch.py` fix you surfaced — that one is yours, and it is a real catch.** I reproduced it: the full suite twice against a single persistent database on cedff33 gives `689 passed` on run 1 and then `test_save_content_with_version_and_author` failing on run 2. On 8e24ac0 it survives repeated runs. My review ran against a fresh database every time, exactly as CI does, so neither my PoC nor CI could have surfaced it — your persistent-stack discipline found a latent break in the commit I had already reviewed.

Also verified: `gofmt` clean, `go build ./...`, `go vet ./internal/...`, `go test ./internal/...` all pass; CI green on the new commit.

Two notes on my local numbers so they are not over-read. `test_security.py::test_brute_force_protection` fails in my environment on every variant — cedff33, 8e24ac0 and the mutant alike — as an artifact of `ISMS_RATE_LIMIT=0` in the local recipe; it is unrelated to this PR and CI passes it, so my counts read `691 passed, 1 failed` where yours read `691 passed, 1 skipped`. The passing count matches exactly. Separately, one run in four collapsed and was not reproducible on the next, the brute-force test's lockout bleeding into the following run — not a code issue, but the suite is still not entirely clean back-to-back.

Nothing blocking. One passing observation, not worth another round: `_next_major()` reads `history[0]["version"]` on the assumption that newest-inserted is highest-versioned, and `int(draft.split(".")[0])` would raise on a non-numeric version string. Both are fine given what the suite actually produces.

Good to merge.

PR #231, opened 2026-08-27
fix(web): split list blocks keep their original numbering

### LINKED ISSUE #230: Read view: split ordered-list blocks all render as "1."
## Bug

The document read view (`Documents.vue`) splits every list item into its own "block" so each bullet gets an inline-comment button (`contentBlocks` computed, now `buildContentBlocks` in `web/src/utils/contentBlocks.js`). Each `<li>` ends up alone in a brand-new `<ol>` — and since that wrapper carried no `start` attribute, the browser has no way to know the item's real position and renders every single one as "1.".

Confirmed with the actual `marked` renderer against a real customer document: `marked` itself parses a 14-item list (including one with a blank line between items 12 and 13, which per CommonMark still keeps it as ONE list) into a single correct `<ol>` with 14 `<li>` in order — the bug is entirely in the block-splitting step afterward, not in markdown parsing.

This is why the edit view (tiptap, which renders the whole list as one WYSIWYG list) and the read view disagreed: only the read view does this per-item split.

## Fix

Each split `<li>` is now wrapped in `<ol start="N">`, where N is the item's actual position (the original list's own `start` attribute, if the source set one, plus the item's index within it) — so the browser renders each isolated block with its correct number. Unordered lists are untouched (no numbers to preserve).

Also extracted the block-splitting logic out of the `.vue` file into `web/src/utils/contentBlocks.js` so it has real, importable test coverage — the logic runs against `document.createElement`, so it could not previously be unit tested without transplanting it into a test file by hand.

Surfaced live on a production self-hosted deployment, reported against `documents/policy-acceptable-use.md`.

unidoc-ahall:
Fixes https://github.com/unidoc/isms/issues/230.

Documents.vue's read view splits every list item into its own commentable block. Each `<li>` ends up alone in a brand-new `<ol>` with no `start` attribute, so the browser renders every item as "1." regardless of its real position — confirmed against a real customer document (a 14-item list, every item showing "1." in the read view).

Traced it precisely rather than guessing: fed the exact markdown to `marked` directly (same instance `useRenderMd.js` uses) and it correctly parses a single `<ol>` with all 14 `<li>` in order — including across a blank line between two items, which CommonMark keeps as one "loose" list rather than splitting it. The bug is entirely in `contentBlocks`' per-item split that runs on the already-correct HTML, not in markdown parsing.

This also explains why edit and read view disagreed (the second part of the report): the editor (tiptap) renders the whole list as one WYSIWYG list and never splits it, so only the read view was affected.

**Fix:** each split `<li>` now carries `start="N"` on its wrapper `<ol>`, computed from the original list's own `start` (default 1, but respects a source list that legitimately starts elsewhere) plus the item's position within it.

**Also:** extracted the split logic out of the `.vue` file into `web/src/utils/contentBlocks.js` — it runs against `document.createElement`, so it had no real test coverage before (would have needed transplanting into a test file by hand, the way `router.test.js` documents having to do for router.js). New `web/test/contentBlocks.test.js` covers: numbering across a blank-line-separated loose list, a list with a non-1 `start`, unordered lists staying unaffected, and the other block types (tables, headings/paragraphs) unaffected by the change.

**Verified:** mutation-checked — reverted just the `start`-threading and confirmed exactly the 3 numbering-related tests fail, nothing else. `just build-web` clean, full `npm test` 83/83 (76 existing + 7 new).

### DIFF
--- a/web/src/utils/contentBlocks.js
+++ b/web/src/utils/contentBlocks.js
@@ -1,0 +1,82 @@
+// Splits rendered document HTML into paragraph-level "blocks" for the
+// document viewer: one block per heading/paragraph/pre/hr, one per list item,
+// one per table row, one per blockquote paragraph. Each block is what gets an
+// inline-comment "+" button and a stable #ph<hash> anchor.
+//
+// Extracted from Documents.vue's `contentBlocks` computed so this can be unit
+// tested directly — the logic runs against a DOM (document.createElement),
+// which node --test only has via jsdom in the test file, not via importing
+// the .vue component itself.
+import { parseMd } from '../composables/useRenderMd.js'
+
+export function buildContentBlocks(rawContent) {
+  if (!rawContent) return []
+  const html = parseMd(rawContent)
+  const div = document.createElement('div')
+  div.innerHTML = html
+  const blocks = []
+
+  function addBlock(html, tag, text) {
+    blocks.push({ index: blocks.length, html, tag, text })
+  }
+
+  for (const child of div.children) {
+    const tag = child.tagName.toLowerCase()
+
+    // Split lists — each bullet is commentable. Each <li> ends up alone in
+    // its own fresh <ol>, which resets an ordered list's visible number to 1
+    // for every item — so the wrapper carries a `start` tracking this item's
+    // real position in the original list, rather than a bare "<ol>" every
+    // time.
+    if ((tag === 'ul' || tag === 'ol') && child.children.length > 0) {
+      const start = tag === 'ol' ? (parseInt(child.getAttribute('start'), 10) || 1) : 1
+      let position = 0
+      for (const li of child.children) {
+        if (li.tagName.toLowerCase() === 'li') {
+          const openTag = tag === 'ol' ? `<ol start="${start + position}">` : '<ul>'
+          addBlock(openTag + li.outerHTML + '</' + tag + '>', 'li', li.textContent || '')
+          position++
+        }
+      }
+    }
+    // Tables — convert to grid-based rows so each gets a "+" button
+    else if (tag === 'table') {
+      const thead = child.querySelector('thead')
+      const tbody = child.querySelector('tbody')
+      const ths = thead ? Array.from(thead.querySelectorAll('th')) : []
+      const rows = tbody ? Array.from(tbody.querySelectorAll('tr')) : []
+      const colCount = ths.length || (rows[0] ? rows[0].children.length : 1)
+      const gridCols = 'grid-template-columns: ' + Array(colCount).fill('1fr').join(' ') + ';'
+
+      // Header as one block
+      if (ths.length > 0) {
+        const headerCells = ths.map(th => {
+          const styleAttr = th.getAttribute('style') || ''
+          return `<div class="tbl-hdr-cell" style="${styleAttr}">${th.innerHTML}</div>`
+        }).join('')
+        addBlock(`<div class="tbl-grid" style="${gridCols}">${headerCells}</div>`, 'thead', thead.textContent || '')
+      }
+
+      // Each body row as separate block — gets its own "+" button!
+      for (const tr of rows) {
+        const tds = Array.from(tr.querySelectorAll('td'))
+        const cells = tds.map(td => {
+          const styleAttr = td.getAttribute('style') || ''
+          return `<div class="tbl-cell" style="${styleAttr}">${td.innerHTML}</div>`
+        }).join('')
+        addBlock(`<div class="tbl-grid tbl-row" style="${gridCols}">${cells}</div>`, 'tr', tr.textContent || '')
+      }
+    }
+    // Split blockquotes — each paragraph inside is commentable
+    else if (tag === 'blockquote' && child.children.length > 1) {
+      for (const bqChild of child.children) {
+        addBlock('<blockquote>' + bqChild.outerHTML + '</blockquote>', 'blockquote-p', bqChild.textContent || '')
+      }
+    }
+    // Everything else — headings, paragraphs, pre, hr — one block each
+    else {
+      addBlock(child.outerHTML, tag, child.textContent || '')
+    }
+  }
+  return blocks
+}
--- a/web/src/utils/contentBlocks.js
+++ b/web/src/utils/contentBlocks.js
@@ -16,8 +16,15 @@ export function buildContentBlocks(rawContent) {
   div.innerHTML = html
   const blocks = []
 
-  function addBlock(html, tag, text) {
-    blocks.push({ index: blocks.length, html, tag, text })
+  function addBlock(html, tag, text, raw) {
+    const block = { index: blocks.length, html, tag, text }
+    // useDocumentComments.blockHash hashes `raw ?? html` to anchor inline
+    // comments; commentsForBlock hard-rejects on a hash mismatch, with no
+    // index fallback once a comment carries a hash. `raw` lets a block's
+    // rendered markup change (e.g. gaining `start=`) without re-hashing and
+    // silently detaching every comment already stored against it.
+    if (raw !== undefined) block.raw = raw
+    blocks.push(block)
   }
 
   for (const child of div.children) {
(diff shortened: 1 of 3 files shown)


unidoc-alip (changes requested):
> **Verdict: REQUEST_CHANGES** — 1 must-fix, 0 should-fix, 1 nit

The fix itself is right and I confirmed it end to end: `marked` really does emit one
correct `<ol>`, the per-item split really was dropping the position, threading `start`
through the wrapper really does restore it, and the attribute survives the read view's
`DOMPurify.sanitize(...)` call — so the rendered numbers will be correct in production.
The extraction into `web/src/utils/contentBlocks.js` is a clean, behaviour-preserving
move and the new tests are good ones.

What the tests cannot see is that `block.html` is not only rendered — it is also the
input to the inline-comment anchor hash. Changing it re-hashes every ordered-list block
and detaches every inline comment already stored against one. That is F1, and it has a
small, self-contained fix that the code already has a hook for.

Findings are keyed `F1`, `F2`, … (not `#1`, which GitHub would autolink to an unrelated issue).

---

### F1 [must-fix] `web/src/utils/contentBlocks.js` lines 30-38 — re-hashing detaches every existing inline comment on an ordered-list item

`useDocumentComments.js` anchors inline comments to a block by hashing the block's HTML:

```js
// web/src/composables/useDocumentComments.js:35-44
function blockHash(index) {
  const block = contentBlocks.value[index]
  if (!block) return ''
  const text = block.raw || block.html || ''
  let h = 5381
  ...
}
```

That hash is stored on the comment as `paragraph_hash` when it is created, and
`commentsForBlock` **rejects outright** on a mismatch — there is no index fallback once a
hash is present:

```js
// useDocumentComments.js:47-55
if (c.paragraph_hash && h) {
  if (c.paragraph_hash !== h) return false     // <- hard reject
  ...
}
```

Every ordered-list block's HTML changes from `<ol>…` to `<ol start="N">…` under this PR,
so every such block's hash changes. Comments already in the database against those blocks
stop matching: their inline badge and the "+"-anchored thread disappear from the item.
They are not deleted, and they still appear in the right-hand comments panel
(`Documents.vue:1002` lists comments independent of the hash), but clicking through no
longer resolves to the block and no new reply lands on the same thread inline. For
`documents/policy-acceptable-use.md` — the very document in #230, a 14-item numbered list
someone was reading closely enough to report a rendering bug on — that is exactly where
inline comments are most likely to already exist.

Confirmed by transplanting the real `blockHash` body verbatim into a harness and running
it over the base and head versions of `buildContentBlocks` for the same markdown: all 14
hashes differ, and `commentsForBlock` returns `[]` for a comment carrying the base hash.
Unordered lists, headings, paragraphs and table rows are unaffected — their HTML is byte
identical to before.

The fix is one line, and the codebase already has the hook for it: `blockHash` prefers
`block.raw` over `block.html`, and nothing currently sets `raw`. Set it to the exact
pre-PR wrapper string so the hash stays stable while the rendered HTML gains `start`:

```js
for (const li of child.children) {
  if (li.tagName.toLowerCase() === 'li') {
    const openTag = tag === 'ol' ? `<ol start="${start + position}">` : '<ul>'
    const block = { index: blocks.length, html: openTag + li.outerHTML + '</' + tag + '>',
                    tag: 'li', text: li.textContent || '' }
    // Comment anchors hash the block's markup (useDocumentComments.blockHash).
    // Keep hashing the wrapper WITHOUT `start`, so adding it doesn't re-hash
    // every ordered-list block and orphan comments already anchored to them.
    block.raw = '<' + tag + '>' + li.outerHTML + '</' + tag + '>'
    blocks.push(block)
    position++
  }
}
```

I ran this: with `raw` set, every head-side hash equals the base-side hash for the 14-item
list, the rendered `html` still carries `start="1"`…`start="14"`, and the suite stays
83/83. Worth a test asserting hash stability directly — the new test file is the natural
home, since this is the exact class of breakage a rendering-only test cannot catch.

Caveat on the evidence: the harness runs `buildContentBlocks` + a verbatim copy of
`blockHash` under jsdom. It does not exercise the live comment API, so "comments already
in the database" is a source-read conclusion about `commentsForBlock`'s reject branch, not
something the harness observed against real rows.

---

### F2 [nit] `web/src/utils/contentBlocks.js` line 31 — a list starting at 0 renumbers from 1

```js
const start = tag === 'ol' ? (parseInt(child.getAttribute('start'), 10) || 1) : 1
```

`0` is falsy, so a source list written `0. …` — valid CommonMark, and `marked` does emit
`<ol start="0">` for it (verified) — renders as 1, 2, 3 instead of 0, 1, 2. Cosmetic,
read-view only, and it takes an author deliberately numbering from zero, hence a nit
rather than a should-fix. If you want it:

```js
const parsed = parseInt(child.getAttribute('start'), 10)
const start = tag === 'ol' && Number.isFinite(parsed) ? parsed : 1
```

Verified: the three-item `0.` list then renders `start="0"`, `"1"`, `"2"`, and the suite
stays 83/83.

---

- "Each `<li>` ends up alone in a brand-new `<ol>` with no `start`, so every item renders
  as 1." — **confirmed.** The base `buildContentBlocks` emits a bare `<ol>` per item;
  `.doc-prose ol > li` uses native `list-style-type: decimal` (`web/src/style.css:79-80`),
  no CSS counters, so the browser genuinely restarts at 1 for each wrapper.
- "`marked` parses the 14-item list, blank line included, into one correct `<ol>`." —
  **confirmed.** Fed the shape to the app's own `parseMd`: a single `<ol>` with all items
  in order, loose (each `<li>` wrapping a `<p>`), no split at the blank line.
- "Only the read view is affected; tiptap renders the whole list as one WYSIWYG list." —
  **confirmed by source read**, not run: the editor path never calls
  `buildContentBlocks`, and `DocumentEditor.vue:812` styles `.tiptap ol` as one native
  decimal list. Not exercised in a harness.
- "The extraction is behaviour-preserving." — **partially confirmed.** The rendered output
  is identical for every non-ordered-list block, and the ordered-list change is the
  intended one. But the change is *not* inert with respect to comment anchoring — see F1.
- "Mutation-checked: reverting just the `start`-threading fails exactly 3 numbering tests,
  nothing else." — **confirmed.** Reverted the threading in a scratch copy: 3 failures,
  all three the numbering tests, 80 passing.
- "`npm test` 83/83 (76 existing + 7 new)." — **confirmed.** 83 pass, 0 fail; the new file
  contributes exactly 7.
- "`just build-web` clean." — **confirmed.** Ran the recipe itself (it is `cd web && npm run build`); completes with no errors.
- CI — the checks on this PR are `go-unit` and `integration`, with no separate web-test
  entry, which reads as though the new tests never run. They do: `npm test` is a step
  inside the `integration` job (`.github/workflows/tests.yml:110-113`), and that job
  passed. Noting it only because the check names are misleading.

---

## Appendix — hash-stability test for F1

Drop-in for `web/test/contentBlocks.test.js` (it needs no new imports — `test`, `assert`
and `buildContentBlocks` are already in scope there). Verified both ways against this
branch: it fails on the current head and passes once F1's `raw` line is applied, with the
suite at 84/84.

```js
// The block's markup is not only rendered — useDocumentComments.blockHash
// hashes it to anchor inline comments, and commentsForBlock hard-rejects on a
// hash mismatch. So adding `start` to the wrapper must not change what gets
// hashed, or every inline comment already stored against an ordered-list item
// silently detaches. `raw` is what keeps the hashed string stable.
test('ordered-list blocks hash the same as before start was threaded', () => {
  // verbatim from web/src/composables/useDocumentComments.js:35-44
  const blockHash = (block) => {
    const text = block.raw || block.html || ''
    let h = 5381
    for (let i = 0; i < text.length; i++) h = ((h << 5) + h + text.charCodeAt(i)) & 0xffffffff
    return h.toString(36)
  }
  const md = Array.from({ length: 14 }, (_, i) => `${i + 1}. Item ${i + 1}`).join('\n')
  for (const b of buildContentBlocks(md)) {
    // the pre-`start` wrapper shape, which is what the hash must still see
    assert.match(b.raw, /^<ol><li>/)
    assert.doesNotMatch(b.raw, /start=/)
    assert.match(b.html, /^<ol start="\d+">/)
    assert.equal(blockHash(b), blockHash({ html: b.raw }))
  }
})
```

unidoc-ahall:
Pushed a fix commit for F1 and F2.

**F1** — real, and a sharp catch: `blockHash` (`useDocumentComments.js:35-44`) hashes `raw ?? html`, and `commentsForBlock` hard-rejects with no index fallback once a comment carries a hash. `addBlock` now accepts an optional `raw` — the pre-`start` wrapper shape, exactly what you suggested — so the ordered-list split keeps hashing the stable pre-fix markup while `html` carries `start=` for rendering. Added your hash-stability test verbatim, plus one confirming unordered-list blocks (whose `html` never changes) get a `raw` that equals their `html`.

**F2** — switched `parseInt(...) || 1` to `Number.isFinite(parsedStart) ? parsedStart : 1`, so a genuine `start="0"` list renumbers from 0, not 1. Added the regression test.

Mutation-checked both independently: reverting just the `raw`-threading fails exactly 2 tests (hash-stability, ul-raw-equals-html), nothing else; reverting just the `Number.isFinite` change fails exactly the start=0 test, nothing else. Full suite 86/86 (83 prior + 3 new), `just build-web` clean.

unidoc-alip (approved):
> **Verdict: APPROVE** — both findings addressed and independently verified.

Re-reviewed at the follow-up commit. I validated the fixes by running the new head's files rather than reading the diff:

**F1 — resolved.** `addBlock` now takes an optional `raw` and the ordered-list split passes the bare pre-`start` wrapper. Re-ran the original PoC (verbatim `blockHash` + `commentsForBlock` transplant) against the new `contentBlocks.js`: a comment carrying the pre-PR hash for item 7 still attaches, `raw` is the bare `<ol><li>…` shape, and the sanitized `html` still carries `start="1"`…`"14"`. Applying `raw` uniformly inside `addBlock` rather than at the call site is better than what I suggested — I confirmed no other block type drifts: every non-list block has `raw === undefined`, and `ul` blocks have `raw === html`.

**F2 — resolved.** `Number.isFinite(parsedStart)` replaces the falsy `|| 1`; a `0.`-started list renders `start="0"`, `"1"`, `"2"`.

**Your verification claims all check out**, each re-run here: suite 86/86; reverting only the `rawWrapper` argument fails exactly the two hash/raw tests; reverting only `Number.isFinite` fails exactly the `start=0` test; `just build-web` clean at this head. The extra `ul`-raw test is a good addition beyond what I proposed.

One thing left standing from the original review, unchanged and not blocking: "only the read view is affected; tiptap renders the whole list as one WYSIWYG list" remains a source-read conclusion — the editor path never calls `buildContentBlocks` and `DocumentEditor.vue:812` styles `.tiptap ol` as one native decimal list — but no harness exercised it.

PR #239, opened 2026-09-01
test(e2e): fix the flaky review-approval test that was randomly failing CI

unidoc-alip:
## What was going wrong

Our automated test suite has been failing at random on the `master` branch — the same test passing one run and failing the next, with no code change in between. Most recently it blocked the build with:

> `POST /reviews/11/merge: 400 review must be approved before merging`

That kind of intermittent red build is expensive: it makes people distrust CI, it prompts "just re-run it" as a reflex, and it hides the moment a *real* regression shows up.

**There is no bug in the product.** Reviews, approvals and merges all work correctly. The defect was in the tests themselves.

## Why it happened

The affected tests drive a real browser: a reviewer logs in, opens a document review, and clicks Approve. Each test was then supposed to wait for the on-screen confirmation before continuing.

They waited for the wrong thing. They watched for wording that is already on the page before anyone clicks anything — a "Round 2" badge in the header in one test, and a permanent "Approved: 1 of 2" counter in the review sidebar in the other. Those matched instantly, so the waits returned without actually waiting for the approval to finish.

Each test then closed the browser, which cut off the approval request mid-flight, and went on to merge a review that had never been approved. Hence the error.

Whether this failed came down to raw timing — on a fast CI machine the approval squeaked through, on a loaded one it did not. That is exactly why it looked random.

## What this changes

Three things, all confined to the test file — no product code is touched:

1. Each test now waits for wording that only appears *after* an approval genuinely succeeds: the "ready to merge" confirmation in one, and the reviewer's own "You approved this round" message in the other. Both are driven by the reviewer's real recorded decision, so neither can appear early.
2. Before merging, the tests now ask the server directly whether the review is approved, and wait until it is. Checking actual system state rather than what happens to be rendered on screen makes them immune to this whole class of timing failure.
3. Both affected tests get both protections, so neither can fail this way again.

The second change also future-proofs the suite: our in-flight internationalisation work is rewriting a lot of on-screen wording, and tests that hinge on exact English phrasing would start breaking as those translations land.

## Risk and validation

Low risk — test-only, no shipped behaviour changes. CI on this branch is the validation: the previously flaky tests should now pass consistently.

### DIFF
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -286,7 +286,10 @@ class TestReviewTwo:
             page.get_by_role("button", name="Approve").click()
             page.locator('button:has-text("Confirm Approve")').wait_for(state="visible", timeout=5000)
             page.locator('button:has-text("Confirm Approve")').click()
-            page.locator("text=/recorded|[Aa]pproved/").first.wait_for(state="visible", timeout=10000)
+            # Per-reviewer confirmation, rendered only once this user's own
+            # assignment status flips to approved. NOT the sidebar's static
+            # "Approved: n of m" label, which is on the page from first paint.
+            page.locator("text=You approved this round.").first.wait_for(state="visible", timeout=10000)
         finally:
             ctx.close()
 
@@ -300,7 +303,10 @@ class TestReviewTwo:
             page.get_by_role("button", name="Approve").click()
             page.locator('button:has-text("Confirm Approve")').wait_for(state="visible", timeout=5000)
             page.locator('button:has-text("Confirm Approve")').click()
-            page.locator("text=/recorded|[Aa]pproved/").first.wait_for(state="visible", timeout=10000)
+            # Per-reviewer confirmation, rendered only once this user's own
+            # assignment status flips to approved. NOT the sidebar's static
+            # "Approved: n of m" label, which is on the page from first paint.
+            page.locator("text=You approved this round.").first.wait_for(state="visible", timeout=10000)
         finally:
             ctx.close()
 
--- a/tests/test_e2e_browser.py
+++ b/tests/test_e2e_browser.py
@@ -4,6 +4,7 @@ Requires: pip install playwright && playwright install chromium
 Run:      pytest tests/test_e2e_browser.py -v
 """
 import os, pytest, importlib
+import time
 import uuid
 import requests
 
@@ -69,6 +70,23 @@ def click_sidebar(page, label):
 
 def wait_for(page, text, timeout=8000):
     page.locator(f"text={text}").first.wait_for(state="visible", timeout=timeout)
+
+def wait_review_status(rid, token, status, timeout=15.0):
+    """Poll the API until a review reaches `status`.
+
+    Browser-driven approvals fire an async POST; a UI text assertion can pass
+    before the request lands (or while it is still in flight, in which case
+    closing the browser context cancels it server-side). Always gate a
+    follow-up API call on the real state, never on rendered copy.
+    """
+    deadline = time.time() + timeout
+    last = None
+    while time.time() < deadline:
+        last = api("get", f"/reviews/{rid}", token, expect_status=200).json().get("status")
+        if last == status:
+            return
+        time.sleep(0.25)
+    raise AssertionError(f"review {rid} never reached status {status!r} (last: {last!r})")
 
 # ── Fixtures ──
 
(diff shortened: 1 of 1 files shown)


unidoc-ahall (changes requested):
Verified the root-cause diagnosis independently — it's correct, and more solidly grounded than the description even claims. Found one real gap in the sibling test, though.

**Confirmed the "Round 2" diagnosis in the actual component**, not just the description: `Reviews.vue` renders `Round {{ review.round || 1 }}` unconditionally in the sidebar/header (lines 46, 400, 437) whenever `review.status !== 'merged' && 'closed'` — so on a review already at round 2, that text is on the page from the very first paint, well before any approval click. `approvalFeedback.value` (the "Ready to merge"/"has been recorded" text the fix now waits for) is only ever set inside `submitApproval()`, after `await api.approveReview(...)` resolves — genuinely gated on the real request completing. Good fix.

**Went further and checked whether the old regex's other branches (`approved|Approved`) were also unsafe** — they are: `Reviews.vue:406` renders a static `Approved:` label (`{{ assignments.filter(a => a.status === 'approved').length }} of {{ assignments.length }}`) in the same always-shown sidebar block, present on page load regardless of anyone's status. So the old regex was fragile on two independent axes, not just the one the description calls out.

**The gap:** `test_both_approve`'s own two waits — `page.locator("text=/recorded|[Aa]pproved/")` at lines 289 and 303, for R1's and R2's individual approvals — still use that same unsafe pattern, and weren't touched by this PR. Since `Approved:` is present from first paint, both waits resolve essentially immediately regardless of whether that reviewer's click-triggered approve POST has landed, which is exactly the failure mechanism this PR's own description identifies ("the test closed the browser, which cut off the approval request mid-flight"). The `wait_review_status(rid, t, "approved")` this PR adds before the merge step is a real safety net — it'll turn a lost approval into a clear timeout `AssertionError` instead of a cryptic 400 — but it doesn't prevent the race at the point each reviewer's own context closes, so `test_both_approve` can still fail intermittently for close to the original reason, just with a better error message when it does.

There's already the right text to wait on instead: `Reviews.vue:567`, gated on `userAlreadyActed && userAssignmentStatus === 'approved'` — "You approved this round." is specific to the acting reviewer and only appears after their own assignment status actually flips, unlike the shared `Approved:` counter. Swapping lines 289 and 303 to wait for that (or an equivalent per-user confirmation) would close this the same way `test_round_ui` was closed, rather than relying on the downstream poll to catch it after the fact.

The description's "the same fix is applied to one sibling test built the same way" undersells what actually happened there — `wait_review_status` was added to both, but the text-wait fix itself only landed in `test_round_ui`.

Everything else here is solid — `wait_review_status`'s polling logic is sound, and it's a good addition regardless. Requesting changes for the two remaining unsafe waits, not the overall approach.

unidoc-alip:
Validated the finding independently — it's correct, and I've addressed it in 3546e0b.

Confirmed in the component: `Reviews.vue:406` renders `<span>Approved:</span> {{ ... }} of {{ ... }}` inside the round block at line 399, whose only guard is `review.status !== 'merged' && review.status !== 'closed'`. Nothing about that is conditional on anyone having approved, so `Approved:` is in the DOM from first paint and `text=/recorded|[Aa]pproved/` matched it immediately. Both waits in `test_both_approve` were no-ops for the same reason `test_round_ui`'s was — I fixed the symptom I'd traced in the CI log and missed that the sibling was broken on a different branch of its regex.

Took your suggested target. Both waits now use `text=You approved this round.` (`Reviews.vue:567`), gated on `userAlreadyActed && userAssignmentStatus === 'approved'`. Checked the backing computeds at lines 869-877: both read `assignments`, which `submitApproval()` refreshes via `loadReviewDetail()` only after `await api.approveReview(...)` resolves — so the text cannot render before that reviewer's own request has landed, and unlike the shared counter it is specific to the acting user.

Also fixed the description, which claimed the text-wait fix landed in both tests when only `wait_review_status` had. It now describes both waits and both mechanisms accurately.

Left `test_round_ui`'s wait as-is per your read of it — `approvalFeedback` is only ever assigned inside `submitApproval()` after the await, so it is genuinely gated.

unidoc-ahall (approved):
Independently verified the fix for the gap, not just the explanation.

Checked the `v-if`/`v-else-if` chain in `Reviews.vue` directly: `<template v-if="review.status === 'approved'">` (line 504) precedes `<template v-else-if="userCanReview">` (line 563, where "You approved this round." lives). So for R2 — whose approval is the one that flips the review to `approved` — the reviewer-actions branch stops rendering entirely and the "approved" branch takes over instead; "You approved this round." genuinely cannot appear for them, confirming it would have been the wrong wait target for that case. The sidebar's "Ready to merge" (line 414) is in a separately-gated block (`review.status !== 'merged' && != 'closed'`) that isn't scoped by role, so it's visible to R2 regardless of their `canWrite` permission — a correct, general target.

Both fixes are sound: R1 waits on the per-user confirmation, R2 waits on the status-driven sidebar text, and both are gated on real state rather than pre-existing page content. `wait_review_status` before the merge step in both tests is the right belt-and-suspenders addition on top.

Approving.

PR #242, opened 2026-09-04
fix(web): send org slug on password login

### LINKED ISSUE #241: Password login drops org context for multi-org users (path-based deployments)
## Problem

On a path-based deployment (`ISMS_SUBDOMAIN_ROUTING=0`), a user who belongs to more than one organization can get bounced back to the login page / org picker after entering a correct email and password, instead of landing in the org they meant to log into.

Surfaced while investigating #240: the demo link itself turned out to be correct (`demo.isms.sh/<slug>` is the right URL for a path-based deployment — no subdomain auto-resolve exists or is needed there), but logging in as the multi-org demo account hit this bug and looped back to login.

## Root cause

Password login (`web/src/api.js`'s `login()`, called from `Login.vue`'s `handleLogin()`) never included the org slug in the `POST /auth/login` body, even though the backend's `loginRequest` struct has an `organization` field for exactly this (`internal/isms/api/api_auth.go`). The OIDC login path already sends it (`&org=<slug>` in `Login.vue`'s `oidcLogin()`) — password login was the one path missing it.

Without it, the backend falls back to `org_slug` from `OrgResolverMiddleware`, which is never set for `/api/` requests by design, then falls back further to auto-selecting an org only if the user belongs to exactly one. A multi-org user gets a token with no org context. The client then notices the mismatch and tries `switch-org` to recover — which only works if the user is actually a member of the org named in the URL; otherwise it lands them on the org picker or back on `/login`, discarding the original destination.

Separately, `goToOrg()` (the org-discovery step on the bare `/login` screen) does a hard `window.location.href` redirect on subdomain-capable hosts and drops the original `?redirect=` query param in the process — a second way to lose the destination, independent of the above.

## Fix

- Send `organization: <slug>` in the password-login request when the org is known, mirroring the OIDC path.
- Preserve `?redirect=` across the org-discovery hard-redirect.

## Target

Fixed for 0.8.0 — the demo will work end-to-end again by then.

unidoc-ahall:
Fixes #241.

Password login (`web/src/api.js`'s `login()`) never sent the org slug in the `POST /auth/login` body, unlike the OIDC path which already does (`Login.vue`'s `oidcLogin()`). The backend has an `organization` field on `loginRequest` (`internal/isms/api/api_auth.go`) for exactly this — it just never received a value from the password-login form.

Without it, a user in more than one org couldn't be resolved: the backend's org auto-select only works when the user belongs to exactly one org, so a multi-org user got a token with no org context, and the client's mismatch-recovery (`switch-org`) only succeeds if they're actually a member of the org in the URL — otherwise they land back on the org picker or `/login`, having lost the page they were trying to reach.

Two changes:
1. `login()` now takes an `organization` argument and includes it in the request body when known; `handleLogin()` passes `orgSlug.value`.
2. `goToOrg()`'s hard redirect (org-discovery step, subdomain-capable hosts) now preserves `?redirect=` instead of dropping it.

New test: `web/test/apiLogin.test.js` — asserts `login()` includes `organization` in the request body when passed and omits it when not. Reverting the `api.js` change makes it fail with the exact symptom (`organization` missing from the body); confirmed locally.

`web/` unit suite: 113/113 passing.

### DIFF
--- a/web/src/api.js
+++ b/web/src/api.js
@@ -149,9 +149,10 @@ async function uploadFile(url, file, extraFields = {}) {
 // enabled, the first call (without `otp`) returns `{otp_required: true}` with
 // an empty token — the caller must prompt for the code and call login again
 // with the otp argument.
-async function login(email, password, otp) {
+async function login(email, password, otp, organization) {
   const body = { email, password }
   if (otp) body.otp = otp
+  if (organization) body.organization = organization
   const res = await fetch(`${API}/auth/login`, {
     method: 'POST',
     headers: { 'Content-Type': 'application/json' },
--- a/web/src/views/Login.vue
+++ b/web/src/views/Login.vue
@@ -265,6 +265,13 @@ function base64urlToBuffer(base64url) {
   return bytes.buffer
 }
 
+// route.query.redirect is a string, unless the URL repeats ?redirect=, in
+// which case vue-router gives an array — normalize to the first value.
+function firstRedirect(query) {
+  const r = query.redirect
+  return Array.isArray(r) ? r[0] : r
+}
+
 function getOrgSlug() {
   // 1. URL query param (?org=acme)
   const params = new URLSearchParams(window.location.search)
@@ -277,6 +284,17 @@ function getOrgSlug() {
   // 3. Route path (/acme/login → acme) — only used on apex/localhost
   const pathParts = window.location.pathname.split('/').filter(Boolean)
   if (pathParts.length >= 2 && pathParts[1] === 'login') return pathParts[0]
+
+  // 4. The org implied by a bounced destination (?redirect=/acme/documents).
+  //    The router guard sends deep links to the bare /login, so this is the
+  //    only place the org survives. Ask the router rather than splitting the
+  //    path: it is the authority on which routes are org-scoped, and in
+  //    subdomain mode none of them are.
+  const redirect = params.get('redirect')
+  if (redirect) {
+    const org = router.resolve(redirect).params.org
+    if (org) return org
+  }
 
   // No localStorage fallback — org context must come from URL/subdomain only.
   // Anything else leaks state between orgs (e.g. visiting isms.sh after being
@@ -292,7 +310,7 @@ function goToOrg() {
   // If the host can serve tenant subdomains (apex like isms.sh or already on a
   // subdomain), hop to <slug>.<apex>/login. Otherwise stay path-based.
   if (canHostSubdomain(window.location.hostname)) {
-    const redirectPath = route.query.redirect
+    const redirectPath = firstRedirect(route.query)
     const suffix = '/login' + (redirectPath ? `?redirect=${encodeURIComponent(redirectPath)}` : '')
     window.location.href = orgEntryURL(slug, suffix)
     return
@@ -303,7 +321,7 @@ function goToOrg() {
 
 async function redirectAfterLogin() {
   // Check for a redirect query param first
-  const redirectPath = route.query.redirect
+  const redirectPath = firstRedirect(route.query)
   if (redirectPath && redirectPath !== '/overview' && redirectPath.startsWith('/')) {
     router.push(redirectPath)
     return
--- a/web/src/views/Login.vue
+++ b/web/src/views/Login.vue
@@ -292,7 +292,9 @@ function goToOrg() {
   // If the host can serve tenant subdomains (apex like isms.sh or already on a
   // subdomain), hop to <slug>.<apex>/login. Otherwise stay path-based.
   if (canHostSubdomain(window.location.hostname)) {
-    window.location.href = orgEntryURL(slug, '/login')
+    const redirectPath = route.query.redirect
+    const suffix = '/login' + (redirectPath ? `?redirect=${encodeURIComponent(redirectPath)}` : '')
+    window.location.href = orgEntryURL(slug, suffix)
     return
   }
   // Path-based — set slug locally and re-render the login form in org context.
@@ -428,7 +430,7 @@ async function handleLogin() {
   error.value = ''
   loading.value = true
   try {
-    const res = await login(email.value, password.value, otpCode.value || undefined)
+    const res = await login(email.value, password.value, otpCode.value || undefined, orgSlug.value || undefined)
     if (res.otp_required) {
       // Backend wants a TOTP code — show the input and stop here.
       otpRequired.value = true
--- a/web/test/apiLogin.test.js
+++ b/web/test/apiLogin.test.js
@@ -1,0 +1,46 @@
+import test from 'node:test'
+import assert from 'node:assert/strict'
+import { JSDOM } from 'jsdom'
+
+const dom = new JSDOM('<!doctype html><html><body></body></html>', { url: 'https://acme-logistics.example/login' })
+globalThis.window = dom.window
+globalThis.document = dom.window.document
+globalThis.localStorage = dom.window.localStorage
+
+// Regression: on a path-based (non-subdomain) deployment, a user who belongs
+// to more than one org would get bounced back to the org picker after a
+// correct password login, because the request never told the backend which
+// org they meant to log into — it silently relied on the backend's
+// exactly-one-org auto-select fallback. Login.vue knows the org (the visitor
+// typed or navigated it in), it just wasn't being sent.
+test('login() sends the org slug when one is known', async () => {
+  let capturedBody = null
+  globalThis.fetch = async (url, opts) => {
+    capturedBody = JSON.parse(opts.body)
+    return {
+      ok: true,
+      json: async () => ({ token: 't', email: 'a@b.com' }),
+    }
+  }
+
+  const { login } = await import('../src/api.js')
+  await login('a@b.com', 'pw', undefined, 'acme-logistics')
+
+  assert.equal(capturedBody.organization, 'acme-logistics')
+})
+
+test('login() omits organization when none is known', async () => {
+  let capturedBody = null
+  globalThis.fetch = async (url, opts) => {
+    capturedBody = JSON.parse(opts.body)
+    return {
+      ok: true,
+      json: async () => ({ token: 't', email: 'a@b.com' }),
+    }
+  }
+
+  const { login } = await import('../src/api.js')
+  await login('a@b.com', 'pw', undefined, undefined)
+
+  assert.equal('organization' in capturedBody, false)
+})
--- a/web/test/loginOrgSlug.test.js
+++ b/web/test/loginOrgSlug.test.js
@@ -1,0 +1,99 @@
+// Regression coverage for #241 review finding F1/F2: the router guard bounces
+// an unauthenticated deep link (e.g. /acme/documents) to the BARE /login with
+// ?redirect=/acme/documents — not /acme/login — so Login.vue's getOrgSlug()
+// has to recover the org from the redirect target, and the org it recovers
+// has to actually reach the login request (Login.vue:handleLogin -> api.js's
+// login()), not just live in a local variable.
+//
+// getOrgSlug() is transplanted verbatim (see router.test.js for the same
+// convention and its caveat: this is a copy, not an import — Login.vue
+// imports .vue components and browser globals that this harness doesn't
+// load under `node --test`). login() itself is the REAL implementation from
+// src/api.js, imported normally, so the request-body assertion below is not
+// testing a copy on both ends.
+import test from 'node:test'
+import assert from 'node:assert/strict'
+import { JSDOM } from 'jsdom'
+import { createRouter, createMemoryHistory } from 'vue-router'
+
+const dom = new JSDOM('<!doctype html><html><body></body></html>', { url: 'https://isms.example/login' })
+globalThis.window = dom.window
+globalThis.document = dom.window.document
+globalThis.localStorage = dom.window.localStorage
+
+const C = { render: () => null }
+function buildRouter(subdomainMode) {
+  const orgRoutes = subdomainMode
+    ? [{ path: '/documents', component: C }, { path: '/documents/:docId', component: C }]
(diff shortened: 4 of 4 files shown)


unidoc-alip (changes requested):
> **Verdict: REQUEST_CHANGES** — 2 should-fix, 2 nits

This PR closes the gap it names: password login now sends `organization`, matching what
`oidcLogin()` has always done with `&org=`, and the org-discovery hard redirect no longer
drops `?redirect=`. I traced the whole entry path — router guard, `getOrgSlug()`,
`goToOrg()`, `orgEntryURL()`, `handleLogin()` — and confirmed the #241 flow reaches the
backend with an org slug after this change. Backend side checks out too: `handleLogin`
resolves an explicit `organization` before falling back to the exactly-one-org
auto-select, so a multi-org user now gets a scoped token.

What follows is one gap in coverage of the reported flow and one gap in the test — I'd
like both addressed before this merges — plus two smaller things. Findings are keyed `F1`, `F2`, … to avoid
GitHub autolinking `#1` to an unrelated issue.

---

### F1 [should-fix] `web/src/views/Login.vue` line 266 — `getOrgSlug()` still ignores the org sitting in `?redirect=`

The #241 entry path is: unauthenticated visitor opens `demo.isms.sh/<slug>/documents`, the
router guard bounces them to `{ path: '/login', query: { redirect: to.fullPath } }`
(`web/src/router.js:205,223`) — note the **bare** `/login`, not `/<slug>/login`.

`getOrgSlug()` then looks in three places: `?org=`, the subdomain, and
`pathParts[1] === 'login'`. On `/login?redirect=/acme/documents` all three miss and it
returns `''`. Because the template gates the password form behind `v-if="!orgSlug"`
(`Login.vue:5`), the visitor is shown "Enter your organization to continue" and has to
re-type the slug that the app is already holding in `?redirect=`.

Change #2 in this PR is what makes that survivable — after `goToOrg()` the redirect is
carried to `/<slug>/login?redirect=…` and the fix fires. So the bug is fixed; the
detour just isn't necessary. Adding a fourth source to `getOrgSlug()` removes it:

```js
  // 4. The org implied by a bounced destination (?redirect=/acme/documents).
  //    The router guard sends deep links to the bare /login, so this is the
  //    only place the org survives. Ask the router rather than splitting the
  //    path: it is the authority on which routes are org-scoped, and in
  //    subdomain mode none of them are.
  const redirect = params.get('redirect')
  if (redirect) {
    const org = router.resolve(redirect).params.org
    if (org) return org
  }
```

`router` is already in scope at `Login.vue:226`. Going through `resolve()` rather than
taking the first path segment matters in two places: `redirect=/organizations` would
otherwise yield the slug `organizations`, and in subdomain mode the org-scoped routes are
mounted at the top level (`router.js:79-88`), so `redirect=/documents` would yield
`documents`. Either would turn a working login into a `400 organization not found`.
`resolve().params.org` is empty in both cases — verified against a memory router built
the same way `web/test/router.test.js` builds one.

Confirmed by source read plus a transplant harness (see the PoC note at the end); the
`?redirect=`-derived slug is **not** exercised by any test today.

Related, and worth a separate look rather than a change here: `useSession.js:74` only
attempts `switch-org` when `me?.organization_slug` is truthy. A token with
`organization_id: 0` has an empty slug, so the recovery the PR description credits
(`switch-org`) is in fact never attempted for exactly the user this PR is about. The
`switch-org` handler itself (`api_auth.go:395`) needs only an authenticated email and
would work fine from an org-less token. Dropping the truthiness guard would make the
recovery real, independent of how the user reached `/login`.

---

### F2 [should-fix] `web/test/apiLogin.test.js` line 16 — the test guards the plumbing, not the bug

Both cases import `../src/api.js` and call `login()` directly. That pins the two lines in
`api.js`, and the description's claim about reverting them is accurate. But the actual
defect in #241 was in the **caller**: `handleLogin()` never passed the slug it already
had. Revert only the `Login.vue:433` hunk and the suite is still 113/113 green — the
regression this PR fixes is unguarded.

`web/test/router.test.js` already stands up a memory router with stubbed components and
API, so the scaffolding to assert "entering a slug then submitting the form results in a
request body carrying `organization`" exists in the repo. A case at that level is what
would actually fail on a revert.

Smaller point in the same file: both tests `await import('../src/api.js')`, and ESM
module caching means the second call gets the first module instance. It works here
because `fetch` is read off `globalThis` at call time, but it makes the file fragile to
any future module-level capture in `api.js`.

---

### F3 [nit] `web/src/views/Login.vue` line 295 — `route.query.redirect` can be an array

`?redirect=/a&redirect=/b` gives vue-router an array, and
`encodeURIComponent(['/a','/b'])` produces `%2Fa%2C%2Fb` — a single nonsense path. The
same shape already breaks `redirectAfterLogin()` (`redirectPath.startsWith` on an array),
so this is a pre-existing pattern rather than something the PR introduces. A shared
`const redirectPath = Array.isArray(route.query.redirect) ? route.query.redirect[0] : route.query.redirect`
would settle both sites.

---

### F4 [nit] Sending the org turns some previously-succeeding logins into a hard 403

`handleLogin` (`internal/isms/api/api_auth.go:125-136`) treats an explicit `organization`
as authoritative: unknown slug → `400 organization not found`, non-member →
`403 not a member of this organization`. Before this PR a single-org user who arrived on
another org's login page (a shared link, a stale bookmark, a typo in the discovery box)
was auto-selected into their own org; now they get an inline error.

The new behaviour is the more honest of the two — the old one led to `/organizations` via
a mismatch bounce anyway — so this is not a regression to undo. It is only that the login
form offers no way out of the 403: no "sign in to a different organization" affordance to
clear `orgSlug` and return to the discovery step. Worth a follow-up.

---

- **"`login()` never sent the org slug, unlike the OIDC path"** — **confirmed.**
  `oidcLogin()` (`Login.vue:353`) has always appended `&org=<slug>`; `login()` built a body
  of `{ email, password }` plus optional `otp` only.
- **"The backend has an `organization` field on `loginRequest` for exactly this"** —
  **confirmed.** `api_auth.go:37`, consumed at line 125 ahead of the auto-select fallback.
- **"org auto-select only works when the user belongs to exactly one org"** —
  **confirmed.** `api_auth.go:137-146`; `orgID` stays `0` for 0 or 2+ orgs.
- **"the client's mismatch-recovery (`switch-org`) only succeeds if they're actually a
  member of the org in the URL"** — **partially confirmed.** True of the handler
  (`api_auth.go:407`), but understated on the client: `useSession.js:74` requires
  `me?.organization_slug` to be non-empty, and it is empty precisely when `orgID` is `0`.
  For the multi-org user this PR is about, `switch-org` is never even attempted. See F1.
- **"`goToOrg()`'s hard redirect now preserves `?redirect=`"** — **confirmed**, and
  `orgEntryURL()` handles the query-bearing suffix correctly in both the path-based and
  the subdomain branch (`useCurrentOrg.js:199`).
- **"Reverting the `api.js` change makes it fail with the exact symptom"** —
  **confirmed** by running a reverted copy: `login() sends the org slug when one is known`
  fails on `capturedBody.organization` being `undefined`. See also F2 — the `Login.vue`
  half of the change has no such guard.
- **"`web/` unit suite: 113/113 passing"** — **confirmed** locally on `25db95e`
  (`cd web && npm test` → 113 pass, 0 fail). Worth noting for the record that this does
  run in CI: `tests.yml` has a "Web unit tests" step inside the `integration` job (added
  in #223), and `integration` is green on this PR.

## How the findings were checked

F1 and F3 were confirmed with a standalone transplant harness: the real `getOrgSlug()`,
the real router-guard redirect expression, the real `goToOrg()` body and the real
`login()` body-builder copied into bare Node with only `window` and `route.query`
stubbed. It runs base / head / suggested-fix over six scenarios and all 13 assertions
matched:

- unauthenticated deep link `/acme/documents` — base and head both produce
  `slug=(none), organization=(absent)`; the F1 fix produces `slug=acme, organization=acme`
- direct `/acme/login` — all three produce `slug=acme` (the path already works)
- `redirect=/organizations` — the `resolve()`-based fix yields no slug, while the
  first-path-segment shortcut yields `slug=organizations`
- subdomain mode, `redirect=/documents` — the `resolve()`-based fix yields no slug, while
  the shortcut yields `slug=documents`; these two cases are why F1 suggests
  `router.resolve()` rather than string splitting
- `?org=beta&redirect=/acme/...` — the explicit `?org=` still wins
- `goToOrg` — base emits `/acme/login`, head emits
  `/acme/login?redirect=%2Facme%2Fdocuments` (change #2 works), and an array-valued
  `redirect` emits `/acme/login?redirect=%2Fa%2C%2Fb` (F3)

The description's revert claim was checked by running the new test against a copy of
`api.js` with the added line deleted: `login() sends the org slug when one is known`
fails with `actual: undefined, expected: 'acme-logistics'`.

Caveats: the harness does not mount Vue, so it observes the value of `orgSlug` and the
outgoing request body, not what is rendered. The "the visitor is shown the discovery
step" statement in F1 comes from reading the `v-if="!orgSlug"` template gate, not from
the harness. F2 and F4 are source-read only (for F2: no test file imports `Login.vue` —
the two matches in `web/test/` are both comments).

unidoc-ahall:
Fixed F1, F2, F3.

**F1** — added the 4th source to `getOrgSlug()` exactly as suggested: `router.resolve(redirect).params.org`, not a path-segment split, specifically because `redirect=/organizations` and (in subdomain mode) `redirect=/documents` would otherwise misread as org slugs. `router` was already in scope.

**F2** — added `web/test/loginOrgSlug.test.js`: covers the redirect-derived org for the plain case, both misread-guard cases (`/organizations`, subdomain `/documents`), the `?org=` precedence case, and — the actual regression surface — asserts the resolved slug reaches the real `login()` request body, not just a copy of the slug logic. Mutation-tested by breaking the transplanted source-4 logic locally: the deep-link and request-body tests fail as expected, the other three don't. `web/` suite is 118/118.

**F3** — added `firstRedirect()`, used by both `goToOrg()` and the pre-existing `redirectAfterLogin()` (same unguarded array-access bug, not new to this PR but same fix serves both).

**F4** — agreed it's a real gap, but leaving it as a follow-up rather than folding into this PR: it's an independent UX affordance (recovering from a 403 on the login form), not part of the #241 flow this PR closes.

Also left the `useSession.js:74` truthiness-guard note as-is per your framing — worth a separate look, not this PR.

unidoc-alip (approved):
> **Verdict: APPROVE** — F1 and F3 confirmed fixed, F2 partly; one correction to the record and three follow-ups

Re-verified everything on `5d50d08`.

---

### F1 — confirmed fixed, and correct against the real router

`Login.vue:288-296` adds source 4 as suggested, via `router.resolve(redirect).params.org`. I checked
the misread guard against the real route table rather than only the test's: `router.js:96` declares
`/organizations` ahead of `...buildOrgRoutes()` at line 98, and there is deliberately no bare `/:org`
route and no catch-all (the comment at `router.js:105-110` explains why). So
`resolve('/organizations').params.org` is empty, and so is `resolve('/documents').params.org` in
subdomain mode — the two cases that would have turned a working login into a
`400 organization not found` under a path-segment split genuinely cannot fire. `router` is in scope at
line 225, and both `getOrgSlug()` call sites (490, 525) run after it.

### F3 — confirmed fixed

`firstRedirect()` (`Login.vue:268-273`) is used at both `goToOrg()` (313) and `redirectAfterLogin()`
(324). `grep -rn "query.redirect" src/` returns only the helper and its comment — no raw accesses left
anywhere in the login path.

### F2 — the fix is real but the claim overshoots, and my original ask was on a wrong premise

Two things, and the second is mine to correct.

The comment and the commit message both say the new test asserts the slug reaches the real `login()`
request body and so covers "the actual regression surface (Login.vue's caller wiring)". `login()` is
indeed the real import — but the slug handed to it comes from the *transplanted* `getOrgSlug()` inside
the test file. `handleLogin()` is never on the call stack, so the caller wiring is not what is being
covered. Reverting the `Login.vue:451` hunk to `login(email.value, password.value, otpCode.value || undefined)`
— i.e. reintroducing the #241 bug verbatim — still gives 118/118 green. Breaking the real source-4 line
in `Login.vue` is likewise green. You were explicit that the mutation test broke the *transplant*, so
this is a wording correction, not a disputed result.

The premise of my original F2 was wrong, though. I pointed at `web/test/router.test.js` as existing
scaffolding that could reach `handleLogin()`. It cannot: `web/package.json` has no `@vue/test-utils`
and no vitest — the suite is bare `node --test` plus jsdom — and `router.test.js` stubs every
component, `Login` included, carrying the same transplant caveat it picked up as F4 on #223. There is
no component-mounting infrastructure in this repo, so closing F2 properly is a test-infra change, not
something a bug-fix PR should absorb. Withdrawing it as a blocker.

### F4 and `useSession.js:74`

Both deferrals accepted — they were raised as nits and follow-ups, and neither is part of the #241 flow
this PR closes.

---

## Verification performed

- `cd web && npm test` on `5d50d08` — **118/118**, matching the claim.
- Mutation A, revert `Login.vue:451`'s `orgSlug.value || undefined` argument — 118 pass, 0 fail.
- Mutation B, `const org = null` in place of the real `router.resolve(redirect).params.org` — 118 pass, 0 fail.
- Route-ordering read of `router.js:72-112` to confirm the two misread cases against the shipped table,
  not just the memory router the test builds.
- CI on this head: `go-unit` pass, `integration` pass (the job that runs the web suite).

## Suggested follow-ups, none blocking

1. Add `@vue/test-utils` and one `handleLogin`-level test. It is the single case that would fail on a
   revert of the original fix, and it would also retire the transplant caveat now duplicated across
   three files in `web/test/`.
2. F4 — an affordance on the login form to clear `orgSlug` and return to org discovery, so a 403 from
   an explicit non-member `organization` is recoverable.
3. `useSession.js:74` — drop the `me?.organization_slug` truthiness guard so `switch-org` is actually
   attempted for an org-less token.

Good to merge as it stands.

PR #245, opened 2026-09-05
feat(i18n): stable error codes on API responses (#212)

unidoc-alip:
Groundwork for translating the error messages users see when something goes wrong. Part of #212, which stays open through the pt-BR contribution.

## The problem

When the app refuses to do something, it says why: *"risk not found"*, *"title is required"*, *"you cannot approve your own review"*. There are about 620 of these sentences, and every one of them is English, hardcoded, and unreachable by a translator.

That matters more than it sounds. A user browsing an Indonesian interface hits an error and the app switches to English at precisely the moment they most need to understand it — the moment they are already confused. Error messages are the part of an interface people read most carefully and trust least when they can't.

The reason none of them are translated is structural rather than neglectful: the server writes the sentence, and the server has no idea who is going to read it. The same message goes to a browser, to the command-line tool, and to the server's own log file.

## What this changes

The server stops writing the sentence and starts naming the situation.

Instead of sending the words *"risk not found"*, it now sends a label — `not_found` — plus the piece of information that varies, `risk`. The browser, which does know who is reading, looks up that label in its own language and renders the sentence. The same label produces "risk not found" for an English reader and "risiko tidak ditemukan" for an Indonesian one.

The English sentence does not go away. It is sent alongside the label, exactly as before, because the command-line tool and the server log have no translations to consult and never will. **Anything reading these responses today keeps working with no change**, which is what makes this safe to land while the rest of the work is still in flight.

## What is and isn't in this change

**In:** the mechanism, a catalogue of 34 named situations covering the most common errors, and their English and Indonesian text.

**Not in:** actually switching the 620 call sites over. That is deliberate — the mechanism and the conversion are separately reviewable, the conversion happens module by module as those files are touched anyway, and the two can proceed at different speeds. Nothing users see changes in this PR.

Also not everything gets translated, on purpose:

- **Internal faults stay English.** About 90 of those 620 are messages like *"creating user: connection refused"* — a server malfunction, not something a user can act on. Translating them would help nobody and would obscure the diagnosis.
- **Deployment messages stay English.** *"email not configured (SMTP_HOST)"* is read by whoever installs the software, in a log file. That is a different audience with a different language.
- **Uncommon sentences wait their turn.** The catalogue covers every message appearing at two or more places in the code. A one-off sentence gets added when its code is converted — inventing wording ahead of that is guesswork that the conversion would just revise.

## Why this can't be checked by eye

There is a failure mode here that is invisible to everyone who would normally catch it.

If a message has no translation, the app quietly falls back to the English one. That fallback is deliberate and worth keeping — it is what stops a half-finished translation from breaking the app. But it means a missing translation **looks exactly like everything working**. The developer sees a correct error message. The reviewer sees a correct error message. Both are reading English. The only person who ever sees the bug is the non-English user, in production, and nothing tells them anything is wrong — the message just arrives in the wrong language, forever.

So this adds automated checks that run on every pull request. They enforce that every situation the server can report has English text to translate from, that no translation has been left behind by a rename, that a call site cannot leave a blank unfilled, and that every blank a translated sentence expects to have filled in — a name, a status — is one the server actually provides, **in every language rather than just English**.

One distinction is worth being precise about, because it decides what actually blocks a merge. A language missing a *whole message* only produces a warning: the reader gets that one sentence in English, which is complete and understandable, and holding up unrelated work over a translation that is a few sentences behind would just get the check switched off. A language missing one of the *values* dropped into a sentence does fail the build, because the sentence around it still translates — so the reader gets a single English word wedged into their own language, which is worse than either language on its own.

That last point is the one worth dwelling on. English is also the fallback, so it renders correctly no matter what its translation says. Checking only English checks the one language that cannot fail.

## Confidence

Rather than trusting the checks to work, each was deliberately broken and confirmed to fail the build — a missing translation, a stale one, a translation expecting information that isn't sent, wording drifting apart between the two places it lives — then restored. Fifteen in total, including one deliberate check that a *correct* translation is **not** rejected.

The change was also put through a full review pass, which found eight issues. All eight were real and seven are fixed. Two were the same class of bug this work exists to prevent, reaching users through a gap in the checks themselves:

- One message would have spliced a user's own mistyped input into a translated sentence as though it were a translatable term — producing a sentence half in each language.
- Another sent status values the translation files had no entry for, so an Indonesian sentence would have ended in an English word. Worse, **the automated check could not have caught it**: it inspects values written directly in the code, and these are computed at runtime. That blind spot is now covered by a second check, and written into the documentation rather than left to be rediscovered.

An automated review pass then found four more, all in the checks themselves rather than the feature — the same class of problem again, and the reason it is worth running more than one pass over work like this. Each was reproduced before being fixed:

- The check was blind to the very shortcuts the documentation tells people to use, so the most likely way to introduce the bug was the one way it wasn't looking.
- Nothing verified that a message was actually given the information it asks for, so a message could ship with a visible blank in it.
- Values were still being checked in English only in three places — the one language that cannot fail, because it is also the fallback.
- The documentation said there were four kinds of information a message can carry, after a fifth was added.

The eighth is a small improvement to how the command-line tool prints errors. It is real, but it changes what operators see on their terminal, so it belongs with the conversion work rather than ahead of it. It is recorded so it isn't lost.

## Cost and risk

Low. No database change, no API break, no visible change for users, and no new dependency. The checks run in the existing fast test job and add about a second.

## Sequencing

Independent of #243 and #244, both still open. This branch shares no files with either and touches no interface templates, so nothing here invalidates their baselines and the three can merge in any order.

The next steps after this — teaching the browser to render the labels, then converting the call sites — are what make it visible to users. The second of those overlaps with #244 and should follow it.


unidoc-alip:
All three Copilot findings and the three suppressed comments were reproduced before being fixed. Every one was valid; all are addressed in 4e7a8ba.

**The scan was blind to the wrappers it recommends.** This was the sharpest of the six. Inside `errNotFound` the constructor call is `Entity(entity)` — a variable — so a scan that only knew `Entity`/`Field`/`Status` let `errNotFound("widget")` straight through. That is the spelling `docs/i18n.md` tells people to use for the common cases, so the check only ever saw the form nobody writes. Confirmed: adding `errNotFound("widget")` passed the whole suite. The scan now recognises the three wrappers as their param kinds.

**Nothing tied a call site to its code.** `Entity` and `Field` are both `ErrorParam`, so `apiError(404, CodeNotFound)` and `apiError(404, CodeNotFound, Field("title"))` both compiled and both rendered a literal `{entity} not found` — verified by running them. Two layers now: `TestCallSitesMatchTheirCodes` compares each call's param set against its code's template at build time, and `apiError` logs and degrades if a slot somehow survives to runtime. The runtime half follows the trade this codebase already settled for the same question in `internal/isms/db/notifications.go` — a degraded message beats a lost response.

**Param values were checked in English only.** Correct, and it is the one language that renders correctly whatever it says, because it is also the fallback. Deleting `enum_inline.status.changes_requested` and `field.title` from `id-ID` passed both the Go and the web suites. Values are now checked in every locale. Since the scan can only see values a call site passes, `common.field.*` is additionally mirrored wholesale, so a key added to `en` and forgotten elsewhere fails before conversion reaches it.

The fifth parameter name in the docs table is fixed.

**On the suppressed comment about missing error translations only warning:** the policy stands, but the asymmetry is now stated rather than implied. A locale missing a whole message falls back to a complete, coherent English sentence — warning is right, and matches `localeKeyset.test.js`. A locale missing a *param value* is a different failure: the sentence around it still translates, so the reader gets one English word wedged into their own language, which is worse than either language and beyond what the fallback can rescue. So messages warn and param values fail. The PR description overstated this and has been corrected.

Each of the four cases was reproduced, then confirmed to fail the intended test, then restored.

unidoc-ahall (approved):
Independently verified, including mutation-testing three of the four second-round fixes against the actual shipped code rather than trusting the description.

`go build ./...`, `go vet ./...`, `go test ./...` all clean. Confirmed no real call site uses `apiError`/`errNotFound`/`errInvalidEntityID`/`errRequired` yet — the two AST-based call-site checks are genuinely vacuous today, exactly as the PR states, not silently broken.

Reproduced three of the described failures myself:

1. **The wrapper-blindness bug.** Added a throwaway `errNotFound("widget")` call site (a value with no catalogue key) in a scratch non-test file. `TestParamValuesHaveCatalogueKeys` caught it in both `en` and `id-ID` — confirming the scan really does see through the convenience wrappers now, not just the raw `Entity(...)` form nobody writes.
2. **Call-site/code mismatch.** Called `apiError(http.StatusNotFound, CodeNotFound)` with no `Entity(...)` param. `TestCallSitesMatchTheirCodes` failed with exactly the right message — `{entity} not found` needs `[entity]`, got `[]`.
3. **English-only param check.** Deleted `field.title` from `id-ID`'s `common.json`. `TestParamVocabularyExistsInEveryLocale` caught it — the vocabulary-mirror check works independently of any call site existing yet.

All three reverted, `git status` clean throughout.

Spot-checked the wire-shape reasoning against the actual Echo source behavior via `TestErrorBodyWireShape` (passed as-is) and confirmed `errorBody` implements neither `error` nor `json.Marshaler` (`TestErrorBodyIsNotSpecialCasedByEcho`), which is what the flat-response claim depends on.

The `value` vs `entity`/`field`/`status` distinction (raw caller input vs. catalogue-looked-up identifier) is the right shape — conflating them would either splice an unbounded string through a translation lookup or block translating known-set values, and the doc explains this clearly enough that I didn't need to dig further to be convinced.

Approving.

PR #246, opened 2026-09-05
fix(test): unblock the go-unit CI job on master

unidoc-alip:
## What is broken

The `go-unit` check on `master` is currently red, so every new pull request inherits a failing build. Nothing is wrong in production — the application itself compiles and runs, and the integration check is still green. Only the automated test suite fails to start.

## Why it happened

Two recent pieces of translation work landed close together:

- one added a guard that keeps notification messages in step with the translation files
- one added a guard that keeps API error messages in step with the translation files

Both needed to point at the same folder of translation files, and each one independently created a setting with the same name for that path. Each change passed its own checks before merging, because neither branch could see the other's new name. They only collide once both are on `master`.

This is a normal side effect of two parallel changes touching the same area, not a defect in either one. Both guards are still wanted and both stay.

## What this changes

The two guards now share a single definition of where the translation files live, instead of each declaring their own. The safety behaviour is unchanged: if the translation files are ever missing or moved, the tests still fail loudly rather than quietly passing.

Test-only change. No effect on the product, the database, or anything a customer sees.

## Verification

The full test suite for the affected area was run locally and passes; formatting and static checks are clean. Merging this returns `master` to green.

## Risk

Low. One test-support file touched, no product code.

### DIFF
--- a/internal/isms/api/errors_test.go
+++ b/internal/isms/api/errors_test.go
@@ -30,13 +30,15 @@ import (
 
 // ---- locale catalogue loading -------------------------------------------
 
-func localesDir(t *testing.T) string {
+// requireLocalesDir returns the shared locale root declared in
+// notification_keys_test.go, failing if it is missing. A missing catalogue is
+// the failure these tests exist to report, so it must never skip.
+func requireLocalesDir(t *testing.T) string {
 	t.Helper()
-	dir := filepath.Join("..", "..", "..", "web", "src", "locales")
-	if _, err := os.Stat(dir); err != nil {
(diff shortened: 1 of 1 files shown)


unidoc-ahall (approved):
Verified the actual breakage, not just the description. Swapped in the pre-fix version of errors_test.go against the current api package (with #244's notification_keys_test.go already present): `go vet` reproduced exactly the stated error, `localesDir redeclared in this block` at notification_keys_test.go:36. Restored, and `go build ./...` / `go vet ./...` / `go test ./internal/isms/api/...` are all clean with this fix applied.

Minimal, correctly scoped, test-only. Approving.

PR #254, opened 2026-09-07
feat(i18n): make API error messages translatable across all modules (#212)

unidoc-alip:
## What this changes

Every error the API returns used to be an English sentence and nothing else, so the web app had no way to show it in the reader's language. This branch attaches a stable machine-readable code, plus a small set of named values, to 385 of them — across every module the shared vocabulary covers.

Nothing a user sees changes yet, and that is deliberate. This is the groundwork; connecting the screens to it is separate work, sequenced after this.

## Why one pull request

There were 24 API modules and roughly 900 error sites to work through. Sent individually that would have been twenty-odd reviews of the same mechanical substitution. Each module still got its own pull request, its own automated checks and its own commit against this branch — the per-module record and history survive — but the decision to ship reaches master once, here.

## What landed

| Module | Sites |
|---|---|
| Reviews and tasks (#249) | 71 |
| Sign-in and accounts (#250) | 62 |
| Audits and findings (#251) | 36 |
| Documents and registers (#255) | 36 |
| Objectives and programs (#253) | 32 |
| Admin and single sign-on (#256) | 21 |
| Passkeys (#257) | 20 |
| Auth middleware (#259) | 16 |
| Incidents (#258) | 12 |
| Corrective actions (#260) | 11 |
| Approval policies (#261) | 10 |
| The remaining ten modules, batched (#262) | 44 |
| Vocabulary fix for "check-in" (#263) | — |

Plus the earlier groundwork already on master: the code catalogue itself and the client-side renderer (#245, #248).

## What a user could notice

Almost nothing. The large majority of these 385 messages read exactly as they did before, word for word. A handful now read better, and each is the shared vocabulary doing its job rather than an incidental edit:

- "invalid provider ID" says "invalid OIDC provider id" — the old text was truncated and did not name which kind of provider.
- "invalid legal id" says "invalid legal requirement id", for the same reason.
- "user is not a member of this organization" says "not a member of this organization", matching the eight other places that already said it that way.
- "path_pattern is required" says "path pattern is required" — the old text quoted an internal field name.
- "checkin" is spelled "check-in", consistently on both the server and the screen.

Every message keeps the HTTP status it had. That was verified per module by arithmetic rather than by eye, since a silent status change in an authentication path is the one mistake this work could plausibly have introduced.

## What was deliberately left alone

Roughly 520 sites remain untouched, and they are not a backlog:

- **Internal faults** — "creating user", "reading document" — which describe a server problem and are read in a log.
- **Operator configuration messages** — the WebAuthn, SMTP, single sign-on and Cloudflare Access setup errors, read by whoever deploys the server.
- **The git endpoints entirely**, which are consumed by `git push` and never rendered in a browser.
- **One-off sentences** whose wording is not settled. A message that appears once gets its phrasing decided on its own rather than guessed at during a mechanical pass.

## Risk

Low, and it did not grow as modules were added. The old and new styles coexist by design: an unconverted error is indistinguishable to any client from a converted one, the added fields are additive, and no schema or API contract changes.

The one genuine defect found during the work was caught by review on #253 — the server and the shared vocabulary disagreed on how to spell "check-in" — and #263 fixes it and adds the check that would have caught it. That check was missing because the existing one verified the declared exceptions rather than the rule itself.

## Merging

**Merge with a merge commit or a rebase, not a squash.** Each module is a separate commit so that any one of them can be reverted on its own, and so history still shows which change touched which file. A squash would collapse 385 changes across 23 files into a single commit.

## Verification

gofmt clean, `go vet ./...`, and the full `go test ./...` suite green on the branch. Master is an ancestor of this branch, so there is no drift to resolve.

### DIFF
--- a/internal/isms/api/api_admin.go
+++ b/internal/isms/api/api_admin.go
@@ -54,22 +54,22 @@ func (s *Server) handleAdminUpdateRole(c echo.Context) error {
 
 	userID, err := strconv.Atoi(c.Param("userId"))
 	if err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, "invalid user ID")
+		return errInvalidEntityID("user")
 	}
 
 	var req struct {
 		Role string `json:"role"`
 	}
 	if err := c.Bind(&req); err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, "invalid request")
+		return apiError(http.StatusBadRequest, CodeInvalidRequest)
 	}
 	if req.Role == "" {
-		return echo.NewHTTPError(http.StatusBadRequest, "role is required")
+		return errRequired("role")
 	}
 
 	validRoles := map[string]bool{"admin": true, "manager": true, "contributor": true, "reader": true}
 	if !validRoles[req.Role] {
-		return echo.NewHTTPError(http.StatusBadRequest, "invalid role")
+		return apiError(http.StatusBadRequest, CodeInvalidRole)
 	}
 
 	// Prevent self-downgrade if last admin
@@ -91,7 +91,7 @@ func (s *Server) handleAdminUpdateRole(c echo.Context) error {
 
 	// Verify user is already a member before updating role
 	if _, err := s.db.GetOrgMember(ctx, orgID, userID); err != nil {
-		return echo.NewHTTPError(http.StatusNotFound, "user is not a member of this organization")
+		return apiError(http.StatusNotFound, CodeNotOrgMember)
 	}
 	if err := s.db.AddOrgMember(ctx, orgID, userID, req.Role); err != nil {
 		return echo.NewHTTPError(http.StatusInternalServerError, "updating role: "+err.Error())
@@ -114,7 +114,7 @@ func (s *Server) handleAdminRemoveMember(c echo.Context) error {
 
 	userID, err := strconv.Atoi(c.Param("userId"))
 	if err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, "invalid user ID")
+		return errInvalidEntityID("user")
 	}
 
 	// Prevent removing yourself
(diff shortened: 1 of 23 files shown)


unidoc-ahall (approved):
Independently verified rather than trusting the description's "verified per module by arithmetic" claim on its word — reproduced that verification programmatically across the whole diff, since eyeballing ~900 call sites isn't a real review.

**HTTP status preservation, checked mechanically, not by eye.** Parsed the full `master...HEAD` diff for `internal/isms/api/` and paired every removed `echo.NewHTTPError(http.StatusX, ...)` against its replacement within the same hunk — either a direct `apiError(http.StatusX, ...)` call or one of the three status-fixed wrappers (`errNotFound`→404, `errInvalidEntityID`→400, `errRequired`→400). 371 status-bearing conversions, paired 1:1, zero mismatches, zero leftovers on either side. This is exactly the "silent status change in an auth path" risk the description calls out, checked directly rather than assumed.

**Build/test.** `go build ./...`, `go vet ./...`, `go test ./...` all clean on `feat/i18n-error-codes` as-is.

**Confirmed the specific claims rather than skimming them:**
- `api_git.go` has zero diff — the git wire protocol really is untouched, as claimed.
- `master` is an ancestor of `HEAD` (`git merge-base --is-ancestor` confirms) — no drift to resolve, matching the merge note.
- The four described wording changes are exactly what's in the diff: both "invalid legal id" (readings endpoints) and "invalid legal requirement id" (CRUD endpoints) now unify to `errInvalidEntityID("legal_requirement")` → "invalid legal requirement id"; "user is not a member of this organization" → `CodeNotOrgMember` → "not a member of this organization"; the two `path_pattern` policy sites now go through `Field("path_pattern")`; `checkin` gets a `paramMessageOverrides` entry. Grepped `tests/` and `web/` for all the old literal wordings — nothing depends on any of them.

**Mutation-tested the one genuine defect this branch fixes (#263).** Removed the new `"checkin": "check-in"` override and reran `TestCatalogueValuesMatchGoProse` — failed with exactly the right diagnostic (catalogue says "check-in", Go renders "checkin"). This is also a good structural addition on its own: it checks the *rule* (every catalogue value matches what Go would actually render) rather than only the previously-existing test that checks the *declared exceptions* — which is precisely why the checkin drift went uncaught the first time. Restored, `git status` clean.

Nothing further to add. Large, mechanical, and the mechanical part is what I checked mechanically rather than assumed. Approving.

Noting the merge instruction for whoever lands this: merge commit or rebase, not squash, so the 24 per-module commits stay revertable individually.

PR #265, opened 2026-09-07
docs: replace local plan references with in-repo ones (#212)

unidoc-alip:
## What this changes

Comments and test headers across the i18n work referred to a planning scheme that only exists on my machine — things like "plan 81 §1", "task 1.5", "plan 82 step 7". There were 21 of these across 15 files, mostly in the header comment at the top of a file or in a test's description, which is exactly where someone new to the code starts reading.

For anyone other than me, each one is a dead end at the moment they most want the background.

## Why it's nearly all deletion

The interesting part: most of these needed no replacement. The sentence containing the reference already said the thing — the citation was decoration.

```diff
-// The governing rule (plan 81 §1): the server never owns a translation
+// The governing rule: the server never owns a translation
```

Where the pointer really was carrying information, it now names something a reader can actually open: a file in the repository, or the tracking issue for work still in progress.

## Two stale statements went out with it

Not the goal, but worth knowing, since both would have misled a reader on their own:

- The raw-text check described the remaining extraction as "~21 PRs" — that was the estimate from before the work was regrouped into six.
- The same file referred to the "outstanding half" of a check that has since been finished.

## Risk

As low as it goes. Comments, test descriptions, and one paragraph of prose in `docs/i18n.md`. No executable line changes, so no behaviour can shift. The whole diff is 36 insertions and 32 deletions.

## Scope notes

Two files with the same problem, `internal/isms/i18n/locale.go` and its release-gate test, are fixed on a different branch — the one that already touches that package — so the two changes don't collide. Once both land the tree is clean.

Left alone on purpose: `docs/suggestions.md`'s "Phase 1/Phase 2" headings are that document's own section names and resolve fine, and the `.claude/settings.json` mentions in `README.md` and `docs/ai-first.md` are real user-facing configuration for a real tool.

## Verification

- `go build ./...`
- `go test ./internal/isms/api/ ./internal/isms/db/ ./internal/isms/i18n/`
- `npm test` — 148 pass
- `npm run build`
- `gofmt` clean; added lines wrapped to 80 columns
- A grep for the old patterns now returns only the two lines handled on the other branch

Relates to #212.

### DIFF
--- a/docs/i18n.md
+++ b/docs/i18n.md
@@ -181,8 +181,9 @@ While extracting a view, `npm run i18n:raw -- --list Risks.vue` lists what the
 scanner still sees in it, with line numbers. The scanner is regex over markup
 and deliberately over-reports, so the numbers are budgets rather than defect
 counts; for a false positive that has to reach zero, mark the line with an
-`i18n-ignore` comment. Plan 80 §6 has the reasoning, including why
-`eslint-plugin-vue-i18n` is the right answer only once extraction is done.
+`i18n-ignore` comment. The scanner's own header comment has the reasoning,
+including why `eslint-plugin-vue-i18n` is the right answer only once extraction
+is done.
 
 ## Out of scope
 
--- a/internal/isms/api/errors.go
+++ b/internal/isms/api/errors.go
@@ -14,7 +14,7 @@ import (
 
 // Stable error codes for HTTP responses, and the helper that emits them.
 //
-// The governing rule (plan 81 §1): the server never owns a translation
+// The governing rule: the server never owns a translation
 // catalogue for anything a client can render. An error response is read by a
 // browser that has the locale files loaded, so the server emits a stable
 // `code` plus a closed set of `params` and the browser renders the sentence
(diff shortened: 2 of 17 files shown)


unidoc-ahall (approved):
Verified rather than trusted.

`go build ./...`, `go vet ./...`, `gofmt -l` (clean) on all six touched Go files, `go test ./internal/isms/api/... ./internal/isms/db/... ./internal/isms/i18n/...` all pass. `web/`: 148/148, `npm run build` clean.

Checked the "grep now returns only the two lines on the other branch" claim directly rather than taking it on faith: a tree-wide grep for `plan 8[0-9]` returns nothing at all in this branch's changed files, and the only two remaining `plan` references anywhere are exactly where the description says — `internal/isms/i18n/locale.go:64` (a `.claude/plans/79-...` path) and `internal/isms/i18n/release_gate_test.go:83` (`plan 79 Phase 3`), both in the package the other branch owns. Nothing missed, nothing over-claimed.

No behavior to regress — pure comments and test-header prose, confirmed by the diff itself (no touched line falls outside a `//` comment or a string literal in a doc/test file). Approving.

PR #267, opened 2026-09-07
feat(i18n): extract the shared components (#212)

unidoc-alip:
## What this does

Continues making the web UI translatable. This batch covers the **26 shared components** — the pieces that appear inside other screens rather than being screens themselves: the risk/asset/supplier assessment panels, the document editor and its toolbars, the review comparison views, inline comments and suggestions, the search dialog, pagination, the reference badges, and a dozen smaller controls.

Nothing changes for an English reader except the three cosmetic items listed below.

Progress on the overall effort: **2295 untranslated strings before, 2003 after.** All 292 in these components are done, along with all 12 places where a server error message was being displayed without going through the translation layer. Every shared component is now at zero on both automated counters; what remains is entirely full screens.

## ⚠️ Please merge with a merge commit, not a squash

Each component is its own commit (23 in total). Squashing collapses them into one, which loses the ability to revert a single component if something turns out wrong, and makes `git blame` point at the batch instead of the component. Same request as #254 and #266.

## Two things this found that were already broken

**Type badges disagreed with each other.** Three panels that show what kind of thing a link points to each carried their own private list of abbreviations, and the lists had drifted apart: the same supplier reference showed as `SUP` in two panels and `SUPPLIER` in a third. Nothing catches that — they are separate pieces of text in separate files. They now share one list.

**A stored value was being shown to the reader as-is.** The supplier review history printed the outcome exactly as the database stores it, so it read `concerns` in lower case rather than "Concerns identified". Because it was a value rather than a piece of text, it was invisible to the tool that counts untranslated strings — reaching zero would never have found it.

Both are the kind of defect that only surfaces when someone reads the screen, which is why the browser pass below matters more than the test suite.

## Visible changes for an English reader

Three, all intentional:

- A suggestion labelled **New** now reads **Create**, matching what every other suggestion screen already showed. It was using a private copy of the wording instead of the shared one.
- Four link badges in one panel shorten to match the other two panels: `SUPPLIER` → `SUP`, `SYSTEM` → `SYS`, `INCIDENT` → `INC`, `CHANGE` → `CR`.
- The change history read **"Changed4 fields"** with the space missing, and now reads **"Changed 4 fields"**. This one is a pre-existing bug, not something this work introduced — I checked it against `master` before touching it. It is fixed here because the line is now a translated sentence, and a translator has no way to add a space that the page strips out.

## A finding worth its own issue

**Three of these 26 components are not reachable in the running application.** Two have no reference anywhere in the code, and a third is pulled into a screen but never actually placed on it. They still had to be converted, because they are counted by the automated checker, but they cannot be opened in a browser and cannot be regression-tested. That is roughly a tenth of this batch's work spent on code that nothing displays. Worth deleting or reconnecting, but that is a separate decision from this change.

## Verification

Every commit builds and passes the test suite on its own, not just the branch as a whole. That distinction earned its place here: partway through, 148 passing tests were sitting on top of a version that would not build at all, because the test runner never compiles page templates. Checking the build per commit is now part of the routine.

For the browser pass I set up a database with the data these components need — a document review with inline comments and a pending suggestion, a risk with an assessment history, a supplier with a completed review, 61 assets to force pagination, and three overdue items — and opened all 23 reachable components in both English and a temporary second language. The three unreachable ones were rendered directly instead.

Running in a second language is what makes the check conclusive: in English a label looks identical whether it is wired up correctly or frozen at page load. Switching languages is the only thing that tells them apart, and it caught several labels that would have stayed English on an otherwise translated screen.

## Risk

Low, and the worst realistic outcome is cosmetic: a label appears in English when it should have been translated, or a tooltip goes missing. No data, permissions or workflow behaviour is touched. The largest single change is the shared insert-menu used by both editors, which moved from fixed English text to looked-up text — its menu, its descriptions and its type-to-filter behaviour were all exercised in the browser.


unidoc-alip:
Good catch, and it turned out to be more than one instance.

Fixed in 976c868 — the toast now uses `components.suggestions.created_toast`, the key you pointed at. That sentence already existed because the suggestion panel shows the same confirmation, so this removes a duplicate rather than adding one.

Since the string sat in `<script setup>`, which the raw-text checker never reads, I swept every component in this batch for the same shape. It found eleven more, and one of them was worse than the original:

**CommentSidebar's composer placeholder was silently untranslated.** The template reads `placeholder || $t(...)`, but the prop's default was still the English sentence — always truthy, so the catalogue fallback was dead code and the hardcoded default won every time. The extraction made that line *look* converted while leaving it hardcoded. The default is now empty.

**HeatMap's two axis-label maps** (`Rare`…`Almost Certain`, `Negligible`…`Severe`) were module-level objects — the same "built before the language is known" problem this PR fixes elsewhere. I had wrongly assumed they came from the parent when I checked the rendered page. Both are now computed, and I confirmed all ten switch language in a browser.

Also fixed: the two editor placeholders now hand the editor a *function* rather than a string — a string is captured once when the editor is created and would freeze in whatever language was active at boot — plus prop defaults in MemberPicker and MarkdownField.

Two deliberately left alone, both now carrying a comment saying why:
- the `console.error(...)` prefixes, which are developer logs and never reach a reader;
- DocumentViewer's placeholder suggestion body, which is persisted and read back by other reviewers, so translating it would store the author's language in shared data.

All re-verified in a browser in both languages, including the toast itself.

unidoc-ahall (approved):
Independently verified against the current code, not just the diff/description — this is a 33-file batch, so I prioritized programmatic baseline verification plus targeted checks against every specific claim in your own follow-up comment, over reading all ~3700 diff lines of mechanical `$t()` substitution by eye.

`go build/vet/test ./...` clean (no Go files touched, sanity-checked anyway). `web/`: 148/148, `npm run build` clean.

**Baselines verified programmatically**, not spot-checked: ran the live `i18nRawText.mjs`/`i18nErrorRender.mjs` scanners and diffed every count against the committed baselines — exact match, no drift. Every `components/*` entry is genuinely absent from both (zero, not low), confirming "all 292 done" and "all 12 error-sites done" rather than trusting the count.

**Checked every fix named in your own follow-up comment against the shipped code:**
- `CommentSidebar`'s prop default is `''`, so `placeholder || $t(...)` genuinely falls through now — with the old truthy default this could never have translated, confirmed by JS truthiness alone.
- `HeatMap`'s `likelihoodLabels`/`impactLabels` are `computed()`, not module-level objects — will re-evaluate on locale change.
- Both editor placeholders (`DocumentEditor.vue`, `MarkdownField.vue`) pass a function to `Placeholder.configure`, not a string — won't freeze at boot.
- `MemberPicker`'s `placeholder` prop default is also `''`, same fix applied consistently.
- The three reference components (`EntityReferences`, `ReferenceManager`, `ReferenceCollector`) share one `typeAbbr()` function; `common.json`'s `entity_abbr` group has `SUP`/`SYS`/`INC`/`CR` — the described drift is genuinely unified, not just moved.
- `SupplierReviewPanel`'s stored `outcome` renders through `outcomeLabel()`/the catalogue, not the raw value.
- `HistoryPanel`'s missing space is fixed with an explicit `&nbsp;` between the two adjacent inline spans — the actual HTML-whitespace-collapsing cause of "Changed4 fields", not a coincidental rewrite.
- `SuggestNewButton`'s toast now calls `t('components.suggestions.created_toast')`, the same key `SuggestionPanel` uses — a real dedup, not a new string.
- `console.error(...)` calls remain untranslated across the batch, as intended (dev logs, never reader-facing).

Nothing further to add. The "3 unreachable components still had to be converted" finding is worth its own cleanup issue as you noted, but that's explicitly out of scope here.

Approving.

Issue #269, opened 2026-09-08
Six of the eight legal-requirement categories the UI offers are rejected on save


unidoc-alip:
`Legal.vue` offers a category list that does not match the one the server accepts. Six of its eight options cannot be stored.

**The two lists**

`web/src/views/Legal.vue:580` offers:
`data_protection`, `employment`, `contractual`, `regulatory`, `intellectual_property`, `financial`, `environmental`, `corporate`

`internal/isms/db/legal.go:14` accepts:
`privacy`, `security`, `sector`, `contractual`, `other`

Only `contractual` is in both. `other` is accepted but never offered.

**Two distinct defects**

1. **Update returns 400.** `internal/isms/api/api_legal.go:208` runs `validateEnum("category", …, db.LegalCategories)`, so picking any of the six unsupported categories on an existing requirement fails validation.

2. **Create fails at the database instead.** The create path at `internal/isms/api/api_legal.go:111` assigns `Category` straight from the request with no `validateEnum` call, so the value reaches Postgres and trips the CHECK constraint at `migrations/20260327000000_initial_schema.sql:639`. A validation error surfaces as a database error rather than a 400.

**Suggested fix**

Decide which list is authoritative — the view's taxonomy is the more descriptive of the two, so widening `db.LegalCategories` and the CHECK constraint may be the better direction than narrowing the view. Whichever way it goes, the create path should call `validateEnum` like the update path does, so an invalid category is a 400 and not a 500.

Found while extracting `Legal.vue` for translation (#212); the extraction PR deliberately leaves the view's list unchanged, since changing which options a user sees is not an extraction's call.

PR #270, opened 2026-09-08
feat(i18n): extract the six register views (#212)

unidoc-alip:
Makes the six register screens translatable: **Risks, Legal, Systems, Suppliers, Incidents and Assets**. Continues #212, following the auth views (#264), the shell and landing page (#266), the shared components (#267) and the admin surface (#268).

**Please merge with a merge commit, not a squash.** There are ten commits and each one is a single view or a single decision; the per-view record is what makes this reviewable and what the next contributor will read. #268 carried the same request and was squashed, so its eight commits are gone.

## What changes for a user

Nothing, today. Every one of these screens renders exactly the same English it did before, because English is the only language the app currently offers. What changes is that the text is now *capable* of being another language — it comes out of translation files instead of being baked into the code.

Three visible differences, all deliberate:

- **Some labels change capitalisation.** "Under Review" becomes "Under review", "Data Protection" becomes "Data protection", "Not Required" becomes "Not required". These labels now come from the app's shared vocabulary, which has always used sentence case — and which the status badges on these same screens were already using. The alternative was two spellings of the same word on one screen. Words are unchanged; only the capital letters are.
- **Two labels are now correct that were wrong.** An asset typed as intellectual property displayed as "Ipr", and one typed as sales & marketing displayed as "Sales marketing". Both were the side effect of the code guessing a label from a database value. They now read "Intellectual property" and "Sales & marketing".
- **A panel that never appeared now appears.** See the bug fix below.

## What this fixes

**The supplier detail page's "Linked systems" panel could never render.** If you had systems pointing at a supplier, opening that supplier showed nothing where the list should be. Cause was a one-line mismatch in how the response was read. Found while checking that a newly added label could actually be seen — it could not, which is what made the bug findable.

## What this deliberately does not fix

**Legal requirement categories are broken and stay broken here — filed as #269.** The category dropdown offers eight options; the server accepts five, and only one of those is in both lists. Six of the eight fail on save. Changing which options a user sees is a product decision, not something a translation pass should slip in, so the list is untouched and the bug is tracked separately.

**Three kinds of text on these screens are still English regardless of language**, because the text does not come from the app:

- Risk categories, which each organisation configures for itself
- Legal jurisdictions, which are free-text country names
- Anything an administrator has typed into a setting

Making those translatable needs the server to send codes instead of sentences, which is separate work already scoped under this issue.

## Risk

**Low, and mechanically checked.** Two automated gates run in CI and both are at zero for all six files: one counts untranslated text, the other counts error messages that bypass the translation layer. Between them they went from 617 and 45 findings to none.

The gates do not see everything, so each of the six screens was also opened in a browser and driven through a stand-in second language, which makes any text that failed to translate immediately obvious. That pass is what caught the last two problems in this branch — a set of letters no automated check can see, and the broken panel above.

**Around 200 lines of unreachable code were deleted** rather than translated: two duplicate edit forms on the risk screen that no navigation could reach, and a status-change helper that five of the six screens each carried a copy of, none of them called. Translating dead screens would have meant shipping vocabulary for buttons nobody can press.

## Scope

- 6 screens, 617 pieces of text, 45 error messages
- 435 new translation entries: 344 belonging to individual screens, 91 shared
- The shared vocabulary lands first, in its own two commits, because these six screens are near-copies of each other and would otherwise have produced six copies of every shared label. The workflow and document screens still to come will draw on the same vocabulary — the tab bar alone is nine labels they all share.

## Verification

- `npm test` — 150 pass
- `npm run build`
- `go test ./internal/isms/i18n/...`
- Both text gates at zero for all six files; the totals they track fall from 1692 to 1075 and from 145 to 100
- All six screens opened in a browser against a stand-in locale, with no missing-translation warnings

Closes the register portion of #212. The workflow screens (audit, tasks, objectives, corrective actions, changes, programs) and the document screens are next.

### DIFF
--- a/internal/isms/i18n/release_gate_test.go
+++ b/internal/isms/i18n/release_gate_test.go
@@ -18,11 +18,12 @@ const rawTextBaseline = "../../../web/test/rawText.baseline.json"
 // punctuation, and attribute values it cannot distinguish from copy — so a
 // fully extracted tree still carries a residue, and 100 is an estimate of it.
 //
-// For scale: that is about 25x below where the tree sits today (2415) and about
-// 25x below where it started (2488). Those two numbers are nearly identical,
-// which is the honest summary of how far extraction has got. The estimate wants
-// revisiting once a mostly-extracted tree shows what the residue actually is —
-// the threshold is a judgement, and says so. The coupling it enforces is not.
+// For scale: the tree started at 2488 and sits at 1075 now that the auth,
+// shell, component, admin and register surfaces are extracted — so roughly 10x
+// above the threshold, with the workflow and document views left. The estimate
+// still wants revisiting once a fully extracted tree shows what the residue
+// actually is; the threshold is a judgement, and says so. The coupling it
+// enforces is not.
 const extractionGateThreshold = 100
 
 // TestSecondLocaleRequiresExtraction is a release gate, not a unit test.
--- a/web/src/locales/en/assets.json
+++ b/web/src/locales/en/assets.json
@@ -1,0 +1,76 @@
+{
+  "title": "Asset Register",
+  "subtitle": "Information assets, classification, and ownership",
+  "error": {
+    "save": "Failed to save: {message}",
+    "create": "Failed to create asset: {message}",
+    "delete": "Failed to delete: {message}"
+  },
+  "action": {
+    "add": "Add Asset"
+  },
+  "create": {
+    "heading": "Add Asset",
+    "name_label": "Name *",
+    "name_placeholder": "e.g. Production Database, Customer Portal",
+    "type_label": "Type",
+    "deselect_hint": "Click to deselect",
+    "fill_in_later": "You can add CIA classification, owner, and more after creating.",
+    "submit": "Add"
+  },
+  "filter": {
+    "empty": "No assets found"
+  },
+  "stat": {
+    "severe": "Severe (CIA = 5)"
+  },
+  "table": {
+    "header": {
+      "asset": "Asset",
+      "type": "Type",
+      "cia": "C/I/A",
+      "status": "Status",
+      "owner": "Owner"
+    },
+    "cia_title": {
+      "confidentiality": "Confidentiality: {level}",
+      "integrity": "Integrity: {level}",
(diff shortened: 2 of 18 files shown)


unidoc-ahall (approved):
Independently verified rather than trusting the numbers.

`go build/vet/test ./...` clean. `web/`: 150/150, `npm run build` clean.

**Both gate totals verified programmatically**: live-scanned with `i18nRawText.mjs`/`i18nErrorRender.mjs`, diffed against the committed baselines — exact match, and the totals are exactly 1075 and 100, matching the description precisely. All six views (`Risks`, `Legal`, `Systems`, `Suppliers`, `Incidents`, `Assets`) are genuinely at zero on both counters, not just low.

**The linked-systems panel fix, traced to its actual cause**: `fetchJSON()` in `api.js` already unwraps a `{data: [...]}` body into a bare array (`if (Array.isArray(json.data)) return json.data`), so `api.listSystemsLinked()` returns an already-unwrapped array. The old code's `res.data` would therefore always have been reading `.data` off an array — `undefined`, falling to `[]` — which matches "the panel never rendered" exactly. The fix's `Array.isArray(res) ? res : (res?.data || [])` is the correct defensive form.

**Both cosmetic label corrections checked against the actual catalogue and wiring**, not just the description: `common.json`'s `asset_type` group has `"ipr": "Intellectual property"` and `"sales_marketing": "Sales & marketing"`, and `Assets.vue`'s `typeLabel` is `(v) => enumLabel('asset_type', v)` — genuinely wired through the catalogue rather than a raw de-slug.

**Dead-code deletion corroborated**: `Risks.vue` alone is -160 net lines in this branch (339 deletions, 179 insertions), consistent with "duplicate edit forms" being removed rather than translated.

**The deferred legal-category bug is genuinely untouched** — `Legal.vue`'s `categoryLabel` is a label lookup only; the option list itself isn't part of this diff, matching "filed as #269, not fixed here."

Threshold math in `release_gate_test.go` checks out: 1075 / 100 = 10.75x, "roughly 10x above the threshold" is accurate (this comment's predecessor needed a correction for exactly this kind of scale claim back in #264).

Nothing further to add. Approving.

PR #271, opened 2026-09-08
feat(i18n): extract the six workflow views (#212)

unidoc-alip:
Makes the six workflow screens translatable: **Audit, Change Management, Tasks, Objectives, Corrective Actions and Programs**. Continues #212, following the auth views (#264), the shell and landing page (#266), the shared components (#267), the admin surface (#268) and the six registers (#270).

**Please merge with a merge commit, not a squash.** There are nine commits and each one is a single screen or a single decision; the per-screen record is what makes this reviewable and what the next contributor will read. #268 and #270 both carried this request and were both squashed, so eighteen commits of that history are already gone.

## What changes for a user

Nothing, today. Every one of these screens renders the same English it did before, because English is the only language the app currently offers. What changes is that the text is now *capable* of being another language — it comes out of translation files instead of being baked into the code.

Three visible differences, all deliberate:

- **Some labels change capitalisation.** "Todo" becomes "To do", "Awaiting Approval" becomes "Awaiting approval", "Major Non-conformity" becomes "Major non-conformity". These labels now come from the app's shared vocabulary, which uses sentence case, and which the status badges on these same screens were already using. Previously the same value could appear two ways on one screen depending on which control showed it. Words are unchanged; only the capital letters are.
- **Eleven labels that displayed a raw database value now read properly.** A task's priority showed the literal text `critical`, its type showed `incident followup`, an audit's type showed `internal`. These were values printed straight to the screen, sometimes with a styling rule capitalising the first letter, which is why they looked plausible in English and would have stayed English in every other language.
- **A tab that never showed anything now shows it.** See the bug fix below.

## What this fixes

**The audit Findings tab was empty for everyone, always.** However many findings an organisation had recorded, the tab said "No findings recorded yet." and its four counters — Total, Open, Closed, Overdue — all read zero. The findings existed and the rest of the app could see them; only this screen could not. Cause was a one-line mismatch in how the response was read.

This is the second consecutive extraction to uncover a panel that had never worked (#270 found the supplier "Linked systems" block). Both were found the same way: seeding real data and looking at the rendered page, because an empty screen and a broken screen look identical when there is nothing to display.

## What this deliberately does not change

**A "Generated N tasks" banner was deleted rather than translated.** The function that would populate it is never called from anywhere, so the banner could never appear. Translating it would have meant shipping vocabulary for a message nobody can see.

**The audit calendar's month names now come from the browser rather than the server.** The server sends them in English; the browser can produce them in whichever language is active. The English output is identical.

## Risk

**Low, and mechanically checked.** Two automated gates run in CI and both are at zero for all six files: one counts untranslated text, the other counts error messages that bypass the translation layer. Between them they went from 653 and 51 findings to none.

The gates do not see everything, so each of the six screens was opened in a browser and driven through a stand-in second language, which makes any text that failed to translate immediately obvious. That pass is what found the broken Findings tab, and what confirmed the eleven raw database values above — none of which any automated check can see, because a value printed from the database was never written down as text in the first place.

**Roughly 150 lines of unreachable code were deleted** rather than translated: five status-change and save helpers that were defined and never called, an unused import, and the dead banner above. Six of them contained error messages, so they came off the count by deletion rather than by translation.

## Scope

- 6 screens, 653 pieces of text, 51 error messages
- 478 new translation entries: 405 belonging to individual screens, 73 shared
- The shared vocabulary lands first, in its own commit, for the same reason it did on the registers: these screens repeat each other, and nine new groups of shared values (audit types, task types, change categories, priorities and so on) would otherwise have been written out six times
- Audit is by far the largest screen in the app — three related things (programmes, audits, findings) in one file — and is deliberately its own commit

## Verification

- `npm test` — 152 pass, including two new ones covering the abbreviated-label lookup
- `npm run build`
- `go test ./internal/isms/i18n/...`
- Both text gates at zero for all six files; the totals they track fall from 1075 to 422 and from 100 to 49
- All six screens opened in a browser against a stand-in locale, over seeded data in every module, with no missing-translation warnings

With this merged, the only screens left in #212 are the document ones — Documents, Reviews and Inbox.

### DIFF
--- a/internal/isms/i18n/release_gate_test.go
+++ b/internal/isms/i18n/release_gate_test.go
@@ -18,9 +18,9 @@ const rawTextBaseline = "../../../web/test/rawText.baseline.json"
 // punctuation, and attribute values it cannot distinguish from copy — so a
 // fully extracted tree still carries a residue, and 100 is an estimate of it.
 //
-// For scale: the tree started at 2488 and sits at 1075 now that the auth,
-// shell, component, admin and register surfaces are extracted — so roughly 10x
-// above the threshold, with the workflow and document views left. The estimate
+// For scale: the tree started at 2488 and sits at 422 now that the auth,
+// shell, component, admin, register and workflow surfaces are extracted — so
+// roughly 4x above the threshold, with only the document views left. The estimate
 // still wants revisiting once a fully extracted tree shows what the residue
 // actually is; the threshold is a judgement, and says so. The coupling it
 // enforces is not.
--- a/web/src/api.js
+++ b/web/src/api.js
@@ -400,7 +400,10 @@ export const api = {
   updateAuditProgramme: (id, data) => putJSON(`${API}/audit/programmes/${id}`, data),
   deleteAuditProgramme: (id) => deleteJSON(`${API}/audit/programmes/${id}`),
   getAuditCalendar: (year) => fetchJSON(`${API}/audit/calendar?year=${year || new Date().getFullYear()}`),
-  getAuditFindingsPaginated: (params) => fetchJSON(`${API}/audit/findings?${new URLSearchParams(params || {})}`),
+  // fetchRaw, not fetchJSON: the caller reads `.data` and `.total` off the
+  // envelope, and fetchJSON unwraps `{data: [...]}` to a bare array — which
+  // left the audit Findings tab and its four counters permanently at zero.
+  getAuditFindingsPaginated: (params) => fetchRaw(`${API}/audit/findings?${new URLSearchParams(params || {})}`),
   listAuditFindings: (params) => fetchJSON(`${API}/audit/findings?${new URLSearchParams(params || {})}`),
   getAudits: (programmeId) => fetchJSON(`${API}/audits${programmeId ? '?programme_id=' + programmeId : ''}`),
   listAudits: (params) => fetchJSON(`${API}/audits?${new URLSearchParams(params || {})}`),
--- a/web/src/composables/useEnumLabel.js
+++ b/web/src/composables/useEnumLabel.js
@@ -47,6 +47,19 @@ export function enumLabelInline(group, value) {
   return i18n.global.te(key, FALLBACK) ? i18n.global.t(key) : enumLabel(group, value)
 }
 
+// Abbreviated form, for a control too narrow to hold the full label — an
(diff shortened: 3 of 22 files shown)


unidoc-ahall (approved):
Independently verified, including tracing the Findings-tab fix end to end through the actual call sites rather than trusting the commit message.

`go build/vet/test ./...` clean. `web/`: 152/152 (150 + 2 new `enumLabelAbbr` tests), `npm run build` clean.

**Both gate totals verified programmatically**: live-scanned with `i18nRawText.mjs`/`i18nErrorRender.mjs`, diffed against committed baselines — exact match, totals exactly 422 and 49 as claimed. All six views (`Audit`, `Changes`, `Tasks`, `Objectives`, `CorrectiveActions`, `Programs`) genuinely at zero on both counters.

**The Findings-tab fix, traced through both ends**: `Audit.vue`'s `loadFindings()`/`refreshFindingCounts()` read `res?.data` and `res?.total` off `getAuditFindingsPaginated()`'s response — exactly the shape that would silently break under `fetchJSON`, which unwraps a `{data: [...]}` body straight to the bare array (confirmed by reading `fetchJSON` itself), leaving `res.data` and `res.total` both `undefined` → `[]`/`0` forever. `fetchRaw`, which the fix switches to, returns `res.json()` with no unwrapping — confirmed by reading it directly. This is the mirror image of #270's Suppliers bug (same root confusion, opposite direction), and both ends of the fix check out.

**Spot-checked three of the eleven "raw database value" fixes against the actual catalogue and wiring**: `Tasks.vue`'s `priorityLabel`/`typeLabel` both route through `enumLabel('priority'|'task_type', ...)`, and `common.json` has `task_type.incident_followup: "Incident follow-up"`; `Audit.vue`'s `auditTypeLabel` routes through `enumLabel('audit_type', ...)` with `audit_type.internal: "Internal"` in the catalogue.

**`formatMonthLong`/`enumLabelAbbr` tested directly**: `formatMonthLong(0)`/`(11)` render "January"/"December", `(null)` degrades to `''`. `enum_abbr` catalogue group exists and is what the new abbreviated-label tests cover.

**Dead code genuinely gone**: grepped for the "Generated N tasks" banner text — no remaining reference anywhere in `web/src/views/`.

Nothing further to add. Approving.

PR #274, opened 2026-09-08
refactor(i18n): fold the six per-register unsaved-changes copies into common (#212)

unidoc-alip:
Removes six duplicate copies of the same three sentences from the translation files, so a translator sees each one once instead of seven times.

## What this is

Every register screen (risks, assets, suppliers, systems, incidents, legal requirements) warns you before you navigate away from an unsaved edit. When those screens were made translatable, each one got its own private copy of the warning. The workflow screens, done afterwards, put the same three sentences in the shared file instead — which is where they belong.

The result is seven identical copies of "You have unsaved changes. Discard and close?" in the translation files. This deletes six of them and points the six screens at the shared one.

## What changes for a user

**Nothing.** Every copy being deleted was character-for-character identical to the shared version, so all 24 places affected show exactly the same sentence as before.

## Why it matters now

A translator is about to start work against a frozen set of these entries. Seven copies of one sentence means seven chances to translate it seven slightly different ways, and the user then sees the wording change as they move between screens. It also means seven times the work for no benefit. Fixing it after translation starts would mean throwing away work already done, which is why it goes in before rather than after.

## Risk

**Very low.** The change is a mechanical rename of which entry each screen reads from, with no wording change on either side. Three things confirm it:

- The deleted entries were verified identical to the shared one before deletion, not assumed to be.
- All 571 translation lookups across the six screens were resolved against the real translation file — none is broken or missing.
- Nothing else in the codebase referenced the deleted entries.

One copy is deliberately kept: the audit screen's version says "unsaved **report** changes", which is genuinely different wording rather than a duplicate.

## Scope

- 18 entries deleted from six translation files.
- 24 lookups repointed across six screens.
- No new entries, no wording changes, no database or API changes.

## Checks

- Front-end suite: 152/152, including the two automated text scanners and the translation-file consistency check.
- Production build: clean.
- Both text scanners report the same numbers as before — this kind of change is invisible to them, and neither tracking file moves.

## Relationship to the other open work

Independent of #272 (the document screens). The two touch no file in common, so they can merge in either order with no rebase.

### DIFF
--- a/web/src/locales/en/assets.json
+++ b/web/src/locales/en/assets.json
@@ -67,10 +67,5 @@
     "warning": "Deleting this asset is permanent and cannot be undone. History, comments, and linked references will be lost.",
     "delete": "Delete asset",
     "confirm": "Delete this asset? This cannot be undone."
-  },
-  "dirty": {
-    "switch_tab": "You have unsaved changes. Discard and switch tab?",
-    "close": "You have unsaved changes. Discard and close?",
-    "discard": "Discard"
   }
 }
--- a/web/src/locales/en/incidents.json
+++ b/web/src/locales/en/incidents.json
@@ -84,10 +84,5 @@
     "warning": "Deleting this incident is permanent and cannot be undone.",
     "delete": "Delete incident",
     "confirm": "Delete incident \"{title}\"? This cannot be undone."
-  },
-  "dirty": {
-    "switch_tab": "You have unsaved changes. Discard and switch tab?",
-    "close": "You have unsaved changes. Discard and close?",
-    "discard": "Discard"
   }
 }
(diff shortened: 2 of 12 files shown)


unidoc-ahall (approved):
Independently verified.

`go build ./...` clean, `web/`: 152/152, `npm run build` clean.

- `common.dirty.*` is character-for-character identical to what was deleted from all six per-register files.
- Grepped for `<register>.dirty.*` across `web/src` — zero remaining references anywhere.
- The audit screen's genuinely-different wording ("unsaved **report** changes") is correctly left as its own `audit.dirty.report_switch_tab`/`report_close` entries, not folded — and its `discard` button already pointed at `common.dirty.discard`, consistent with only the six register screens having had the duplicate.
- Neither `rawText.baseline.json` nor `errorRender.baseline.json` changed — matches "invisible to both scanners."

Small, mechanical, correctly scoped. Approving.

PR #276, opened 2026-09-08
fix(cli): show the API's message instead of the raw error body (#212)

unidoc-alip:
Fixes command-line error messages, which have been printing raw JSON at users since the API error work landed.

## What a user sees

Before:

```
API error 404: {"message":"risk not found","code":"not_found","params":{"entity":"risk"}}
```

After:

```
API error 404: risk not found
```

## How it happened

Recent work added a machine-readable code to API error responses, so the web app can show error messages in the user's own language. The command-line tool prints whatever the server sends back, verbatim — so it started printing those extra fields too. Every converted part of the API regressed the same way: 394 places across 22 modules.

The developer documentation predicted this precisely, and said the fix belonged with the conversion work rather than ahead of it. The conversion happened and the fix did not. That is the whole defect — nothing was mis-designed, a follow-up simply went unclaimed.

## The important half is what it does *not* touch

An error status does not always come from our API. A proxy, a Cloudflare Access challenge or a load balancer answers with an HTML page or plain text, and whoever is debugging a failed request needs to see exactly what arrived. So anything that is not a well-formed error response is passed through untouched:

- HTML from a gateway — verbatim
- plain text — verbatim, whitespace trimmed
- malformed or truncated JSON — verbatim, so the corruption is visible
- a response with a code but no message — verbatim, because that is not the documented shape and hiding it would hide a real problem
- an empty body — reads `(no response body)` rather than ending in a bare colon

A decoder that swallows an unexpected response is worse than no decoder at all, so that behaviour is tested case by case rather than assumed.

## Risk

**Low, and narrow.** One function, one call site, in the command-line client only. Nothing about the server, the API, or the web app changes. Around 520 error responses that were never converted still send a plain message and must read exactly as they did before — including server-fault messages and operator configuration hints like "WebAuthn not configured (set ISMS_BASE_URL)". Those are pinned by tests.

## Checks

- This is the first test file for the client package: 13 cases over the decoding rules, plus two that drive a real local HTTP server so the behaviour is checked on the actual request path.
- That last part matters here specifically. The bug was a helper the documentation asked for that nobody wired in — so a test of the helper alone would have passed against the broken build. Reverting the one-line call makes the end-to-end test fail with the exact JSON string shown above, which is how the test was confirmed to be worth having.
- `go vet ./...` clean; client and API package tests passing.

## Note

Also updates the paragraph in the developer docs that predicted this, so it describes the behaviour instead of promising it.


unidoc-alip:
Copilot round addressed in `85fcce1` — three comments, all three valid.

**The two documentation ones were the same finding in two places.** The function comment and `docs/i18n.md` both said a non-envelope body is passed through "untouched" / "verbatim", while the code trims surrounding whitespace — and a test named *"plain text passes through, trimmed"* already pinned that it does. The trimming is the behaviour I want (a response body ends in a newline more often than not, and one that does should not break the operator's line), so the over-promise was the defect rather than the trim. Both now say content intact, surrounding whitespace trimmed only.

**The whitespace-only `message` was a real defect.** `{"message":"   "}` decoded as non-empty, so the CLI printed `API error 400:` followed by three spaces. The decoded message is now trimmed before the empty check, so whitespace-only counts as no message and falls back to the raw body — consistent with the `{"message":""}` case that was already there. A real message comes back trimmed too, for the same reason the fallback is.

Worth recording that it is **not reachable from this server**: every `errorMessages` value is static prose, and `apiError`'s degraded placeholder path is already `TrimSpace`'d to empty, so it cannot emit a whitespace-only message. Fixed regardless — the only reason this function is defensive is that the body may not be ours, and a blank line after the colon tells an operator strictly less than the body it came from.

Three cases added (whitespace-only, newline-only, and a real message with surrounding whitespace); each fails with the fix reverted, checked rather than assumed.

unidoc-ahall (approved):
Independently verified with two mutation tests against the actual code, not just read.

`go build/vet ./...` clean. `go test ./internal/isms/client/...`: 16/16.

**Reverted the one-line `do()` call site** (`errorDetail(respBody)` → `string(respBody)`) and reran `TestDoRendersTheServersMessage`: fails with the exact envelope string from the PR's own "before" example — `API error 404: {"message":"risk not found","code":"not_found","params":{"entity":"risk"}}` — confirming the end-to-end test genuinely catches the real defect (a correct helper nobody wired in), not just a synthetic one.

**Reverted the whitespace-trim fix** (dropped `strings.TrimSpace` before the empty-message check) and reran `TestErrorDetail`: exactly the three cases that should fail do — whitespace-only message, newline-only message, and the surrounding-whitespace-on-a-real-message case — while every other case, including the fallback-to-raw-body ones, stays green. Confirms the fix is precisely targeted, not overbroad.

Both mutations reverted, `git status` clean throughout.

The fallback design is the right shape for a CLI: passing through anything that isn't a well-formed `{message: ...}` envelope (HTML from a proxy, plain text, malformed JSON, a code-only body) means an operator debugging infrastructure noise still sees the real response, while a converted API error reads clean. The whitespace-only-message defensive case is correctly noted as unreachable from this server today but worth keeping since the function has to be defensive about bodies that aren't ours.

Approving.

PR #280, opened 2026-09-10
Complete the id-ID translation and enable Indonesian (#212)

unidoc-alip:
## What this does

Finishes the Indonesian translation — **10% to 100%**, 244 keys to all 2,448 — and **switches the language on**. Users can now pick Bahasa Indonesia from the language selector, including before they sign in.

It started smaller. We were about to tell a volunteer translator to begin, and our own instructions say "translate `common.json` first, and look at `id-ID` for how it should look". That example did not hold up: `id-ID` was described as shipping "2 of the 25 files", true but flattering, because one of those two was itself half-finished — missing the **Add**, **Approve** and **Reply** buttons, every filter dropdown, every tab label and every role name. Finishing the example meant finishing the language.

## What users will notice

- A **language selector appears** where there was none — on the sign-in page, the landing page, and in Settings. It was hidden until now because a picker with one option is noise.
- Choosing Indonesian translates the whole application, not part of it.
- **Anyone whose browser is set to Indonesian will now get Indonesian automatically**, without choosing anything. That is the intended behaviour of the language precedence chain, but it is the one change that reaches users who never ask for it, so it is worth knowing before release. An explicit choice always wins, and English remains the fallback.

## Impact and risk

**Moderate — this is a user-visible release change**, unlike the earlier translation-only state of this branch.

Nothing changes for English users: English is still the default and still what everyone sees unless they choose otherwise or their browser asks for Indonesian. The change is additive; no English text was touched.

The reversal path is one line. Turning the flag back off removes the language from the picker and rejects it on save, and any preference already stored for it quietly degrades back to English rather than breaking — that behaviour is deliberate and covered by tests.

## Two gates had to pass, and both now do

A language may only be offered when the application is fully extracted *and* its translation covers at least 90% of the text. These measure different things and a reader meets the worse of them.

Indonesian is the reason we have both. Its bundle was **complete** when the app had 162 pieces of text. Extraction then grew that to 2,448 and left the bundle at 244 — under 10% — **without a single line of it changing**. Extraction was done, so the first gate passed; the second is what held it back. It now covers everything.

## What needs a native speaker's eye

Most vocabulary follows the Indonesian ISO terminology table already in our contributor guide, checked against the national adoption of the standard. **Three choices are new and not backed by an adopted standard**, so they were deliberately kept out of that cited table:

1. **"Control" as *Kendali*** for document types — common usage also says *kontrol*.
2. **Data-protection roles** — controller/processor as *Pengendali*/*Prosesor*, following Indonesia's personal data protection law.
3. **"Data protection" as *Pelindungan data***, matching that law's spelling rather than the more common *perlindungan*.

Each is a one-line change. If any should be different, that is worth settling before release rather than after, since these are words users will see everywhere.

## Why the badge codes stay in English

**DOC**, **RISK** and the **C / I / A** confidentiality markers are unchanged, which looks like an oversight, so the reasoning is now in the contributor guide:

- **A translated abbreviation can collide where the full word does not.** Confidentiality and availability both begin with K in Indonesian, so C / I / A would render as **K / I / K**.
- **An abbreviation that is not shorter is not an abbreviation.** *RISK* becomes *RISIKO*.

Where the language has an established short form we use it — audit findings render as *KTS Mayor* / *KTS Minor*, what Indonesian auditors actually write. It is a judgement per abbreviation, not a blanket exemption.

## A test-quality note worth reading

Three tests checked that a switched-off language really is unavailable — rejected on save, ignored from browser settings, degraded to English if it was stored earlier. **They used Indonesian itself as their example of a switched-off language.**

That was fine while something happened to be off. It becomes a trap the moment nothing is: with no switched-off language left, all three keep passing while checking nothing at all — green, and describing a safeguard nobody is exercising. The safeguard still matters, because the next language added starts switched off and depends on it.

They now run against a synthetic test-only language instead, so the coverage is tied to the mechanism rather than to whichever language happens to be unfinished.

**This was checked rather than assumed, and the check paid for itself.** Deliberately switching the test language *on* has to break all three. On the first attempt only two broke — the third had never been wired to the fixture and was passing for the wrong reason, which is precisely the failure the rewrite exists to prevent. Fixed; all three now break under that check and pass without it.

## How the translation was checked

Translation quality needs human review, but the mechanical failures — the ones that produce a broken screen rather than an awkward sentence — were checked by script on every one of the 2,448 keys:

- **Key sets identical** to English: nothing missing, nothing extra.
- **Interpolation slots identical per key**, so no counter or name placeholder was dropped or renamed into a slot the app will never fill.
- **Plural forms match English structurally**, so a count of zero still renders through the branch the original intends.
- **Special characters preserved** — the 11 placeholders containing an `@` keep their escaping, and no unescaped `@` was introduced. Either would break the page.

The checker was proven to work by deliberately breaking a key and confirming it was caught. (Its first two runs were silently checking nothing, from a path mistake, and reported a clean pass — which is why proving it fires is a step and not a formality.)

**Then it was run for real in a browser**: 17 screens with the language on, no untranslated placeholders leaking through, no text overflowing its container. The overflow check was itself validated by padding a label until it did overflow — an earlier version of it detected nothing at all and would have reported clean regardless.

## Verification

`go test ./internal/isms/i18n/...` and `go vet` pass. `npm test` 182/182, with the missing-key warning silent for the first time. Production build clean.

Live against an unpatched build: the language list advertises both, the sign-in picker appears and switches the page to Indonesian, saving the preference returns success and reads back correctly, and an unsupported language code is still rejected.

## Related

Enabling a second language incidentally hides #281 — the empty "Language" label on Settings only appears when a deployment has exactly one language. The bug is still real for single-language deployments and is tracked separately.


unidoc-alip (commented):
Review the translation and verify it visually, looks good

unidoc-ahall (approved):
Independently verified the mechanical claims — I don't have the Indonesian to review the translation quality itself, so I focused on everything checkable by script, plus one gap I found that isn't.

`go build/vet ./...` clean. `web/`: 182/182, `npm run build` clean.

**The core numeric claims, computed independently:**
- `en` keyset unchanged at 2448, matching the frozen snapshot exactly (no drift).
- `id-ID` now has exactly 2448 keys, zero missing against the frozen keyset — genuine 100% coverage, not read off the description.
- The 11 `{'@'}`-escaped placeholders in `en` are all present in `id-ID` too.

**Both release gates pass live**: `TestOnlyEnabledLocalesAreOffered`, `TestEnabledLocalesAreTranslated`, and the three disabled-locale tests all green as shipped.

**Mutation-tested the `gatedTag` fixture rewrite**, since the PR's own account of it (two of three broke on the first attempt, not three) is exactly the kind of claim worth reproducing rather than trusting: flipped the synthetic `zz-ZZ` fixture to `enabled: true` and reran `TestDisabledLocaleIsNotCanonical`, `TestDisabledLocaleDegradesToDefault`, and `TestDisabledLocaleIsNotNegotiated` — all three genuinely fail, with the exact diagnostics you'd expect (`Canonical("zz-ZZ") = ("zz-ZZ", true), want rejected`, etc.). Reverted, `git status` clean.

**One real gap, found by testing it rather than reading the description**: the PR states the translation was checked by script for "interpolation slots identical per key" across all 2448 keys, but no such check is committed — this PR touches no test or script files besides the gate machinery and `test_coverage_boost.py`. I renamed `{email}` to `{surel}` in `id-ID/auth.json`'s `signup.sent` frame (a real UI string, not a notification or error code — those two surfaces already have AST-derived placeholder-parity tests from earlier PRs in this series) and reran the full suite: **182/182 still passes.** A translator silently dropping or renaming an interpolation slot in a general UI string — exactly the bug class this whole series has been closing everywhere else, with real teeth (`TestFramesOnlySpendGuaranteedParams`, `TestPlaceholdersMatchInEveryLocale`) — has no regression protection for the 2448 general UI keys now shipping live to real users.

This doesn't call the *current* content into question — it was checked once by an ad-hoc script per the description, and I have no reason to doubt that ran. But it's the one surface in this entire i18n effort where the safety net was described rather than committed, and now that id-ID is live, the next edit to either bundle has nothing to catch this. Worth a fast-follow porting the general shape of `TestPlaceholdersMatchInEveryLocale` (already proven in `errors_test.go`) to cover the full UI keyset, not just error/notification frames. Not blocking this PR — the gate-flip itself and everything mechanically checkable here is sound — but flagging clearly since it's a real, reproduced hole rather than a hypothetical one.

The three flagged terminology decisions (kendali/pengendali-prosesor/pelindungan) and the abbreviation reasoning (C/I/A colliding as K/I/K, RISK→RISIKO not being shorter) are exactly the kind of judgment calls that need the native-speaker sign-off the PR itself asks for — I can't verify Indonesian correctness, only that the reasoning given is internally consistent and documented for the next translator.

Approving — the release-gate mechanics are solid and independently reproduced. Recommend opening the placeholder-parity follow-up separately rather than blocking on it here.

unidoc-alip:
Thanks @unidoc-ahall — the placeholder-parity gap is real and I've opened it as a fast-follow: #282.

You're right that the check described in the PR body was never committed. The issue records your reproduction (renaming `{email}` → `{surel}` in `id-ID/auth.json`'s `signup.sent` leaves the suite at 182/182 green) and scopes the fix as a Vitest case alongside `web/test/localeKeyset.test.js`, comparing slot sets per key shared between `en` and each locale, with a missing key staying a skip so a lagging translation doesn't block unrelated work.

Merging this one as approved; the parity test lands separately.

PR #283, opened 2026-09-10
Hide the Language setting entirely when there is only one language (#281)

### LINKED ISSUE #281: Settings shows an empty "Language" label when the deployment has only one language
## What happens

On **Settings → Profile**, a **Language** label appears with nothing underneath it, followed by a blank gap before **Role**.

This affects every deployment today. The language picker deliberately hides itself when there is only one language to choose from, and Indonesian — the only other language in the product — is currently switched off. So the picker is hidden for everyone, and the label it belongs to is left stranded.

The rendered markup is the whole story:

```html
<div><label>Language</label><!----><!----></div>
```

The label is real; both things that should follow it render nothing.

## Expected

**When only one language is available, nothing language-related should appear at all** — no label, no empty row, no leftover gap. The section should read Email → Name → Role, as though the setting did not exist.

When a second language is enabled, the label and the picker should both appear together.

## Cause

`LocalePicker` hides itself with a condition on its own root element:

```
web/src/components/LocalePicker.vue:4
  <select v-if="options.length > 1" ...>
```

But its caller renders the label, outside the component and unconditionally:

```
web/src/views/Settings.vue:25-29
  <div>
    <label ...>{{ $t('common.locale.label') }}</label>
    <LocalePicker @error="localeMsg = $event" />
    ...
  </div>
```

Only the component knows whether it will render. The caller cannot know, so it always draws the label. The empty wrapper `<div>` is also still a child of a `space-y-4` container, which is where the extra vertical gap comes from.

## Suggested fix

Move the label inside `LocalePicker`, behind the same condition that already governs the picker — for example an optional `label` prop, with the component rendering a wrapper that carries both.

The point is that **the visibility rule should exist in exactly one place**. Duplicating `options.length > 1` into `Settings.vue` would fix the symptom and set up the next bug: two copies of a condition that must agree, with nothing enforcing that they do.

## Other places to check — two are fine, one is worth a look

There are three `LocalePicker` call sites. I checked all three:

- **`views/Settings.vue:27`** — the bug above. The only one with a visible artifact.
- **`views/Login.vue:202`** — no label, but the picker sits alone in `<div class="mt-8 flex justify-center">`. With the picker hidden the wrapper renders empty and still contributes its `mt-8`, so roughly 32px of dead space is left below the sign-in card. No stray text, and the login footer is fixed-position so nothing visibly shifts — worth tidying in the same change rather than filing separately. **Reasoned from the markup, not observed in a browser**, unlike the Settings case.
- **`views/Landing.vue:171`** — clean, and worth recording so nobody "fixes" it. The picker shares a `flex … gap-4` row with the powered-by text; a gap only applies *between* children, so with the picker hidden the spacing collapses on its own.

## The general pattern, for whoever picks this up

A component that hides itself while its caller draws chrome for it is the actual defect class here, so I checked the other components that do the same thing:

- `Pagination.vue` (`v-if="totalPages > 1"`) — 13 call sites, all inside table footers with no adjacent labels. Clean.
- `SuggestNewButton.vue` (`v-if="canSuggest"`) — 12 call sites, all in flex action rows. Clean.
- `EntityReferences.vue` (`v-if="refs.length > 0"`) — **has no call sites at all.** Unrelated to this issue, but it means the component is dead code and nothing renders it. Probably wants its own issue.

## How to reproduce

Sign in and open Settings. The stranded label is visible immediately, since Indonesian ships switched off.

To confirm the diagnosis, temporarily set `enabled: true` for `id-ID` in `internal/isms/i18n/locale.go` and restart: the picker appears and the label becomes correct. (Doing this makes `TestOnlyEnabledLocalesAreOffered` fail, which is expected — that test pins the shipped language set.)

unidoc-alip:
Fixes #281.

## What users see

On **Settings → Profile**, a **Language** heading appeared with nothing under it and an extra gap before **Role**. The language picker deliberately hides itself when a deployment offers only one language, but the heading was written into the Settings page separately, so it stayed behind on its own. Now the whole setting disappears together — the section reads Email → Name → Role, as though it were never there — and when a second language is available the heading and the picker appear as a pair.

The sign-in page had a milder version of the same thing: an invisible block of empty space where the picker would have been. That is gone too.

## Why it happened, in one line

Two places had to agree about whether the setting was visible, and only one of them could actually know. The picker decided for itself; the page around it guessed and drew the heading regardless. The decision now lives in one place, so there is nothing left to disagree.

## Scope and risk

Small and contained: the language picker component, plus the two pages that surround it. No API, database, or server changes, and **no new dependencies**. Nothing about *which* languages are offered changes — only whether the setting is drawn at all.

Note on when this is visible today: the product currently ships two languages, so the picker renders and the reported empty heading is not on screen right now. The underlying flaw is still live — it returns for any single-language deployment, and during a rolling deploy where the server offers a language the browser's copy of the app cannot yet display. So this is a fix worth having, not an urgent one.

## Verification

- 188 automated web tests pass. Five are new and cover this behaviour directly; deliberately putting the bug back makes two of them fail, so they will catch a repeat.
- Checked by hand in a browser against a local server, on the two-language configuration that ships today: the Settings rows are evenly spaced with the heading correctly attached to the control, the footer picker on the marketing page sits level with the text beside it, the sign-in picker is centred where it should be, and a deliberately failed save still shows its error message in the right place with the picker snapping back to the previous language.

## One thing reviewers may notice

The test suite could not previously render a Vue component at all — tests that looked at component files read them as plain text, which cannot catch a visual regression like this one. This adds a small shared helper that renders a component for real, through the same start-up path the live app uses, so the fix is verified by what the component actually produces rather than by what its source looks like.

It needs nothing new installed: the compiler it uses ships inside the Vue package as a published entry point, and the browser-simulation library was already in use. One incidental correction came with it — a file extension missing from an import, which the app's build tool tolerated and plain Node did not.

### DIFF
--- a/web/src/components/LocalePicker.vue
+++ b/web/src/components/LocalePicker.vue
@@ -1,33 +1,54 @@
 <template>
   <!-- Hidden entirely when the deployment ships a single locale: a picker with
-       one option is noise, and it would suggest a choice that does not exist. -->
-  <select v-if="options.length > 1" :value="selected" @change="onChange"
-    :disabled="saving"
-    :aria-label="$t('common.locale.label')"
-    :class="compact
-      ? 'bg-transparent border border-slate-800 rounded-lg px-2 py-1 text-xs text-slate-400 hover:text-slate-200 focus:outline-none transition-colors disabled:opacity-50'
-      : 'w-full px-3 py-2 bg-slate-800 border border-slate-700 rounded-lg text-sm text-white focus:outline-none focus:border-blue-500 disabled:opacity-50'">
-    <!-- Only offered when signed in: with no account to store it against,
-         "follow the org default" is not a state the user can be in. -->
-    <option v-if="persist" value="">{{ $t('common.locale.org_default') }}</option>
-    <option v-for="l in options" :key="l.tag" :value="l.tag">{{ l.name }}</option>
-  </select>
+       one option is noise, and it would suggest a choice that does not exist.
+       The wrapper is inside that condition, and so is the label, because only
+       this component knows whether it will render. A caller drawing its own
+       label had no way to find out, and left it stranded above nothing
+       (#281). Anything a call site wants beside the picker belongs in the slot
+       for the same reason.
+
+       Attribute fallthrough still works: when hidden the root is this comment
+       plus a `v-if` placeholder, which Vue treats as a single comment root and
+       exempts from the extraneous-attrs warning, so a caller's classes land on
+       the wrapper when there is a wrapper and are dropped when there is not. -->
+  <div v-if="options.length > 1">
+    <label v-if="label" :for="selectId" class="block text-xs text-slate-500 mb-1">{{ label }}</label>
+    <select :id="selectId" :value="selected" @change="onChange"
+      :disabled="saving"
+      :aria-label="label ? undefined : $t('common.locale.label')"
+      :class="compact
+        ? 'bg-transparent border border-slate-800 rounded-lg px-2 py-1 text-xs text-slate-400 hover:text-slate-200 focus:outline-none transition-colors disabled:opacity-50'
+        : 'w-full px-3 py-2 bg-slate-800 border border-slate-700 rounded-lg text-sm text-white focus:outline-none focus:border-blue-500 disabled:opacity-50'">
+      <!-- Only offered when signed in: with no account to store it against,
+           "follow the org default" is not a state the user can be in. -->
+      <option v-if="persist" value="">{{ $t('common.locale.org_default') }}</option>
+      <option v-for="l in options" :key="l.tag" :value="l.tag">{{ l.name }}</option>
+    </select>
+    <slot />
+  </div>
 </template>
 
 <script setup>
-import { computed, ref } from 'vue'
-import { useLocale } from '../composables/useLocale'
+import { computed, ref, useId } from 'vue'
+import { useLocale } from '../composables/useLocale.js'
 import { renderApiError } from '../composables/useApiError.js'
 
 const props = defineProps({
   // false on login / landing, where there is no session to save against.
   persist: { type: Boolean, default: true },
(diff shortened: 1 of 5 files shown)


unidoc-ahall (approved):
Independently verified.

`go build ./...` clean. `web/`: 188/188 (182 + 5 new), `npm run build` clean.

**Mutation-tested the fix**: reverted `LocalePicker.vue` to its pre-fix shape (`v-if` on the bare `<select>`, no `label` prop, no wrapper `<div>`, no slot) and reran the new suite — 3 of 5 tests fail (the "two locales render label+select+slot together" and "class lands on the wrapper" tests fail because the wrapper and label plumbing are gone entirely; the class-fallthrough one fails because there's no `<div>` for a caller's class to land on any more). That's a broader revert than the PR's own "put the bug back" claim of 2 failures — I reverted the whole component shape rather than narrowly reintroducing just the visibility decoupling — but it confirms the same thing either way: these tests are not vacuous, they genuinely catch the component regressing. Restored, `git status` clean, all 5 pass again.

The `renderSfc.js` helper is a real addition to the test suite's capabilities (first actual SFC compile-and-mount in this repo, as the PR notes) and the design reasoning holds up: using `vue/compiler-sfc`'s public subpath avoids a new dependency, and mounting via `createApp().mount()` into jsdom rather than `renderToString` matches the app's actual runtime path since there's no SSR here. The "import this before anything that touches vue" constraint is a real footgun but it's clearly flagged in the header and enforced by convention (single point of documentation) rather than silently relied on.

The `useId()`-based `for`/`id` association is genuinely exercised and passing ("the label points at the select it labels"), and the aria-label-vs-visible-label mutual exclusivity (no redundant accessible name when a visible label exists) is correctly tested too.

Small, well-scoped, good new test infrastructure that the rest of the suite can reuse for future component-behavior regressions. Approving.

PR #284, opened 2026-09-11
Make record references like ASSET-3 resolve to the right record everywhere (#201)

### LINKED ISSUE #201: One `:id` param, three different resolution strategies — display identifiers are looked up, arithmetically stripped, or rejected outright depending on the route
### What happened?

Every register entity has a display identifier (`ASSET-3`, `CR-1`, `ISMS2026-2`), #185 put a **Copy link** button on every detail header, and `#ENTITY-N` references make those identifiers the way people refer to entities. But the API has three different implementations of "turn `:id` into a row", and which one a route uses is unrelated to which module it belongs to:

| strategy | what it does with `ASSET-3` | routes |
|---|---|---|
| **Lookup** — `s.resolveAssetID` etc. (`server.go:3170-3298`) | `SELECT … WHERE identifier = 'ASSET-3'` — correct | all `:id` routes for incident, corrective action, task, program and legal requirement — **except legal's two readings routes**; plus the `GET` one-entity route for asset, risk, system, supplier, objective and change request |
| **Arithmetic strip** — `parseID` (`server.go:3153-3158`) | `TrimPrefix` → `3` → treats it as the **primary key** | asset/risk/system/supplier `PUT` + `DELETE`; all 10 readings routes (`api_readings.go`); all 4 audit-finding routes; the 3 access-review routes; `handleRiskAdvisories` |
| **Strict numeric** — `strconv.Atoi` / `ParseInt` | `400 invalid id` | every objective write and sub-resource (12 routes); change-request `PUT`/`DELETE`/status; asset-review and supplier-review sub-resources; the whole audit module (11 routes) |

Three answers to the same question, and the resolver that gets it right already exists for all ten entity types that have identifiers.

**The arithmetic strip is the part worth fixing first, because it is silently wrong rather than loudly wrong.** `parseID` hand-maintains a prefix list:

```go
func parseID(s string) (int64, error) {
	for _, prefix := range []string{"RISK-", "ASSET-", "AST-", "SUPPLIER-", "SYSTEM-", "FIND-"} {
		s = strings.TrimPrefix(s, prefix)
	}
	return strconv.ParseInt(s, 10, 64)
}
```

Identifiers are allocated from a **per-org** counter — `identifier_sequences`, `INSERT … ON CONFLICT DO UPDATE SET next_value = next_value + 1` (`db/postgres.go:65-87`) — while primary keys come from a **global** `GENERATED ALWAYS AS IDENTITY` column. The two agree only in the first organization of an install, and only until a delete. Everywhere else `ASSET-3` and PK 3 are different rows, so:

- **Ordinary case: a confusing 404.** `handleUpdateAsset` (`server.go`) strips to `3` and calls `GetAsset(ctx, orgID, 3)`, whose query is `WHERE id = $1 AND organization_id = $2` (`db/assets.go:122-127`). If PK 3 belongs to another org the row isn't visible and the caller gets `asset not found` for an asset they are looking at in the UI.
- **Bad case: the write lands on the wrong entity.** When PK 3 *does* exist in the same org — normal in any org whose creates interleaved with another org's, or after deletions — the update, the delete, or the reading is applied to a different asset of that same org, with a 200 and a changelog entry attributing it to the caller.

This is precisely the defect **#177** fixed for suggestion-apply ("resolve seq-based identifiers via lookup"), and the warning at `db/changes.go:109-111` spells it out for `CR-` references citing #174/#177. The HTTP write routes never got the same treatment. **#172** ("entity endpoints accept identifier form") appears to have been closed on the strength of this strip — it makes the identifier form *accepted*, not *resolved*.

Two smaller signs both prefix lists are hand-maintained and drifting: `AST-` is still in `parseID` although #173 removed that prefix from the minting path (its only other survival is an `AST-001` example in `openapi.yaml:3356`), the minting map carries three entity types nothing calls it for (above), and `LEGAL-`, `CR-`, `TASK-`, `INC-`, `CA-`, `PROG-` and objective display ids were never in it — which is exactly why those types 400 instead.

### Steps to reproduce

**(a) The rejections — observed during the run.**

```
GET    /objectives/ISMS2026-3            → 200        GET  /changes/CR-1                → 200
GET    /objectives/ISMS2026-3/checkins   → 400        PUT  /changes/CR-1/status         → 400
POST   /objectives/ISMS2026-3/checkins   → 400        POST /legal/LEGAL-1/readings      → 400
PUT    /objectives/ISMS2026-3            → 400        GET  /suppliers/SUPPLIER-1        → 200
DELETE /objectives/ISMS2026-3            → 400        GET  /suppliers/SUPPLIER-1/reviews → 400
POST   /objectives/ISMS2026-3/archive    → 400
```

Five different error strings for one cause, none of which tells a caller that the sibling route accepts the form it just rejected: `invalid id`, `invalid objective id`, `invalid change request id`, `invalid legal id`, `invalid supplier id`.

**(b) The arithmetic strip — proven on a single-org install, no special setup.** All of the following was run against a live server at `dd3b6b2`; the only entity created was a throwaway asset, which step 6 deletes.

1. `POST /api/v1/assets` → `201`, `{"id":12,"identifier":"ASSET-12"}`.
2. Confirm the read path resolves by lookup — a wrong prefix must not match:
   ```
   GET /api/v1/assets/ASSET-12   → 200  identifier=ASSET-12
   GET /api/v1/assets/RISK-12    → 404  asset not found      ← correct
   GET /api/v1/assets/12         → 200  identifier=ASSET-12
   ```
3. Now the same identifiers on the **write** path:
   ```
   PUT /api/v1/assets/RISK-12      {"notes":"written via RISK-12"}      → 200  identifier=ASSET-12
   PUT /api/v1/assets/SUPPLIER-12  {"notes":"written via SUPPLIER-12"}  → 200  identifier=ASSET-12
   PUT /api/v1/assets/FIND-12      {"notes":"written via FIND-12"}      → 200  identifier=ASSET-12
   PUT /api/v1/assets/AST-12       {"notes":"written via AST-12"}       → 200  identifier=ASSET-12
   ```
   A **risk** identifier, a **supplier** identifier, an **audit-finding** identifier and a retired prefix all write to asset 12. `GET /assets/ASSET-12` afterwards returns `notes: "written via AST-12"` — the last write won. The route never checked that the identifier belongs to an asset, or to anything.
4. The sub-resources behave the same: `POST /api/v1/assets/RISK-12/readings` → `201`, with the reading attached to `entity_id: 12`.
5. Other modules, checked non-destructively — an invalid enum value proves the id resolved before validation ran, without changing any data:
   ```
   PUT /api/v1/risks/ASSET-13     {"status":"bogus"} → 400 invalid status: "bogus" (allowed: [draft open closed])
   PUT /api/v1/systems/ASSET-1    {"status":"bogus"} → 400 invalid status: … [active under_review decommissioned]
   PUT /api/v1/suppliers/FIND-1   {"status":"bogus"} → 400 invalid status: … [active under_review suspended terminated]
   ```
   Each reached the entity. Had `parseID` rejected the mismatched prefix, these would have returned `invalid risk id` — which is exactly what the strict-numeric routes do return:
   ```
   PUT /api/v1/changes/CR-1/status {"status":"bogus"} → 400 invalid change request id
   PUT /api/v1/objectives/ISMS2026-3 {"status":"bogus"} → 400 invalid id
   ```
6. Deletion too: `DELETE /api/v1/assets/FIND-12` → `200 {"status":"deleted"}`, and the asset is gone (`GET /assets/ASSET-12` → `404`; `deleted_at` set on row 12).

**Why this matters beyond the cosmetic.** On a single-org install created in one pass, the per-org identifier counter and the global PK sequence coincide, so `ASSET-12` → PK 12 lands on the right row by luck. They do not coincide in general — identifiers come from a per-org counter, PKs from a global sequence — and then the same arithmetic sends the write elsewhere:

- if PK 12 belongs to another org, the caller gets `404 asset not found` for an entity they are looking at;
- if PK 12 exists in the caller's own org, the write, delete or reading is applied to **that** asset instead, with a `200` and a changelog row attributing it to the caller.

The multi-org divergence is code-derived (`identifier_sequences` in `db/postgres.go:66-87` vs the `GENERATED ALWAYS AS IDENTITY` PKs) — the single-org proof above is what was executed.

### Expected vs. actual

**Expected:** any route that takes an entity's `:id` accepts both the numeric id and the display identifier, resolving the identifier by lookup, and returns a consistent error naming what it accepts.

**Actual:** identifiers work on most `GET`-one routes, are silently converted to a possibly-unrelated primary key on the asset/risk/system/supplier write and reading routes, and are rejected with a bare 400 on every objective write, every objective sub-resource, change-request writes, and the review sub-resources.

Suggested direction:

1. **Replace `parseID` with the existing `resolveXID` helpers** at its call sites, and delete the prefix list. The helpers exist for all ten types (`server.go:3170-3300`) and are already used by the `GET` handlers next door — for most call sites this is a one-line swap. That closes the wrong-entity hazard, not just the 400s.
2. **Give the strict-numeric routes the same helper**, starting with objectives, whose 12 routes make it the worst module: `GET /objectives/ISMS2026-3` works and *nothing else about that objective does*.
3. **One shared error string** — "expected a numeric id or an identifier like `ASSET-3`" — instead of `invalid id` / `invalid objective id`.
4. **A test worth adding:** for every entity type with an identifier, drive its full route set with the display form and assert none returns 400. Table-driven, and it is the only thing that will stop this drifting again — the three strategies are individually reasonable and only the sweep makes the inconsistency visible.

Worth deciding as part of this: **the Audit module has no display identifiers at all** (no `identifier` or `display_id` column on `audits`, `audit_programmes` or `audit_findings`), yet #185 gave both Audit detail headers a **Copy link** button, and `handleGetAuditFinding` runs `parseID` for a `FIND-` prefix that is never minted: `NextIdentifier`'s prefix map (`db/postgres.go:66-72`) carries entries for `audit_finding` → `FIND`, `audit` → `AUDIT` and `objective` → `OBJ`, and **no caller passes any of those three entity types** — objectives build their own `program.key + seq` display id instead. Either the module gets identifiers like every other register, or its copy-link is documented as a numeric-id link. That's a design call rather than part of this fix.

### Component / surface

REST API

### Version

`dd3b6b2` (0.7.x, unreleased `master`).

### Anything else?

**Both clients already work around this, which is a decent measure of the cost:**

- The **CLI** resolves display ids itself before calling the API: `isms objective checkin add <display-id>` fetches the whole objective list and matches `DisplayID` in Go (`cmd/isms/objective.go:348-368`), and `isms change status <id|ref>` calls a client-side `resolveChangeID` (`cmd/isms/change.go:216`) because `PUT /changes/CR-1/status` would 400. An extra full-list round trip per command, to do what the server has a helper for.
- The **web UI** never sends an identifier on a mutation: it fetches by identifier, then uses the numeric `.id` off the fetched object (`Risks.vue:1104`, `Assets.vue:630`, `Systems.vue:812`, `Suppliers.vue:730`). So browser deep-links render *and* edit correctly — the report's claim that they don't is wrong, and this issue is not a UI bug.
- Where `isms asset edit ASSET-3` / `rm` do pass the identifier straight through (`internal/isms/client/client.go:277-289`), they land on the arithmetic-strip path — correct today in a single-org install, wrong under the conditions above.

**Scope of this report:** the REST API surface and the CLI. The MCP server's tool definitions were **not** examined, so whether agents hit the same routes with display identifiers is unverified — worth a look before scoping the fix, since MCP is the surface most likely to be handed an identifier out of a comment or a suggestion payload.

*Found during a multi-user manual test run against `dd3b6b2` on 2026-08-05 (the 400s in section (a) are observed responses); the `parseID` hazard was found by reading the code at that commit and then **confirmed against a live server on 2026-08-06** — every status code in section (b) is a real response, not a prediction. Only the multi-org divergence remains code-derived.*

unidoc-alip:
Closes #201.

## What was wrong

Every record in the registers has a short reference people actually use — `ASSET-3`, `CR-1`, `LEGAL-2`. Those references are how colleagues point each other at things in comments, in chat, and through the Copy link button.

The API had three different ways of turning one of those references into the right record, and which one a given operation used had nothing to do with which module it belonged to. Looking a record up usually worked. Changing or deleting one often did not, and some operations refused the reference outright.

The dangerous case was silent. For most editing operations the server took `ASSET-3`, threw away the `ASSET-` part, and used the leftover `3` as an internal row number. Those two numbers are not the same thing. Each organization counts its own records from 1, while the internal numbering runs across the whole install — so `ASSET-3` in your organization is almost never internal row 3. They line up only in the very first organization on a fresh install, which is exactly why this was never noticed in testing.

When they don't line up, one of two things happened:

- **"Not found" for a record that's on screen.** The leftover number belonged to another organization's record, which you can't see, so the edit failed with a confusing error.
- **The edit landed on the wrong record.** If that number happened to be a different record of your own organization, the change, deletion, or reading was applied to *that* one instead — and it succeeded, with a history entry recording you as the author.

The server also never checked that the reference matched the kind of record being changed. Editing an asset while quoting a *risk* reference edited the asset anyway.

## Are customers affected?

**Yes, but narrowly, and not through the web app.** The web interface happens to sidestep all of this: it looks a record up by its reference and then edits it by internal number, so it was never exposed. Anyone using the command-line tool or an AI agent could hit it, and the AI agent surface was the most exposed of all — it stripped references even more aggressively than the API did, converting the one operation that worked correctly into one that didn't.

One thing **is** visible in the web app, and it's worth naming because it looks like a design choice rather than a bug. Where a record links to a change request or a task, the link chip shows the linked item's title. Under the old code it could show the bare reference — `CR-1` — instead of the title, because the label lookup used the same faulty arithmetic. Checked in a browser before and after: the chip now reads the change request's actual title.

No data needs repair. Nothing about how references are stored or displayed has changed — only which record the server resolves them to.

## What this changes

References are now resolved by looking them up, everywhere, which is what a handful of operations were already doing correctly. Concretely:

- Editing, deleting and recording readings against assets, risks, systems, suppliers and legal requirements now reliably affect the record you named.
- Operations that used to reject a reference with a bare "invalid id" now accept it: all change-request edits, six objective operations, supplier and asset review entries, and legal-requirement readings, which had never worked by reference at all.
- Quoting the wrong kind of reference is now a clean "not found" rather than a successful edit of something else.
- The AI agent surface passes references through untouched instead of mangling them first.
- The command-line tool no longer needs the workarounds it carried for this — it used to download the entire list of objectives and search it locally just to translate a reference the server should have understood. Those workarounds are now redundant; removing them is separate work.

Audit findings are the one deliberate exception. Their reference genuinely *is* their internal number, so the old arithmetic is correct there and was left alone.

## Verification

Full test suite passes. The project's integration suite was run against a live server: 704 passed, 1 unrelated failure caused by a rate-limit setting in the local run, confirmed by re-running with that setting restored.

The two new database-backed tests were checked against the *old* code as well as the new, because a test that passes either way proves nothing here. Against the old code they fail exactly as reported in #201 — an edit by reference returning "not found" across organizations, and four mismatched-reference edits returning success and writing to the wrong record. A first draft of one test passed against the old code by luck and was rewritten until it didn't.

Also exercised by hand against a live server, in an organization where the numbering had genuinely drifted (`ASSET-1` was internal row 8, `LEGAL-1` was row 12, `CR-1` was row 22): the command-line tool editing by reference, every previously-rejected operation, and the link chip in a browser before and after.

A guard test now fails the build if this shortcut is reintroduced anywhere. It was checked by deliberately reintroducing it — and it immediately found three more places doing the same thing that the original report had not covered, which are fixed here too.

## Risk

Low. The numeric form still short-circuits before any lookup, so the web app's behaviour is unchanged and no extra database work is done on its path.

Two things a reviewer should weigh. Some operations that previously returned "invalid id" now return "not found", so anything branching on that distinction would see a different code — the only callers were the command-line workarounds this makes redundant. And the worst realistic outcome of a mistake here is an edit being refused with "not found" when it should have succeeded — visible and safe — rather than the current behaviour of an edit silently landing on the wrong record.

## What's left

A broader test sweep. The new tests drive read and write operations in both reference and numeric form for four record types; change requests, objectives, readings and deletions are not yet covered. Worth following up, not a blocker — the guard test catches the specific mistake this fixes.

### DIFF
--- a/internal/isms/api/api_asset_reviews.go
+++ b/internal/isms/api/api_asset_reviews.go
@@ -1,8 +1,8 @@
 package api
 
 import (
+	"errors"
 	"net/http"
-	"strconv"
 
 	"github.com/labstack/echo/v4"
 	"isms.sh/internal/isms/db"
@@ -26,9 +26,12 @@ func (s *Server) handleCreateAssetReview(c echo.Context) error {
 	ctx := c.Request().Context()
 	actor := getUserEmail(c)
 
-	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
-	if err != nil {
+	// Accept the numeric id OR the display identifier, resolved by lookup (#201).
+	id, err := s.resolveAssetID(ctx, orgID, c.Param("id"))
+	if errors.Is(err, errInvalidID) {
 		return errInvalidEntityID("asset")
+	} else if err != nil {
+		return errNotFound("asset")
 	}
 
 	// Verify asset exists in this org (cross-org safety).
@@ -69,9 +72,12 @@ func (s *Server) handleCreateAssetReview(c echo.Context) error {
 
 func (s *Server) handleListAssetReviews(c echo.Context) error {
 	orgID := getOrgID(c)
-	id, err := strconv.ParseInt(c.Param("id"), 10, 64)
-	if err != nil {
+	// Accept the numeric id OR the display identifier, resolved by lookup (#201).
+	id, err := s.resolveAssetID(c.Request().Context(), orgID, c.Param("id"))
+	if errors.Is(err, errInvalidID) {
 		return errInvalidEntityID("asset")
+	} else if err != nil {
+		return errNotFound("asset")
 	}
 
 	reviews, err := s.db.ListAssetReviews(c.Request().Context(), orgID, id)
--- a/internal/isms/api/api_audit.go
+++ b/internal/isms/api/api_audit.go
@@ -641,7 +641,11 @@ func (s *Server) handlePaginatedAuditFindings(c echo.Context) error {
 
 func (s *Server) handleGetAuditFinding(c echo.Context) error {
 	orgID := getOrgID(c)
-	id, err := parseID(c.Param("id"))
+	// Findings are the one register whose display id IS its primary key —
+	// db.SoftDeleteAuditFinding mints "FIND-<id>" from the row id, and
+	// audit_findings has no identifier column to look up. Stripping is correct
+	// here, and only here (#201).
+	id, err := strconv.ParseInt(stripPrefix(c.Param("id"), "FIND-"), 10, 64)
 	if err != nil {
 		return apiError(http.StatusBadRequest, CodeInvalidEntityID, Entity("audit_finding"))
 	}
@@ -726,7 +730,7 @@ func (s *Server) handleUpdateAuditFinding(c echo.Context) error {
 		return err
 	}
 	orgID := getOrgID(c)
-	id, err := parseID(c.Param("id"))
+	id, err := strconv.ParseInt(stripPrefix(c.Param("id"), "FIND-"), 10, 64)
 	if err != nil {
 		return apiError(http.StatusBadRequest, CodeInvalidEntityID, Entity("audit_finding"))
 	}
@@ -791,7 +795,7 @@ func (s *Server) handleUpdateAuditFindingStatus(c echo.Context) error {
 		return err
 	}
 	orgID := getOrgID(c)
-	id, err := parseID(c.Param("id"))
+	id, err := strconv.ParseInt(stripPrefix(c.Param("id"), "FIND-"), 10, 64)
 	if err != nil {
 		return apiError(http.StatusBadRequest, CodeInvalidEntityID, Entity("audit_finding"))
 	}
@@ -833,7 +837,7 @@ func (s *Server) handleDeleteAuditFinding(c echo.Context) error {
 		return err
 	}
 	orgID := getOrgID(c)
-	id, err := parseID(c.Param("id"))
+	id, err := strconv.ParseInt(stripPrefix(c.Param("id"), "FIND-"), 10, 64)
 	if err != nil {
 		return apiError(http.StatusBadRequest, CodeInvalidEntityID, Entity("audit_finding"))
 	}
--- a/internal/isms/api/api_changelog.go
+++ b/internal/isms/api/api_changelog.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"errors"
 	"net/http"
 	"strconv"
 
@@ -27,9 +28,20 @@ func (s *Server) handleListChangelog(c echo.Context) error {
 func (s *Server) handleEntityChangelog(c echo.Context) error {
 	orgID := getOrgID(c)
 	entityType := c.Param("type")
-	entityID, err := strconv.ParseInt(c.Param("id"), 10, 64)
+	// Accept the numeric id OR the entity's display identifier (#201); which
+	// forms are valid depends on the :type, so dispatch on it.
+	entityID, hasIdentifier, err := s.resolveEntityID(c.Request().Context(), orgID, entityType, c.Param("id"))
 	if err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, "invalid entity id")
+		if !hasIdentifier {
+			// No display identifier for this type, so the param could only ever
+			// have been numeric — and entityType may have no entity_inline key
+			// to name in the message.
+			return apiError(http.StatusBadRequest, CodeInvalidID)
+		}
+		if errors.Is(err, errInvalidID) {
+			return errInvalidEntityID(entityType)
+		}
+		return errNotFound(entityType)
 	}
 
 	entries, err := s.db.ListEntityChangelog(c.Request().Context(), orgID, entityType, entityID)
--- a/internal/isms/api/api_collab.go
+++ b/internal/isms/api/api_collab.go
@@ -2718,10 +2718,15 @@ func (s *Server) handleUpdateChange(c echo.Context) error {
 	}
 	orgID := getOrgID(c)
 	ctx := c.Request().Context()
-	id, err := strconv.Atoi(c.Param("id"))
-	if err != nil {
-		return apiError(http.StatusBadRequest, CodeInvalidEntityID, Entity("change_request"))
+	// Accept the numeric id OR the CR-12 identifier, resolved by lookup like
+	// handleGetChange next door (#201). resolveChangeID returns int64 → cast.
+	id64, err := s.resolveChangeID(ctx, orgID, c.Param("id"))
+	if errors.Is(err, errInvalidID) {
+		return errInvalidEntityID("change_request")
+	} else if err != nil {
+		return errNotFound("change_request")
 	}
+	id := int(id64)
 
 	old, err := s.db.GetChangeRequest(ctx, orgID, id)
 	if err != nil {
@@ -2846,10 +2851,15 @@ func (s *Server) handleUpdateChangeStatus(c echo.Context) error {
 	}
 	orgID := getOrgID(c)
 	ctx := c.Request().Context()
-	id, err := strconv.Atoi(c.Param("id"))
-	if err != nil {
-		return apiError(http.StatusBadRequest, CodeInvalidEntityID, Entity("change_request"))
+	// Accept the numeric id OR the CR-12 identifier, resolved by lookup like
+	// handleGetChange next door (#201). resolveChangeID returns int64 → cast.
+	id64, err := s.resolveChangeID(ctx, orgID, c.Param("id"))
+	if errors.Is(err, errInvalidID) {
+		return errInvalidEntityID("change_request")
+	} else if err != nil {
+		return errNotFound("change_request")
 	}
+	id := int(id64)
 
 	old, _ := s.db.GetChangeRequest(ctx, orgID, id)
 	oldStatus := ""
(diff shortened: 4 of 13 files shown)


unidoc-ahall (approved):
Independently verified against a real Postgres database, not mocks — this container has one reachable, so I used it rather than settling for the non-DB test subset.

`go build/vet ./...` clean. Full suite green including the DB-backed tests: `ISMS_TEST_DATABASE_URL=$DATABASE_URL go test ./...` — all packages pass.

**Reproduced the actual reported bug live**, not just read the description: temporarily reintroduced the old `parseID` arithmetic strip in `handleUpdateAsset` (restoring the deleted function, swapping one call site back) and reran `TestUpdateByIdentifierHitsTheRightRowAcrossOrgs` against the real database — it fails with `asset not found`, exactly the "not found for a record that's on screen" failure mode the issue describes, because org B's `ASSET-1` stripped to a primary key that belongs to org A. Restored the real fix and reran — passes. This is the strongest form of verification available here: the actual cross-org bug, reproduced against real data, not simulated.

**Mutation-tested the AST guard** (`TestNoArithmeticIDStripSurvives`) by injecting a `strings.TrimPrefix(c.Param("id"), "RISK-")` into `handleListRiskReadings` — caught immediately, naming the exact file and line. Restored, `git status` clean throughout both mutations.

**Read the resolver design directly**: every `resolve*ID` helper tries numeric-parse first, then `hasIdentifierShape()` to separate "malformed" (400) from "well-formed but not found" (404), then an org-scoped lookup (`GetAssetByIdentifier(ctx, orgID, param)` etc.) — so a cross-org collision on the display identifier can't reach another org's row, which is the actual mechanism of the bug being closed. `entityIDResolvers`'s audit/audit_finding entries are correctly the one deliberate exception (their display id IS the primary key, no identifier column exists to look up), and the guard test's allow-list is scoped to exactly those two literals rather than to a file, which is why it caught the `CR-`/`TASK-` strips in `resolveEntityTitle` during development.

**The MCP change is a clean removal**, not a workaround: `stripPrefix` deleted outright, both call sites now pass the display identifier straight through to routes that resolve it server-side — correctly not duplicating resolution logic on the client side of that boundary.

Genuinely well-executed fix for a serious, silent data-integrity bug, backed by tests that were checked against the old code first (which I independently repeated) rather than assumed to be meaningful. Approving.

PR #285, opened 2026-09-13
Add admin-configurable custom fields on risks

### LINKED ISSUE #213: [Feature]: Custom Risk Categories
### Problem / motivation

Currently you are only allowed to pick from specific risk categories. 

### Proposed solution

Allow the admin to define the Risk Categories, and potential custom fields in risks, the app should come with sane defaults and allow admins to set their own category risks. 

### Alternatives considered

_No response_

### LINKED ISSUE #216: [Feature]: Custom fields on risks
Follow-up to #213. That issue asked for two things: admin-defined risk categories, and custom fields on risks. Categories shipped in #215; custom fields were deliberately left out of that PR because they are a much larger, structurally different feature. This issue tracks the remaining half.

Originally requested by @SonamorN in #213.

### Problem / motivation

The risk register has a fixed schema. Every org gets the same columns — type, origin, category, likelihood/impact, CIA, treatment, owner, dates. Orgs that need to record anything else have nowhere to put it except free text in the description, where it cannot be filtered, sorted, reported on, or kept consistent between risks.

Real examples this blocks: a business-unit or cost-centre attribution, a regulatory reference the risk maps to, an internal scoring dimension the built-in matrix does not model, a "date last discussed at committee" field.

Categories (#213 / #215) solved exactly one dimension of this by making one existing field's vocabulary configurable. Custom fields is the general case: letting an org add dimensions the core does not know about at all. This also aligns with the core-engine principle — anything specific to a particular standard or organisation belongs in configuration, not the core.

### Proposed solution

Per-org custom field definitions on risks: an admin defines a set of fields (label, immutable key, type, optional constraints), and those fields then appear on the risk create and edit forms, on the risk detail view, and ideally in filtering and export.

Field types worth scoping, roughly in order of value: single-line text, number, date, single-select (with its own option list), multi-select, boolean. A first cut could ship text/number/date/select and defer the rest.

### Design questions to settle before implementation

These are the decisions that make this an order of magnitude larger than categories, and they should be answered explicitly rather than discovered mid-build:

- **Storage.** A `JSONB` column on `risks` versus an entity-attribute-value table. JSONB is far simpler and probably right, but the choice constrains filtering and indexing, so it should be deliberate.
- **Definition storage.** Categories fit in an org setting because the payload is tiny. Field definitions carry more structure (types, per-field option lists, ordering, required flags) and likely justify a real table.
- **Validation.** Values must be validated against their definition on every write path — including the suggestion-apply path, which is easy to miss. Categories already showed this: that path had no validation at all and silently injected a default.
- **Definition changes over time.** What happens to stored values when a field is deleted, retyped, or has a select option removed. Categories chose orphan-not-cascade; custom fields need an equivalent rule per case, and retyping is genuinely harder than removal.
- **Reach.** Whether custom fields participate in search indexing, list filters, sorting, CSV/report export, the MCP tools, and the CLI — each is real work and each can be deferred, but the cut line should be explicit.
- **Scope beyond risks.** #213 asked for risks specifically, but the same need exists for every register. Designing risk-only is fine; designing risk-only in a way that cannot generalise is not.

### Alternatives considered

**Keep using the description field.** Works today, costs nothing, but the data stays unstructured — no filtering, no consistency, no reporting.

**Add the requested fields to the core schema.** Contradicts the core-engine principle and does not scale: every org wants different fields, and the union of them is not a coherent schema.

**Extend the category mechanism to more fields.** Only works for fields that are a fixed vocabulary on an existing column. It cannot add a date, a number, or a genuinely new dimension.

unidoc-alip:
## Summary

Admins can now define extra fields on the risk register — free text, a number, a date, or a dropdown of options — and mark any of them as required. Once defined, those fields show up whenever anyone creates or edits a risk, right alongside the built-in fields like title and category.

- Nothing changes for organizations that don't define any custom fields — the risk register behaves exactly as before.
- Fields are managed from Admin → Settings, next to the existing Risk Categories editor. An admin can add, reorder, rename, or remove a field, mark it required, and (for dropdown fields) manage its list of options.
- If an admin deletes a field definition, any risk that already had a value for it keeps that value — it's just no longer editable or shown on the create/edit forms as a labeled field. Nothing is silently deleted.
- A field's type (text/number/date/dropdown) is locked in once created. To change a field's type, delete it and create a new one — this avoids any risk of existing data being misinterpreted.
- Required fields are enforced everywhere a risk is created by a person — the full create form, the quick "Add Risk" shortcut, and the command-line tool — so nobody can skip a field the org has decided must be filled in. Risks created automatically by an AI assistant are exempt from the required-field check, since an assistant can't fill out a form the way a person can.

## Test plan

- [x] Automated tests pass (unit tests, and again against a real database)
- [x] Frontend tests and build pass
- [x] Manually verified in a running instance: defined one field of each type (including a required one), created and edited risks with values, confirmed values display correctly, deleted a field definition and confirmed the risk still opens and keeps its old value

Closes #213
Closes #216


unidoc-ahall (dismissed):
Independently verified against a live Postgres, including mutation-testing the two most load-bearing behaviors. Found one real, reproducible test-coverage gap worth fixing.

`go build/vet ./...` clean. After running this PR's migration (`isms server migrate --dir migrations`) against the test database, the full suite is green including every DB-backed test: `go test ./...` with `ISMS_TEST_DATABASE_URL` set — all pass, 24+ new backend tests across `api`, `db`, and `cmd/isms`. `web/`: 198/198 (188 + 10 new), `npm run build` clean.

**Mutation-tested "the regression that matters most"** (the test file's own words): broke `changedCustomFields` to validate the whole map instead of only changed entries, and reran `TestEditRiskWithOrphanedSelectValueSucceeds` against the real database — it fails exactly as it should (`custom field "severity" value "Low" is not one of the configured options`), confirming the orphan-not-cascade rule (an edit must not be blocked by a value whose select option was later removed) is genuinely enforced by this test rather than passing by construction. Restored, `git status` clean.

**Found a real gap via mutation testing, not reading**: `applyRiskCreate` passes `checkRequired=false` to `ValidateCustomFieldValues` because "agents can't fill out a form" — explicitly documented and clearly the right behavior. I flipped that one argument to `true` and reran the *entire* test suite: zero failures, anywhere. The existing pg test (`TestSuggestionApplyWithInvalidCustomValueRejectedNoRiskCreated`) checks an *invalid* value on the agent path, not a *missing required* one, and the unit test covering the bypass (`TestValidateCustomFieldValues`'s checkRequired=false case) only exercises the pure function directly, never the actual call site. So today, a future refactor that accidentally hardens `applyRiskCreate` back to `true` — exactly the kind of change a well-meaning cleanup could make — would silently start blocking every agent-created risk missing a required field, and nothing in CI would notice. Restored, `git status` clean.

This doesn't call the current behavior into question — I read and verified it does the right thing — but it's the one path in an otherwise very thoroughly tested feature with no regression protection at the integration level. Recommend adding a positive pg test alongside the existing negative one: define a required field, omit it from a suggestion payload, and assert `applyRiskCreate` succeeds. Small addition, not blocking.

Everything else checked out: the migration is idempotent (`ADD COLUMN IF NOT EXISTS`, `ON CONFLICT (key) DO NOTHING`) and applied cleanly. `NormalizeCustomFieldValues` never drops an orphaned key (confirmed by its own unit test and consistent with the design comment). `customFields.js`'s helpers (`orphanCustomValues`, `requiredCustomFieldsSatisfied`, `customFieldInputType`, `deslugFieldKey`) are all genuinely wired into `Risks.vue` — the orphan display, the required-field client-side gating, and the per-type input rendering — not dead exports only covered by their own unit tests. The CLI's client-side required-field check (fails fast with no API call) is tested both for the rejection and the success path.

Approving. Recommend the one follow-up test above, but it's not a blocker.

unidoc-ahall (approved):
Follow-up commit `eeea7ad` closes the one gap flagged in my prior review — re-verified against a live Postgres.

`go build/vet ./...` clean. Full `go test ./...` green with `ISMS_TEST_DATABASE_URL` set, including this new test.

**Mutation-tested it myself, not just trusting the commit message's own account**: flipped `applyRiskCreate`'s `ValidateCustomFieldValues(defs, payload.CustomFields, false)` call site (api_suggestions.go:715) to `true` and reran `TestSuggestionApplyRiskCreateSucceedsWithMissingRequiredField` alone — fails with `custom field "priority" is required`, the exact regression this test exists to catch. Restored, `git status` clean, full suite reran green.

This is precisely the positive pg test I recommended last round (define a required field, omit it from a suggestion payload, assert `applyRiskCreate` succeeds) — the previously-flagged path (a future refactor silently re-hardening the agent-create path) now has real regression coverage. No new concerns.

Approving.

PR #286, opened 2026-09-14
Archive review branches on merge/close so they stop cluttering the branch list (#101)

### LINKED ISSUE #101: Archive merged/closed review branches out of refs/heads (declutter git branch)
review/<id> branches accumulate in refs/heads/* and clutter `git branch` over time.

  Decision (from #6 discussion): do NOT delete them — git refs are cheap, markdown blobs dedupe, and deleting a non-merged proposal would break the per-event diff's
  durability (#6 anchors decision_log.commit_ref and resolves the commit by hash). Governance trail > disk.

  Scope (cosmetic declutter only): on review merge/close, move the ref heads/review/<id> -> refs/reviews/<id> (git update-ref). It disappears from `git branch` but the
  commit stays reachable (GC-safe), so per-event diff keeps working. Belongs in the 'isms server manager' cron or the merge/close handlers.

  Already in place (#106): push ref-protection covers both refs/heads/review/* and refs/reviews/* (server-managed, immutable via push), so archived refs are protected too. No
  DB migration. Low priority — purely keeps `git branch` tidy.

unidoc-alip:
Fixes #101.

## What this changes

When a document is sent for review and the reviewer proposes edits, those edits are stored on a hidden git branch named after the review. Until now those branches were never cleaned up: closing a review left its branch behind forever, so anyone listing branches in an organisation's repository saw a growing pile of finished reviews mixed in with real ones. Merging a review had the opposite problem — it deleted the branch outright.

Now, when a review is merged or closed, its branch is moved into an archive area of the repository instead. It disappears from the branch listing, but the underlying history is kept.

## Why not just delete them

Deleting was the wrong fix, and the merge path was already doing it. The approval trail records the exact commit behind each decision, and the "what changed in this one revision" view reconstructs a diff from that commit. Once a branch is deleted, that commit is no longer anchored to anything and can be discarded by routine repository maintenance — which would quietly break the audit trail for past decisions. Keeping the history and just hiding it costs effectively nothing (git storage here is shared and deduplicated) and preserves the governance record.

## Impact and risk

Low risk, no user-visible feature change and no database migration.

The one real hazard was that several screens look up a review's proposed content by branch name regardless of whether the review is still open — the review's content view, its diff view, the approval content hash, and inline comment context. Had the branch simply moved, those screens would not have failed loudly; they would have silently fallen back to showing the *current published document* instead of what the reviewer actually proposed. That is a correctness problem in the audit trail, not a crash, so it would have been easy to miss.

This is handled centrally: branch-name lookups now fall back to the archive location automatically, so every one of those screens keeps working unchanged. Archiving is also best-effort — if it ever fails, it is logged and ignored, and can never block a merge or a close.

## Verification

Unit tests cover the archive operation, the no-op case when a review never had proposed edits, and the changed merge behaviour. The whole flow was also exercised against a running server with a real repository:

- A **closed** review and a **merged** review both had their branches archived, each pointing at exactly the same commit as before.
- A review left **open** was correctly untouched, so the branch listing showed only the genuinely in-progress review.
- After archiving, the review content view, diff view, per-revision diff and timeline all still returned the reviewer's proposed text rather than falling back to the published document.
- A repository integrity check confirmed no history was left unreferenced.

## Not included

Existing branches from reviews finished before this ships are left alone; this applies going forward only, per the scope agreed on the issue. A retroactive cleanup would need repository access added to the background job that does not exist today, and is better raised separately if it is wanted.


unidoc-ahall (approved):
Independently verified.

`go build/vet ./...` clean, full `go test ./...` green (`ISMS_TEST_DATABASE_URL` set).

**Mutation-tested the actual hazard the PR describes**, not just the happy path: removed the `resolveRef` archive-fallback block (git.go, the `strings.HasPrefix(ref, "review/")` retry against `refs/reviews/<id>`) and reran `TestArchiveReviewBranch` — `ReadFileAtRef("review/42", ...)` fails with `reference not found`, exactly as expected. More importantly, I traced what that failure means at the API layer: every call site that reads a review branch (handleReviewDiff, handleGetReviewContent, decision-log content-hash, inline-comment context) already has its own `if err != nil { raw, err = st.ReadFile(docPath) }` fallback — pre-existing, there to handle "review with no proposed edits". Without the `resolveRef` fix, that pre-existing fallback would silently catch the archived-ref lookup failure too and serve the *published* document instead of the reviewer's proposal, for every closed/merged review — a correctness bug in the audit trail with no error, no crash, nothing to notice. The fix closes that at the one shared choke point (`resolveRef`) rather than patching each call site. Restored, `git status` clean, full suite green again.

**Verified all seven `review/<id>` call sites in api_collab.go** route through `resolveRef` (via `ReadFileAtRef`, `DiffFiles`, or `DiffSuggestion`, which itself calls `ReadFileAtRef`) — none bypass it with a direct ref lookup, so the fallback genuinely covers every screen the PR description lists.

**Checked ordering and cleanup correctness directly in the code**: `ArchiveReviewBranch` creates `refs/reviews/<id>` before removing `refs/heads/review/<id>` (comment explicitly calls out why — never let the commit go momentarily unreferenced), and it's a true no-op (`nil`, not an error) when no branch was ever created, which matters since not every approved review has reviewer edits. `handleUpdateReviewStatus` only calls it on the `closed` transition (that endpoint's sole allowed transition, per the existing status state machine), and `performMerge` calls it once, after `MergeSuggestion` returns — sequential, not nested, so no lock contention with `Store`'s internal `s.mu`. Grepped for other `MergeSuggestion` callers to make sure removing its old auto-delete didn't orphan a branch anywhere else — `performMerge` is the only caller, and it already archives right after.

**One thing I checked expecting a gap and didn't find one**: the wire-protocol ref-protection allowlist (`api_git.go`'s `reviewRefViolation`, from an earlier PR) already treats `refs/reviews/*` as server-managed alongside `refs/heads/review/*` — so a client push can't touch the archive namespace either. That protection predates this PR and was apparently written with this exact archive path in mind; nothing needed updating there.

Correct, well-scoped fix for a real audit-trail integrity risk. Approving.

PR #287, opened 2026-09-15
Surface expiring and expired supplier contracts (#44)

### LINKED ISSUE #44: Surface expiring/expired supplier contracts in overdue & notifications (Annex A 5.20)
## Context

`Supplier.ContractExpiry` (`contract_expiry DATE`, nullable — `db/suppliers.go:58`) is stored, settable, and returned, but **never read** in any overdue, alert, or task path:

- `GetOverdueSummary` (`db/overdue.go:64`) filters suppliers only on `next_review < now`.
- `CreateOverdueReviewTasks` (`overdue.go:294`) derives tasks solely from that summary.
- Nothing outside CRUD/DTOs references `ContractExpiry`.

So a supplier that *has* a contract expiry set is still ignored, and expiring/expired contracts go silently unnoticed (Annex A 5.20).

## Behavior (sensible defaults, no configuration)

- **Notify at three points:** 30 days before expiry, 7 days before, and on the expiry date.
- **After expiry: no more notifications** — don't nag.
- **Persistent UI state:** an expiring (≤30 days) or expired contract is shown clearly in the overdue summary / dashboard as a problem, so it stays visible even after notifications stop.
- Suppliers with no `contract_expiry` set are never flagged (NULL = nothing to check).

No org settings, no thresholds to configure — the 30 / 7 / 0-day points are hardcoded sensible defaults. Keep it simple.

## Implementation notes

- **No schema change** — reads the existing `contract_expiry` column and reuses the existing overdue/notification machinery.
- Notifications fire once per threshold by matching the discrete day marks (30 / 7 / 0) in the daily job — no repeated spam, and no need to store notification state.

## Acceptance criteria

- A supplier whose contract expires within 30 days, or is already expired, appears in the overdue summary and dashboard with a clear state.
- Notifications fire at the 30-day, 7-day, and expiry marks — and not after.
- Suppliers without a `contract_expiry` are unaffected.
- Test covers a supplier with an expired contract appearing in the overdue summary.

unidoc-alip:
Supplier contracts have an expiry date that the platform stored and let you edit, but never actually looked at. A contract could lapse without anyone being told and without it showing up anywhere — the date just sat in the record.

## What changes for users

- **Contracts expiring within 30 days, or already expired, now appear in the overdue list** — on the dashboard, in the web overdue widget, and in the CLI (`isms overdue`, `isms status`). Each one is labelled as either expiring or already expired, so the two read differently at a glance rather than both being "late".
- **Email-style in-app notifications fire three times: 30 days out, 7 days out, and on the expiry date.** They go to the contract owner, or to admins and managers when the supplier has no owner — an unowned expiring contract is exactly the case somebody needs to pick up.
- **Nothing fires after the expiry date.** Once a contract has lapsed it stays visible in the overdue list, which is persistent, rather than generating a repeating nag.
- **Suppliers with no expiry date set are completely unaffected.** Nothing new appears for them and no notification is ever sent.

## Scope and judgement calls

- **Terminated suppliers are excluded.** This isn't spelled out in the original request, but without it an expired contract on a supplier you've already ended would sit in the overdue widget forever with no way to clear it.
- **The dashboard "Overdue" count now includes contracts that are expiring but haven't lapsed yet.** This is deliberate: the point is to see a contract coming up for renewal *before* it becomes a problem, not after.
- **No tasks are auto-created for contract expiry.** A contract running out isn't a review that someone performs, so it doesn't belong in the task list — the notification and the persistent overdue entry are the prompt.
- The 30/7/0-day marks are fixed, not configurable. No new setting to manage.

## Risk

Low. **No database migration** — the expiry column already existed and was simply never read. The change adds two read-only queries and one notification path; nothing existing was modified, so suppliers without an expiry date, and every other overdue item type, behave exactly as before.

The one thing worth knowing: the background job that sends these notifications is documented as running hourly, so it deliberately checks whether it already sent a given notice today before sending again. That check is per-supplier, which matters when one person owns several suppliers that hit the same mark on the same day — they should get one notice each, not one in total.

## Verification

- Full Go build, vet and test suite pass; web test suite passes including the translation-key and raw-text checks.
- Live-database tests cover the boundaries directly: a contract 31 days out is absent, 30 days out is present, and the 30/7/0 notification marks match only on those exact days.
- Job-level tests cover who receives each notice, that a second run the same day sends nothing, that two suppliers owned by one person both notify, and that an already-expired contract notifies nobody.
- Verified end to end against a real database, API server and browser: the correct suppliers appear with the correct states, expired entries sort above expiring ones and read in red against amber, the notifications render as readable sentences with the supplier name filled in, and clicking one opens the right supplier record.

## Noted but not addressed here

Two pre-existing behaviours turned up during testing that are outside this change's scope:

1. The dashboard's overdue widget only renders when the organisation has at least one risk recorded. On an organisation with suppliers but an empty risk register, this new contract information wouldn't be visible.
2. Rows in that widget aren't clickable for any item type. The notification links work and land on the right record, so the feature is reachable, but the widget itself isn't an entry point.

Happy to take either as a follow-up.

Closes #44.


unidoc-ahall (approved):
Independently verified against a live Postgres.

`go build/vet ./...` clean, full `go test ./...` green (`ISMS_TEST_DATABASE_URL` set). `web/`: 198/198, `npm run i18n:raw`/`i18n:error-raw`/`i18n:keyset-diff` all clean (0 findings, keyset matches at 2485 keys), `npm run build` clean.

**Mutation-tested "the regression that matters most"** (the test file's own words, case 4): removed the `AND n.params->>'id' = $4` clause from `NotificationExistsToday`'s query in `db/notifications.go`. Both `TestContractExpiryTwoSuppliersSameOwnerBothNotify` (api package) and `TestNotificationExistsTodayDiscriminatesBySupplier` (db package) fail immediately — the latter with a driver-level arg-count error, the former with `sent = 0, want 2`, confirming that without per-entity discrimination the second of two suppliers owned by the same person at the same mark would be silently swallowed as a false "already sent". Restored, `git status` clean, both suites rerun green.

**Traced the notification-vs-overdue-display split**, which is the actual design the PR is making a case for: `SupplierContractsAtMarks` matches only the exact 30/7/0-day marks (discrete `IN`, not a range) for the notification job, while `GetOverdueSummary`'s separate query is unbounded below 30 days (`<= 30`, including negative/expired) for the persistent widget/CLI listing — so a contract that's been expired for months still shows up every time `isms status` runs, without re-notifying. That matches "nothing fires after expiry, but it stays visible" exactly.

**One thing worth knowing, not blocking**: the notification match is a discrete day count with no catch-up. If the hourly cron has a full-day outage that spans exactly a 30/7/0 mark, that specific notice is silently skipped forever — `days_until` moves past the mark before the next run and `IN (30,7,0)` never matches again. Given the job is a simple hourly cron with no state beyond "did I already send this", this is a reasonable trade for the simplicity gained (no separate "pending marks" table), but worth being aware of if the manager cron is ever down for an extended stretch — the affected contract would need a manual nudge, not just a service restart, to be re-flagged. Not something this PR needs to solve.

Also checked: `TaskViewer` (the private-task visibility mechanism) is not applied to the new supplier_contracts query in `GetOverdueSummary`, same as the pre-existing risks/suppliers/systems/legal queries in that function — consistent with how that function already works, not a new inconsistency. `NotifySupplierContractExpiry`'s Title/Body/TitleKey/BodyKey/Params shape matches the existing `incident_new` notification convention exactly (raw English strings alongside the i18n key+params pair). Terminated-supplier exclusion, owner-fallback-to-admin/manager, and the SQL-side date arithmetic (avoiding a Go-side timezone-dependent day boundary) are all correct and match the stated design.

Well-tested, well-reasoned feature. Approving.

Issue #290, opened 2026-09-16
Clearing any org setting to an empty value permanently shadows its registry default


unidoc-alip:
`GetOrgSetting` resolves a value with `COALESCE(os.value, s.default_value, '')`. An explicitly stored empty string is not NULL, so COALESCE returns the empty string and never falls through to the registry `default_value`.

`SetOrgSetting` writes whatever it is given, including `''`, as a real row. So clearing a setting through the admin settings endpoint does not revert it to the platform default — it pins it to empty forever, and every caller then sees "" and applies whatever its own local fallback happens to be. `GetOrgSettings` (the plural, used to render the admin settings UI) has the identical COALESCE, so the UI shows the setting as empty rather than showing the default it is supposed to have reverted to.

**How it surfaced:** live testing of the new per-org supplier review cycles. Clearing `supplier_review_cycle_medium` produced a 12-month schedule instead of the registry default of 6. That specific case was fixed by deleting the row instead of storing an empty string, but the fix was deliberately scoped to the review-cycle keys — the underlying resolver is unchanged and every other setting in the registry still has this behaviour.

Worth knowing: the bug has a masking condition that makes it easy to miss in tests. For the review-cycle keys, if no other level is configured the resolver returns an empty map and the caller falls back to correct defaults anyway, so a test that sets one key, clears it, and checks the result passes even with the bug present. A test has to leave a second key configured to expose it.

**Two candidate fixes:**
- Change the resolver to `COALESCE(NULLIF(os.value, ''), s.default_value, '')`, treating empty as absent. Applies everywhere at once, but changes semantics for any setting where empty is a legitimate stored value — that needs an audit first.
- Have the settings write path delete the row whenever the incoming value is empty. `DeleteOrgSetting` already exists for this and had no callers before the review-cycle fix.

The second is narrower and matches what "revert to default" means in the settings UI.

unidoc-alip:
Verified against master (3b40a4c). The mechanism is exactly as described — `internal/isms/db/settings.go:52` and `:75` both resolve with `COALESCE(os.value, s.default_value, '')`, and `SetOrgSetting` stores `''` as a real non-NULL row, so a cleared setting shadows its registry default permanently.

[...]

Issue #289, opened 2026-09-16
Risk and legal readings ignore the org's configured review cycles


unidoc-alip:
Two related gaps, both now easy to close since supplier review cycles were moved into the settings registry.

**1. Systems still use a hard-coded switch.** `systemReviewMonths` in the systems DB layer returns 1/3/6/12 by criticality with no per-org override, exactly as suppliers did before. The same Annex A 5.22 argument applies: review intervals should match the org's risk context. The shared helper that resolves per-level cycles from `organization_settings` under a key prefix already exists, so bringing systems onto it is mostly mechanical — a migration seeding four `system_review_cycle_*` keys and a signature change on `System.CalculateNextReview`.

**2. Risk and legal readings ignore the org's configured cycles entirely.** Two call sites in the readings API pass a nil cycles map when computing next-review dates for risks and legal requirements. Nil means "fall back to built-in defaults", so an org that has configured `risk_review_cycle_*` has those values silently ignored on this path — the setting appears to work in the UI but does not apply to readings.

The accessor these paths need is currently unexported, which is why nil was passed. It should be exported (the supplier equivalent already is, for this reason) and both call sites updated to resolve real cycles.

Item 2 is a user-visible correctness bug and is the more urgent of the two.

unidoc-alip:
Verified against master (3b40a4c). Both items are real, but they belong in different releases, and item 2 needs a bigger fix than the description suggests.

**Item 2 does not work as described.** The two readings call sites (`internal/isms/api/api_readings.go:154` and `:392`) do pass nil, but fixing them alone changes nothing. `UpdateRiskTx` opens with `r.CalculateScore(nil)` and `UpdateLegalRequirementTx` with `lr.CalculateRiskScore(nil)`; both end in `CalculateReviewDate(cycles)`, which reassigns `NextReview` unconditionally. Whatever the caller computed is overwritten by the nil-cycles result inside the transaction helper. Neither helper takes a cycles parameter, unlike `UpdateSupplierTx`.

So the real fix is: export the per-org cycle accessor, thread cycles through `UpdateRiskTx` and `UpdateLegalRequirementTx` (five call sites between them, across readings and the suggestions apply handlers), and drop the nil at the readings call sites. That is the same pair of helpers #288 needs to touch for the same reason, so the two should be fixed together in one change.

[...]

Issue #288, opened 2026-09-16
Supplier next_review date is silently overwritten when explicitly set


unidoc-alip:
When a supplier reading is submitted with an explicit next-review date, that date does not persist.

`UpdateSupplier` and `UpdateSupplierTx` call `CalculateNextReview` unconditionally, so whatever explicit `next_review` the caller set is recomputed from criticality and discarded before the row is written. `writeSupplierFromReading` in the readings API does exactly this: it parses the user-supplied date, assigns it, then calls the update helper that overwrites it.

The doc comment on the calculation previously claimed this only happened "when not already set", which was never true — the comment has since been corrected to describe the real behaviour, but the behaviour itself is unchanged.

**Why it wasn't fixed inline:** making the recalculation conditional changes the semantics of `PUT /suppliers/:id` for every caller, not just the readings path. That deserves its own review rather than riding along inside an unrelated feature change.

**Suggested fix:** only recalculate when the caller has not supplied an explicit `next_review`, and add a test covering the readings path end to end.

unidoc-alip:
Verified against master (3b40a4c). The diagnosis holds, but the scope is wider than reported: this is not supplier-only.

All four register `*Tx` update helpers recompute `next_review` unconditionally:

- `UpdateSupplierTx` (`internal/isms/db/suggestions_tx.go:369`) — `s.CalculateNextReview(cycles)`
- `UpdateSystemTx` (`internal/isms/db/suggestions_tx.go:617`) — `sys.CalculateNextReview()`
- `UpdateRiskTx` — `r.CalculateScore(nil)`, which ends in `r.CalculateReviewDate(cycles)`
- `UpdateLegalRequirementTx` — `lr.CalculateRiskScore(nil)`, same tail call

`CalculateNextReview` and `CalculateReviewDate` both assign `NextReview` with no guard. So the "explicit user input wins" branch in the readings API is dead code for every entity type that has one, not just suppliers:

- `internal/isms/api/api_readings.go:148` (risk)
- `internal/isms/api/api_readings.go:386` (legal requirement)
- `internal/isms/api/api_readings.go:491` (supplier)
- `internal/isms/api/api_readings.go:595` (system)

Each parses the user-supplied date, assigns it, and then calls a helper that recomputes over the top of it. Assets are the exception and the working counter-example: `writeAssetFromReading` derives the date inline and `UpdateAssetTx` does no recalculation, so the explicit date survives there.

[...]

PR #291, opened 2026-09-16
Let each organization set its own supplier review intervals (#43)

### LINKED ISSUE #43: Make supplier review cycle configurable per org (Annex A 5.22)
## Context

Supplier review-cycle lengths are hard-coded per criticality (critical=1 / high=3 / medium=6 / low=12 months) with no org-level override — `supplierReviewMonths` (`db/suppliers.go:81`) is a pure switch with no ctx/orgID.

Risk review cycles already do exactly what we want here: `riskReviewCycles(ctx, orgID)` (`db/risks.go:221`) reads `risk_review_cycle_{level}` from `organization_settings`. This is bringing suppliers up to the same, already-proven pattern.

Annex A 5.22 calls for reviewing supplier relationships at intervals appropriate to risk — the hard-coded defaults won't match every org's supplier risk context or contractual obligations.

## Scope

- Seed `supplier_review_cycle_{critical,high,medium,low}` keys in `organization_settings` (mirror the risk-cycle seed).
- Change `supplierReviewMonths` / `CalculateNextReview` to take `ctx + orgID` and read these settings.
- Optional: per-supplier `review_cycle_months` override column.
- Preserve current values (1 / 3 / 6 / 12) as defaults — existing orgs see no behavior change.

## Schema / release

Seeds new settings (and optional override column) → feature release (0.8.0).

## Acceptance criteria

- `supplier_review_cycle_*` org settings exist and drive supplier review scheduling.
- Defaults match today's hard-coded values; existing orgs unchanged.
- If included, a per-supplier override takes precedence over the org setting.

unidoc-alip:
Closes #43

## What changes for users

Until now, how often a supplier had to be reviewed was fixed in the code: critical suppliers every month, high every 3 months, medium every 6, low every 12. No organization could change those numbers.

An admin can now set them per organization, under **Review Cycles** in the admin settings screen — the same place risk review cycles already live. ISO 27001 Annex A 5.22 asks that supplier relationships be reviewed at intervals appropriate to the risk, and a single fixed schedule doesn't fit every organization's supplier mix or contractual commitments.

## Nothing moves for existing organizations

The four new settings ship with defaults of 1 / 3 / 6 / 12 months — exactly the values the old code used. Those defaults live in the settings registry, and any organization that hasn't changed them simply resolves to them.

Concretely: **no data migration, no backfill job, and no supplier's review date shifts on deploy.** An organization only departs from today's behaviour if someone deliberately edits a value.

## Where the setting takes effect

Every path that schedules a supplier review honours it — creating a supplier, editing one, recording a supplier review, submitting a supplier reading, applying an AI or human suggestion, and the background job that fills in missing review dates. The goal was that the setting can't be quietly ignored on one screen while working on another.

Bad values are rejected at the point of entry rather than silently misbehaving: a review cycle must be a whole number of months between 1 and 120. Previously a typo like "soon" would have been stored and quietly degraded to a 12-month schedule. That guard also covers the existing risk review cycle settings, which had no validation at all.

## A bug this work surfaced and fixed

During live testing, clearing a review cycle back to blank — the obvious way to say "go back to the default" — didn't work. It left the organization on a 12-month schedule instead of restoring the real default.

The cause sits deeper than this feature: clearing a setting stored an empty value that permanently shadowed the platform default. This PR fixes it for the review cycle settings only, by removing the stored value rather than blanking it. The broader flaw affects every other setting in the registry and is filed separately as #290, since fixing it globally changes behaviour well beyond suppliers and deserves its own review.

## Deliberately left out

- **Per-supplier overrides.** The original issue floated an optional per-supplier interval. It's a much larger surface — new column, API and CLI fields, suggestion handling, a new form field and its translations — and isn't needed to close the issue. Worth doing as its own change if there's demand.
- **Systems.** System review cycles have the identical hard-coded problem. Left untouched here to keep the change reviewable; filed as #289.
- **A pre-existing supplier bug.** An explicitly chosen next-review date is overwritten by the automatic calculation. Fixing it changes how the supplier update endpoint behaves for all callers, so it's filed as #288 rather than bundled in.

## Risk

Low. The default path is unchanged by construction, and the new behaviour only activates when an admin edits a value. The main risk would be a schedule shifting unexpectedly for an organization that never touched the settings, which is what the defaults-and-no-backfill design and the tests below are there to prevent.

## Verification

Unit tests cover the schedule calculation, the defaults contract, and the settings validation. Database-backed tests cover resolving and reverting the settings.

Beyond the test suite, the feature was exercised against a running server and a real database: defaults confirmed at 1/3/6/12 with no settings rows written, overrides confirmed to take effect on create, update, supplier reviews, suggestion apply (both create and update), and the background backfill, and validation confirmed to return real rejections for bad input. The revert bug above was found this way — it passed every unit test and the full build, because clearing a setting returned a success response while doing the wrong thing underneath.


unidoc-ahall (approved):
Independently verified against a live Postgres.

Applied the renamed migration (`20260915000000_v0.8.0.sql`, third rename of this unreleased-release file) via `isms server migrate --dir migrations` — reapplies cleanly under the new filename, idempotent as the file's own header requires. `go build/vet ./...` clean, full `go test ./...` green with `ISMS_TEST_DATABASE_URL` set. No web files in this PR, so no frontend run needed.

**Mutation-tested the actual bug this PR fixes**, not just the description of it: reverted the `handleAdminUpdateSetting` clear-path from `DeleteOrgSetting` back to `SetOrgSetting(ctx, orgID, req.Key, "")` (the pre-fix behavior). `TestAdminUpdateSettingAcceptsValidReviewCycle`'s "empty reverts to the registry default" case fails exactly as described — `stored value="", want "6"`, plus the explicit row-count assertion catches it too (`organization_settings` row still present after "clearing"). Confirms the root cause is real: `GetOrgSetting`'s `COALESCE(os.value, s.default_value, '')` cannot distinguish an explicitly-stored empty string from no row at all, so `SetOrgSetting("")` permanently shadows the registry default. Restored, `git status` clean, both `api` and `db` suites rerun green.

**Traced every write path that resolves review cycles**, since the PR's central claim is that none of them can silently keep using the hardcoded 1/3/6/12 switch: grepped all `Supplier.CalculateNextReview` call sites — `CreateSupplier`, `UpdateSupplier`, `CreateSupplierReview`, `writeSupplierFromReading` (readings), `CreateSupplierTx`/`UpdateSupplierTx` (suggestion apply, both create and update), and the manager cron's backfill loop — every one now resolves `d.SupplierReviewCycles(ctx, orgID)` rather than passing nil. The only remaining no-arg `CalculateNextReview()` calls are on `System`, a distinct type with its own method — correctly untouched, since systems are explicitly out of scope here (#289). Cycles are resolved once per org before the backfill loop in `manager.go`, not once per supplier — correct, since the value can't change mid-loop and it's 4 settings lookups either way.

**Checked for bypasses**: `SetOrgSetting` for these two key prefixes has exactly one caller in the whole codebase — the now-guarded handler. Nothing (CLI, MCP, another handler) can write an unvalidated review-cycle value around the new 1–120 whole-month check.

**Read the defaults contract test** (`TestReviewCycleDefaultsContract`) and confirm it pins `reviewCycleDefaults` at 1/3/6/12 — since AC-2 (no behavior change for untouched orgs) depends entirely on those numbers never drifting, and on the migration's `default_value` column matching them, which it does.

Well-scoped, honestly bounded (the three deliberately-excluded items are each filed as their own issue rather than silently dropped or scope-crept into this change), and the revert-bug fix is proven by a test that actually distinguishes "looks fixed" from "is fixed" at the storage-row level. Approving.

PR #293, opened 2026-09-16
Optional External ID on register records (#39)

### LINKED ISSUE #39: Optional external_id field on register entities
## Context

From discussion #34. A user keeps their own identification numbers (date + company shortcut) in an external system and wants to record that reference against incidents. The ask started as customizing our own incident identifiers (custom IDs + numbering patterns); in the thread it narrowed to the real need: **an optional field to store an external/own identifier** — without changing our canonical identifier.

This is the better-scoped version. Our canonical identifiers (`INC-N`, `RISK-N`, …) stay intact, so the entity-reference system and its integrity guarantees are untouched. The external ID is purely additive metadata.

## Decision

Add an optional `external_id` field to register entities — **generic across all registers** (incidents, risks, suppliers, corrective actions, systems, assets, legal requirements, …), not incident-only. The same need — linking an entity to an external system's identifier — applies everywhere.

## Scope (v1)

- Nullable `external_id` text field on register entities.
- Optional in create/edit forms — no burden on users who don't need it.
- Indexed so it's searchable; shown where identifiers are shown (entity detail; ideally a list column).
- **No hard uniqueness** in v1 — it's a reference, not a key. A soft uniqueness check can come later if a concrete need appears.

## Explicitly out of scope

- Changing our canonical identifiers or numbering scheme — that was the original framing; it would touch reference integrity for no real gain here.
- Configurable numbering patterns (`INC-[YYYY]-[ID]`). If a concrete need resurfaces, that's a separate discussion.

## Schema / release

Additive migration → ships in a feature release, not a patch.

## Open questions (input welcome)

- Which registers matter most first? (incidents confirmed; others?)
- UI label — "External ID", or something clearer?
- Should it be a list column by default, or detail-only?
- Need to search/filter by it from day one?

unidoc-alip:
Closes #39.

## What this gives users

Organizations that already track records in another system (a ticketing tool, an ERP, their own spreadsheet numbering) can now record that outside reference directly on a record here. It is an optional free-text **External ID** field — type it in, or ignore it entirely.

Our own identifiers (`INC-7`, `RISK-12`, …) are untouched. The external ID sits alongside them as extra information; it is a reference, not a key.

## Where it appears

All seven registers at once — incidents, risks, assets, suppliers, systems, legal requirements, and corrective actions. The field is trivial per register, and shipping only some of them would be more confusing than useful.

In each register you can:

- set an External ID when creating a record,
- edit or clear it later from the record's overview,
- see it next to the record's own identifier in the list and on the record detail — but only when it has been filled in, so nothing changes visually for organizations that never use it,
- search for it: the existing search box on each register list now matches on the External ID as well as the fields it already searched.

Translated into English and Indonesian.

## Answers to the questions raised on the issue

- **Which registers first?** All seven together, for the reason above.
- **UI label?** "External ID".
- **A list column by default?** No. A dedicated column would cost every organization screen space for a field most will leave blank. Instead it renders inline beside the record's identifier, only when set.
- **Searchable from day one?** Yes.

## Risk and scope

Low. The change is purely additive:

- One database migration adds a nullable text column and a per-organization index to seven tables. Nothing is rewritten and no existing data is touched; records without an external ID simply have nothing there.
- No uniqueness constraint. Two records may carry the same external ID — deliberate, since it mirrors whatever another system already uses.
- No change to how our identifiers are generated or to the cross-record reference system that depends on them.

The one genuine risk in a change like this is a new field accidentally wiping unrelated data when a record is saved from somewhere else in the app. Every code path that loads and re-saves these records was enumerated and checked against that; a case where a form failed to clear the field was caught by tests and fixed before this was opened.

## Verification

- 28 new API tests (four per register: set on create, clear, leave untouched, find by search).
- 21 new browser tests driving the real UI across all seven registers — creating with an External ID, editing it, clearing it and confirming it disappears from the page, confirming a record without one shows nothing stray, and searching by it. These were checked against a deliberately broken build to confirm they actually fail when the feature is broken.
- Full suite green: Go build, vet and tests; frontend build and unit tests (198); Python suite (722 passed, 1 skipped).


unidoc-ahall (approved):
Independently verified against a live Postgres. One real, non-blocking gap found by reading the code, not by trusting the description.

Applied the renamed migration (`20260916000000_v0.8.0.sql`, fourth rename of this unreleased-release file) via `isms server migrate --dir migrations` — clean, idempotent. `go build/vet ./...` clean, full `go test ./...` green. `web/`: 198/198, `npm run i18n:raw`/`i18n:error-raw`/`i18n:keyset-diff` all clean (keyset now at 2492 keys), `npm run build` clean. The Python suite (28 API tests + the new `test_e2e_external_id.py` browser suite) I could not run myself — no `pytest`/`requests` in this container — so that part is unverified on my end, same caveat as always; I did read `test_e2e_external_id.py` in full and it's a genuinely careful suite (independent tags for title vs. external_id so the search assertions can't pass by accident through the title match, polling the real API state before asserting on rendered copy rather than racing the async PUT, its own file specifically to avoid the `test_e2e_browser.py` merge-conflict problem flagged a while back).

**Found a real gap: External ID changes never appear in a record's History panel.** `Asset.ToChangeMap()`, `Risk.ToChangeMap()`, `Supplier.ToChangeMap()`, `LegalRequirement.ToChangeMap()`, `Incident.ToChangeMap()`, `CorrectiveAction.ToChangeMap()`, and `System.ToChangeMap()` (all in `internal/isms/db/changelog.go` and `systems.go`) were not updated to include `external_id`. `DiffFields` (changelog.go:158) only iterates `newFields` — the `ToChangeMap()` output — so a key that was never added to that map literal can never produce a changelog entry, for any before/after values, deterministically. Every register's detail view wires `entity_changelog` straight into a visible `<HistoryPanel>` (confirmed in `Assets.vue`/`Risks.vue` and the rest), so a user setting, changing, or clearing the External ID today gets a `200 OK` and the value saves correctly, but the record's own audit history shows nothing happened — the one field on these seven registers with no trace in an otherwise fully-tracked change log. This doesn't need a live-DB test to prove (it's a map literal missing a key, not a runtime condition), so I didn't force one; I did confirm by reading `DiffFields` itself that there's no other path (e.g. comparing raw DB columns) that would catch it. Recommend adding `\"external_id\": x.ExternalID` to all seven `ToChangeMap` functions — small, low-risk, and worth doing before or immediately after merge rather than as a deferred follow-up, since this is a compliance platform and \"the change happened but the audit trail doesn't show it\" is exactly the failure mode that matters here.

Two smaller, non-blocking notes:
- Suggestion-apply create paths (`applyAssetCreate` etc. in `api_suggestions.go`) unmarshal into narrow local payload structs that don't include `external_id`, so an AI-suggested new record can't carry one today. Not claimed anywhere in the PR description (only direct create/edit is promised), so this isn't a broken claim — just a natural next step if agent-driven record creation needs it later.
- All seven search-clause changes (`... OR COALESCE(external_id,'') ILIKE $N`) are correctly parameterized — the pattern value goes through `args`, never string-interpolated into the query — so no SQL injection concern there.

Everything else checked out: the migration adds a nullable column plus a per-org index to all seven tables with no backfill and no touch to existing rows; the `nilIfEmpty(strings.TrimSpace(...))` pattern is applied uniformly on every create/update path (including the `*Tx` variants used by suggestion-apply update, which do carry `external_id` through correctly — only suggestion *create* is the gap above); the `*string` update-DTO pattern (nil = untouched, non-nil-empty = clear) is the same convention already used elsewhere in this codebase, applied consistently across all seven registers.

Solid, mechanically consistent feature across a wide surface, with one real fixable gap. Approving, with the History-panel gap called out as the one thing worth addressing.

PR #294, opened 2026-09-17
Release v0.8.0

unidoc-ahall:
Bumps version.txt to 0.8.0. After merge: `just release 0.8.0` tags master and CI publishes the release.


unidoc-alip (approved):
> **Verdict: APPROVE** — 0 must-fix, 0 should-fix. Release-ready.

The diff is `version.txt` only, `0.7.1` to `0.8.0`, verified against `origin/master`. Nothing
else rides along, so the only real question is whether master at this commit is fit to ship. I
ran the full suite locally on top of the green CI run, including the parts CI does not cover.

**What was tested**

Static and build gates: `gofmt -l .` empty, `go vet ./...` clean, `go build ./...` clean. The
embedded migration set was checked against `migrations/` by name and byte-compared per file —
identical, so the binary carries the migrations it should. `just build-go` produces a binary
stamping `0.8.0.157`, and the migration filename convention holds across all five migrations.

Go tests: the unit suite across `./internal/isms/...` and `./cmd/...` passes. Separately I ran
the live-Postgres tests in `internal/isms/db` with `ISMS_TEST_DATABASE_URL` pointed at a freshly
migrated throwaway database — 136 tests, zero skipped. These `t.Skip` without that variable, so
CI never exercises them; they pass.

Frontend: 198 unit tests pass. The i18n gates are clean — keyset matches the snapshot at 2492
keys with no drift, and both the raw-text and error-render ratchets report zero findings. The
production build succeeds.

API integration: 723 collected, 718 passed, 4 expected skips (subdomain-routing tests, with
that mode deliberately off). One failure, `test_brute_force_protection`, is a harness problem
rather than a product defect — its skip gate reads `ISMS_RATE_LIMIT` from the pytest process
while the server reads its own environment, so it fails to skip against a correctly configured
server. Rate limiting was genuinely disabled server-side; filed as #297.

End-to-end: the six browser suites, 100 tests, pass in full. A first cold run had three
failures, all the same login-page timeout on the org-slug input. They pass in isolation and the
whole suite passed 100 of 100 on a second run, so it is intermittent rather than a defect,
though the root cause is not established. E2E is not a CI gate.

A fresh-database boot was also exercised repeatedly during this work: all five migrations apply
cleanly from empty and the server comes up with no errors in the log.

**Why this is approved**

Every gate that guards a release is green, and the two blind spots in CI — the live-database Go
tests and the browser suites — were covered here and pass. The version bump is correct and
isolated.

Four bugs surfaced while verifying, filed as #295, #296, #297 and #298. None blocks this
release: all four are pre-existing and have shipped in earlier releases, none is a regression
introduced by 0.8.0, and none involves the release mechanics or the migration path. Recording
them here so the approval is not mistaken for "nothing was found":

- #295 — register readings never write an `entity_changelog` row; the insert violates the CHECK
  constraint, the error is only logged, and the endpoint still returns 201. A governance-trail
  gap, present since the initial commit.
- #298 — an `update` suggestion whose payload omits the `fields` wrapper is marked `applied`
  with a timestamp while the entity is untouched, so the trail records a change that never
  happened. Eleven apply handlers share the unguarded pattern.
- #296 — three client-error conditions answer 500 instead of 4xx, including the same
  open-corrective-action rule returning 409 through the direct endpoint but 500 through
  suggestion apply.
- #297 — the false-failure test gate described above.

#295 and #298 are the two worth prioritising after the release, since both damage the accuracy
of the governance record rather than just the response status.

PR #303, opened 2026-09-17
Add pt-BR (Brazilian Portuguese) translation

### LINKED ISSUE #302: pt-BR translation for i18n
Follow-up to #212. Now that multi-language support has landed, I've finished the Brazilian Portuguese (pt-BR) translation following the same structure used for Indonesian (#280). Translation complete and reviewed: 2492/2492 keys (100%), zero extra keys. Opening this issue as requested by @unidoc-alip in #212 so the PR has a dedicated tracking issue. PR coming right after this.

leonardo-ods:
## Summary

Adds Brazilian Portuguese (`pt-BR`) localization, following the structure introduced in #280 for Indonesian.

This is a follow-up to #212, where I offered to contribute the Brazilian Portuguese translation once internationalization support was available.

## Translation coverage

* **2,492/2,492 keys translated**
* **100% translation coverage**
* **Zero missing keys**
* **Zero extra keys**
* Translation reviewed for consistency and Brazilian Portuguese usage

## Localization decisions

The translation follows the conventions documented in the existing locale README.

The `pt-BR` section documents terminology decisions based on Brazilian standards and established local usage, including:

* ISO terminology such as *melhoria*, *melhoria contínua*, *correção*, *ação corretiva*, *não conformidade* and *oportunidade de melhoria*
* Brazilian audit terminology, including the use of *análise crítica* for management-system review
* Information classification terminology while preserving the semantics and ordering of the application's existing levels
* LGPD terminology, including *controlador*, *operador*, *encarregado* and *proteção de dados*
* Security-management vocabulary based on the Brazilian adoption of ISO/IEC 27001
* Translation of the full confidentiality, integrity and availability terms while preserving the existing `C`, `I` and `A` identifiers

The full terminology table and rationale are documented in the locale README, under the Brazilian Portuguese (`pt-BR`) section.

## Related issues

Closes #302 

Follow-up to #212


unidoc-alip (commented):
> **Verdict: COMMENT** — 0 must-fix, 1 should-fix, 0 nits

Thanks for this — and for the README section, which is the part that makes the locale maintainable rather than just present. I verified the coverage claims independently rather than taking them from the description, and they hold up: 2,492/2,492 keys, file set identical to `en`, zero missing and zero extra. Beyond raw key counts I also checked the things a key count can hide — placeholder parity (`{count}`, `{name}`, …), HTML tag parity, value-type parity, and vue-i18n plural-arm (`|`) counts — all zero mismatches. The 90 strings left identical to English are all legitimately locale-invariant (*Status*, *Total*, *Global*, brand names, `{value} KB`), so that is correct rather than missed work.

Wiring is complete. I grepped every `id-ID` reference outside the locale directories looking for a per-locale table that a new locale has to be added to and that is easy to miss; `loaders` in `web/src/i18n.js` is the only one, and you got it. There is no separate date/number-format or plural-rule map left unhandled. `pt-BR/index.js` matches `id-ID/index.js` apart from the locale name.

Checks run: `go test ./internal/isms/i18n/` passes, the web suite passes 198/198, the production build emits a `pt-BR` chunk (98.6 kB), and `/api/v1/config` returns `[en, id-ID, pt-BR]`, matching the updated assertion in `tests/test_coverage_boost.py`. Worth noting for reassurance: `localeKeyset` and `messageCompile` enumerate the locales directory dynamically, so `pt-BR` is already protected against future key drift without anyone adding it to a list.

I also ran it in a browser against a seeded organization. The switcher lists **Português (Brasil)** and switches immediately; `localStorage` and `<html lang>` both become `pt-BR` and survive a reload. Across 18 routes there were no raw translation keys leaking through, no unsubstituted placeholders, and no vue-i18n fallback warnings in the console. Interpolation renders correctly ("0 de 113 documentos aprovados", "1 incidente aguardando investigação"), dates localize ("18 de set. de 2026"), and the layout holds despite Portuguese running longer than English — the sidebar, the admin tab row and the register table headers all still fit.

One finding below. Findings are numbered `F<n>` to avoid GitHub autolinking them to unrelated issues.

---

### F1 [should-fix] `web/src/locales/pt-BR/` — CIA badge codes are localized in five places and left English in a sixth

The README section added in this PR states that the application "uses the existing `C`, `I`, and `A` codes as internal badge identifiers, so these codes remain unchanged in the `pt-BR` locale." The locale files do not match that statement — five of the six CIA code strings are localized to the CID form, and one is not:

- `common.json` → `cia_abbr.a` is `"A"` (unchanged, matching the README)
- `components.json` → `readings.table.availability` is `"D"` (`en` and `id-ID`: `"A"`)
- `risks.json`, `assets.json`, `systems.json`, `suppliers.json` → `table.cia` is `"C/I/D"` (`en` and `id-ID`: `"C/I/A"`)

The user-visible effect is that availability is labelled differently on adjacent screens in the same locale. The incident detail panel renders its classification badges from `common.cia_abbr.*`, so it shows **C I A**; the risk, asset, supplier and system register tables render the **C/I/D** column header, and the readings panel renders **C** / **I** / **D** column headers. I confirmed both renderings in the running app.

This also diverges from `id-ID`, which kept all six in their English form under the existing README rule that `common.enum.entity_abbr.*` and `common.cia_abbr.*` are badge codes rather than copy, constrained by the fixed-width chips they render in.

Either direction is a small change and you are better placed than I am to choose, since this is a question about Brazilian convention rather than about the code:

- **Localize consistently** — also set `common.json` `cia_abbr.a` to `"D"`, and update the README paragraph to say the codes are localized to the CID triad, noting that C/I/D does not collide the way some other languages' initials would.
- **Preserve consistently** — revert the five localized strings to `"A"` and `"C/I/A"`, matching what the README currently claims and what `id-ID` does.

The first option reads better for Brazilian users if CID is what they expect on screen; the second is the smaller diff. What should not ship is the current split, mostly because the README presents it as a deliberate decision, and the next locale to be contributed will copy that paragraph.

---

Two things I checked and am explicitly **not** raising against this PR: the risks table shows an untranslated category value `Technology` and a stray `Description Potential consequences` string. Both render identically in English, so they are pre-existing and unrelated to your changes.

leonardo-ods:
Thanks for the thorough review and guidance!

I resolved the inconsistency by standardizing on the Brazilian **CID** triad across the entire locale:
- Updated `web/src/locales/pt-BR/common.json` to set `cia_abbr.a` to `"D"`.
- Standardized all UI copy, field labels, and descriptions referencing CIA to CID (`assets.json`, `components.json`, `risks.json`, `suppliers.json`, `systems.json`).
- Updated the `web/src/locales/README.md` section to document the CID decision.
- Keyset re-verified against snapshot (2,492/2,492 keys, zero mismatches).

---

*(Separate observation)*: While testing the UI, I noticed that on `/admin/settings`, the setting titles and descriptions (e.g. `ai_enabled`, `ai_review_max_rounds`) render in English because titles are generated via `humanSettingLabel(s.key)` and descriptions come directly from the DB metadata, neither passing through vue-i18n. Just flagging this for your awareness as a future improvement area, since it requires a frontend/schema architectural decision outside the scope of this PR. Let me know if you'd like me to open a separate issue to track it.

unidoc-alip (dismissed):
> **Verdict: APPROVE** — 0 must-fix, 0 should-fix, 1 nit

F1 is resolved, and resolved more thoroughly than the finding asked for. You took the localize-consistently option, and the sweep caught a site I had not enumerated: the incident **list** row summary in `Incidents.vue` (the `cia_abbr` joiner around line 877), which builds its `C/I/D` string separately from the detail-panel badges. Had only the six I listed been changed, the incident list would still have read `C/I/A` next to a detail panel reading `C I D`. Thank you for widening it rather than matching the letter of the finding.

Findings are numbered `F<n>` to avoid GitHub autolinking them to unrelated issues.

**Static verification.** `grep` for `CIA` and `C/I/A` across `web/src/locales/pt-BR/` now returns zero hits. I also checked the case a CIA-text grep cannot see — a bare `"A"` availability code — by grepping the `en` baseline for `": "A"`: only `common.cia_abbr.a` and `readings.table.availability` exist, and both are handled. Key parity re-verified after the edit, since values changing can still break interpolation: 2,492/2,492 keys, zero missing, zero extra, zero placeholder mismatches, zero plural-arm mismatches, zero HTML-tag mismatches, zero value-type mismatches.

I also checked the README for a self-contradiction, because the general abbreviations section says `common.cia_abbr.*` are badge codes and notes that `id-ID` keeps them English. There is none: that section's closing rule is that the point is not that codes stay English but that the choice is deliberate and written down. Your new paragraph does exactly that, and C/I/D has no initial collision, unlike the Indonesian K/I/K case the section cites. The two sections agree.

**Visual verification.** I ran a production build of the head commit against a seeded organization and switched to Português (Brasil) through the settings picker. The triad is consistent everywhere it renders:

- Incident detail, under `CLASSIFICAÇÃO`: the badges are **C**, **I**, **D**
- Incident list row summary: `C/I/D`
- Risk, asset, supplier and system register table headers: `C/I/D` on all four
- Asset row chips: `C4` `I3` `D5`
- Asset stat chip: `Severo (CID = 5)`
- Readings panel header row: `Data`, `Avaliador`, `C`, `I`, `D`, `Notas`

The adjacent-screen mismatch described in F1 is gone. While the app was up I swept all sixteen organization routes for regressions and found no raw translation keys and no unsubstituted placeholders anywhere.

**Checks run:** `go test ./internal/isms/i18n/` passes, the web suite passes 198/198, and the production build emits a `pt-BR` chunk at 98.6 kB (28.4 kB gzipped) with no warnings.

---

### F2 [nit] `web/src/locales/README.md` — the CID paragraph does not mention the localized likelihood code

Entirely optional, and not worth a round trip on its own if you would rather leave it.

`components.json` `readings.table.likelihood` is `"P"` (*Probabilidade*) where `en` is `"L"`. That is the same deliberate-localization choice as the CID codes and it is the right call, but the new paragraph names only `common.cia_abbr.a`, `readings.table.availability` and the `table.cia` headers. Since the whole value of that paragraph is that the next contributor reads a localized code as a decision rather than an oversight, `likelihood` belongs in the same list.

---

On your separate observation about the settings screen: you are right, and I confirmed it in the running app. It is worth its own issue, so please do open one — I will add what I found while checking.

It is broader than titles and descriptions. On that screen every one of the six category headings renders in Portuguese, as do the surrounding controls, while seventeen of nineteen setting labels and all nineteen descriptions stay English. The result reads as a half-broken screen rather than as a missing feature. The label transform also yields "Ai Enabled" rather than "AI Enabled".

Two useful details for whoever picks it up. First, it is locale-independent and predates internationalization entirely — both call sites date to the initial commit, so Indonesian has the same gap. Nothing about it is caused by this PR, and it does not affect this approval. Second, it likely needs neither a schema change nor an architectural decision: the two labels that *are* translated, *Categorias de Risco* and *Campos Personalizados de Risco*, are bespoke blocks in the same component already calling `t()` with their own keys. A map keyed by setting key, falling back to the current de-slugged label and the stored description when a key is absent, would follow that existing pattern and let new settings degrade to today's behaviour until someone adds copy for them.

leonardo-ods:
Thanks for the nit and the detailed feedback throughout this review.

**F2 — likelihood code.** Added `readings.table.likelihood → "P"` to the README's localization decisions table.

**Additional undocumented codes found.** While revisiting the locale for F2, I noticed two other localized short-form codes that followed the same pattern but were not yet documented: `enum_abbr.finding_type.opportunity → "OM"` (*Oportunidade de Melhoria*, the established Brazilian audit abbreviation) and the NC abbreviations (`NC Maior` / `NC Menor`). Both are now covered in the same table. The README section on `common.cia.na` was also added to document a deliberate "kept as-is" decision: `"N/A"` is the native Brazilian Portuguese abbreviation for *Não Aplicável* — not the English form carried over — and the commonly used alternative `"N/D"` was explicitly rejected because *Não Disponível* carries a different and adverse meaning in a compliance context (missing evidence, potential nonconformity), which would directly contradict what the value represents to an auditor.

**N/D → N/A reversion.** `common.cia.na` had been translated to `"N/D"` in the original submission. This has been reverted to `"N/A"` for the reason above.

**Settings screen issue.** Will open a separate issue to track the untranslated setting labels and descriptions on `/admin/settings`, as requested.

unidoc-ahall (approved):
Verified independently on the technical/structural side; a genuine content spot-check on the linguistic side, with one honest caveat below.

**Mechanical verification**: `go build/vet ./...` clean, and both release gates pass with `pt-BR` enabled (`TestSecondLocaleRequiresExtraction`, `TestEnabledLocalesAreTranslated`). `web/`: 198/198, `npm run i18n:raw`/`i18n:error-raw`/`i18n:keyset-diff` all clean (2492/2492 keys, matching the PR's own count), `npm run build` clean.

**Went beyond the existing tooling**: wrote a standalone key-by-key comparison of every `en/*.json` against `pt-BR/*.json` — 0 missing keys, 0 extra keys, and (not something anything in this repo currently checks automatically for general UI keys) 0 mismatches in `{placeholder}` interpolation tokens across all 2492 keys. Every `{count}`, `{title}`, `{message}`, etc. that appears in an English string appears — same names — in its Brazilian Portuguese counterpart, so nothing will render a literal unfilled placeholder or silently drop a variable.

**Content spot-check**: read the full README addition. It's unusually well-sourced — cites specific LGPD articles (art. 5º VI/VII/VIII for controlador/operador/encarregado, which matches the statute), ABNT NBR clause numbers, and correctly calls out that Brazilian usage says "tríade CID" rather than "CIA" (confidencialidade/integridade/disponibilidade) — a detail easy to get wrong by translating literally. The follow-up commit (`6d27f7b`) shows the author catching and fixing a CIA/CID inconsistency across several files after the fact, which is the same self-correcting discipline I'd want to see. Spot-read several translated namespaces (legal.json, common.json, admin.json) — fluent, professionally-toned, and internally consistent with the terminology the README documents (e.g. `external_id` → "ID Externo", already aligned with the recently-shipped external ID feature, so this isn't translated against a stale snapshot).

**One honest caveat, not a blocker**: I can evaluate structure, consistency, and check specific factual claims (LGPD article numbers, standard terminology) against what I know, but I'm not a substitute for native-speaker sign-off on fluency and tone throughout 2492 strings, particularly given how much of this touches legal/compliance vocabulary (LGPD roles, ISO audit terminology) in a product whose whole job is getting compliance language right. Given the terminology work documented here is unusually careful and alip's already approved it, I don't think this should block merge — but a native Brazilian Portuguese speaker giving it a pass at some point (ideally someone with ISO 27001/LGPD familiarity) would be worth doing before or shortly after this reaches real pt-BR users, the same way we'd want for id-ID.

Well-documented, structurally verified, self-corrected once already. Approving.

leonardo-ods:
Thanks for the thorough review and for validating the translation, consistency, and integration details.

I appreciate the feedback throughout the review, especially the checks around terminology and locale consistency. The additional findings also helped make the pt-BR locale better documented and more maintainable.

Thanks again to everyone who reviewed and helped get this merged!

PR #306, opened 2026-09-19
Add is-IS (Icelandic) translation

unidoc-ahall:
## Summary

Adds a full Icelandic (`is-IS`) localization, following the structure
established for `id-ID` (#280) and the `pt-BR` pattern in #303.

## Translation coverage

- 2492/2492 keys translated
- 100% coverage, 0 missing keys, 0 extra keys
- Verified with a standalone key-by-key + `{placeholder}` interpolation
  parity check across every `en/*.json` vs `is-IS/*.json` pair: 0 missing,
  0 extra, 0 placeholder mismatches.

## Localization decisions (documented in full in `web/src/locales/README.md`,
under "Worked example — Icelandic (`is-IS`)")

- **CIA -> L/H/T.** Confidentiality/integrity/availability are *leynd*,
  *heilleiki* and *tiltækileiki* in established Icelandic infosec usage — the
  triad and every abbreviated badge/table-header in the app is standardized
  to L/H/T, not a literal CIA transliteration.
- **GDPR terminology follows Icelandic law directly** (Iceland applies GDPR
  under the EEA agreement; the operative statute is *Lög um persónuvernd og
  vinnslu persónuupplýsinga, nr. 90/2018*) rather than an ISO adoption:
  ábyrgðaraðili (controller), vinnsluaðili (processor), persónuverndarfulltrúi
  (DPO, art. 35), persónuupplýsingar (personal data).
- ISO/IEC 27001 has an official Icelandic translation (ÍST EN ISO/IEC 27001,
  published by Staðlaráð Íslands) which was not read clause-by-clause for
  this PR since it is a paywalled purchase — general management-system and
  audit vocabulary here follows established Icelandic ISO-9001-adjacent and
  infosec-industry usage instead, and the README section says so plainly
  rather than citing clauses that weren't checked.
- Nonconformity/audit-result grading (frávik, ábending, tækifæri til umbóta,
  meiri háttar/minni háttar) is certification-body practice, not standard
  text — flagged as such, same as the existing Indonesian section's
  equivalent caveat.
- One disclosed, not-fixed-here limitation: the app's `{count} X | {count} Y`
  pluralization uses vue-i18n's default two-way rule (singular only at
  exactly 1), which doesn't match Icelandic grammar's actual rule (singular
  after any count ending in 1 except 11). Fixing that is a plural-rule
  registration change to the i18n setup, not a per-locale content fix, so
  it's called out in the README rather than silently worked around.

## Verification

- `go build/vet ./...` clean.
- `go test ./...` green, including both i18n release gates
  (`TestSecondLocaleRequiresExtraction`, `TestEnabledLocalesAreTranslated`)
  with `is-IS` enabled.
- `web/`: full test suite green (198+), `npm run i18n:raw` /
  `npm run i18n:error-raw` both 0 findings, `npm run i18n:keyset-diff`
  matches (2492 keys — unchanged, since this locale doesn't add or remove
  any `en` key), `npm run build` clean.
- Independent key + placeholder parity check described above.

## Not done here

Native-speaker sign-off. I'm confident in the terminology sourcing (cited
above, verified against public sources for the legal/standards terms) and
did a full read-through for fluency, but this hasn't had an independent
Icelandic-speaker or ISO/GDPR-practitioner review pass yet — flagging that
openly rather than presenting it as fully signed off.


unidoc-alip:
> **Verdict: COMMENT** — 1 must-fix

CI is red on the `integration` job: 1 failed, 821 passed, 1 skipped. The root cause is a single missed edit — enabling a locale requires updating two pinned lists, and only one of them was updated. Findings are labeled `F<n>` to avoid GitHub autolinking bare `#N` to unrelated issues.

---

### F1 [must-fix] `tests/test_coverage_boost.py` line 378

`test_config_advertises_only_enabled_locales` pins the enabled-locale set as an exact list, and it still reads `["en", "id-ID"]`:

```
tests/test_coverage_boost.py:378: in test_config_advertises_only_enabled_locales
    assert [l["tag"] for l in cfg["locales"]] == ["en", "id-ID"]
E   AssertionError: assert ['en', 'id-ID', 'is-IS'] == ['en', 'id-ID']
E     Left contains one more item: 'is-IS'
```

Adding `"is-IS": {name: "Íslenska", enabled: true}` to the supported map means `GET /config` now advertises three locales, so the pin has to grow with it.

The shipped set is pinned in two places on purpose — one over the catalog, one over the HTTP handler — and this PR updated only the catalog one (`TestOnlyEnabledLocalesAreOffered` in `internal/isms/i18n/locale_test.go`, correctly). That is why `go-unit` is green and only `integration` is red. Both the test's own docstring and step 6 of `docs/i18n.md` call out that the two must change together, and the docs note this same miss happened once before, so no doc update is needed here — just the second edit.

Suggested fix:

```diff
         r = requests.get(f"{api_url}/config", headers=admin_headers)
         assert r.status_code == 200
         cfg = r.json()
-        assert [l["tag"] for l in cfg["locales"]] == ["en", "id-ID"]
+        assert [l["tag"] for l in cfg["locales"]] == ["en", "id-ID", "is-IS"]
```

`is-IS` goes last: `Supported()`'s contract is default first, then the rest by tag, and that matches the order the server actually returned above.

---

Nothing else appears to be blocking. I grepped every `id-ID` reference under `tests/`, `docs/`, `internal/` and `web/src` and this is the only remaining pin of the enabled set. Both release gates that guard the `enabled` flag — `TestSecondLocaleRequiresExtraction` and the 90% bundle-coverage check in `TestEnabledLocalesAreTranslated` — ran in the green `go-unit` job, so the `is-IS` bundle clears them, and the web keyset test passed. The one remaining item from the enablement checklist that CI cannot check is the layout QA for text overflow, since Icelandic copy runs longer than English in table headers, status badges and button rows.

unidoc-ahall:
Fixed in the latest commit — thanks for catching this.

`tests/test_coverage_boost.py:378` now asserts `["en", "id-ID", "is-IS"]`, matching `Supported()`'s actual contract (default first, then the rest sorted by tag — confirmed `id-ID` < `is-IS` alphabetically, so the order is right). Verified this is the only remaining pinned list: grepped `tests/`, `docs/`, `internal/` and `web/src/` for `"en", "id-ID"` and for any other `id-ID` reference in the Python suite — `test_coverage_boost.py` was the only file, and now has the single corrected occurrence.

Re-ran `go build/vet ./...`, `go test ./...` (including both release gates), and the full `web/` suite (198 tests, both i18n ratchets, keyset diff, build) — all still green. Couldn't run the Python integration suite myself in this environment, but the fix is a one-line, mechanically-verified pin update against logic I traced directly in `Supported()`, so I'm confident it resolves F1.

unidoc-alip (approved):
LGTM

PR #307, opened 2026-09-20
fix(web): keep table header rows in the document read view (#304)

### LINKED ISSUE #304: Table header row is entirely dropped in document preview / print (PDF) — content, not just styling, is lost
### What happened?

A table created via the editor's "Insert table" command, with a labelled header row (e.g. "Internal Issue" / "Overview" / "Risk Ref. on Risk Register") visible and correctly bordered while in Edit mode, is completely absent from the document when viewed in preview mode or printed/exported to PDF (tested via both a browser "Save to PDF" print destination and a physical/virtual PDF printer).

This is not a styling regression — the header row's text content does not appear anywhere in the rendered output. The table in preview/print begins directly with the first data row. A visible tell: that first (data) row is missing its top border in the rendered output, while every subsequent row has normal top/bottom borders — consistent with a `<thead>` (or equivalent bounding element) being removed entirely, taking its content with it, rather than merely losing its styling.

### Steps to reproduce

1. Open any document's Content tab in Edit mode.
2. Insert a table with a header row and label each header cell (e.g. "Internal Issue", "Overview", "Risk Ref. in any document).
3. Populate at least one data row beneath it.
4. Confirm the header row and its labels display correctly, with normal borders, while still in Edit mode — this is confirmed working correctly.
5. Save and switch to the read-only view of the same document (Cancel out of Edit mode, or navigate away and back).
6. Observe: the header row and its labels are gone entirely in this read-only view. The table starts with the first data row, which is missing its top border.
7. Print the document (browser print dialog → Save as PDF, or a PDF printer) from the read-only view.
8. Observe: same result in the exported PDF — header row and labels absent.

### Expected vs. actual

Expected behaviour

The header row — both its content and its semantic/structural role — should be preserved identically between the editor, the rendered preview, and any print/PDF export.

### Component / surface

Web UI

### Version

0.7.1

### Anything else?

**Environment**
isms.sh version: v0.7.1  (self-hosted, built from source via just build)
Browser: Chrome (latest)
OS: Windows (client), Ubuntu 24.04 LTS (server)
Affected document type: a Document (git-backed content, ISO 27001 template clause — documents/iso27001/clauses/04-context-of-the-organisation/4.01-external-and-internal-issues.md), not a register entity (Risk/Asset/Supplier/etc., which are Postgres-backed). Not tested against register entities — the bug may be specific to the Documents surface given it involves the git-backed content/rendering path, but this hasn't been confirmed either way.

**Suspected area** (confirmed via raw storage inspection + browser console)

Checked the document's raw stored content directly (git show HEAD:<path> on the org's bare repo). 
Findings:

-  The table is stored as raw HTML embedded inline in the Markdown file (Tiptap's output for tables includes:
`<table><colgroup><col style="width:...">, <th colwidth=...>` header cells with inline `<span style="color: rgb(...)">` styling, `<td colwidth=...>` data cells), not as Markdown table syntax (| --- | --- |).

- The header row and its <th> cells are correctly present in the stored file, with the right labels ("Internal Issue", "Overview", "Risk Ref. on Risk Register") and marked as <th>, not <td>. This rules out a save-side / serialization bug — nothing is lost when the editor writes the document.

- This means the bug is in the render pipeline (preview and print/PDF), specifically wherever the stored HTML is processed before display.

Additionally, opening the browser's DevTools Console while viewing the broken preview shows a directly relevant warning, reproducible on reload of this exact document (iso27001-4-1), originating from a file named useMarkdownConvert-*.js:

[tiptap warn]: Duplicate extension names found: ['link']. This can lead to issues.

Critically: this document's table renders its header row correctly in Edit mode. Comparing Edit mode against the read-only preview of the same document, same table, same underlying stored content:

- Edit mode (live Tiptap editor instance): header row displays correctly — labels, bold styling, borders all present and correct.
- Read-only preview / print / PDF export: header row entirely absent, as described above.

Since Edit mode loads the stored content directly into Tiptap's editor (no useMarkdownConvert step involved), while the read-only preview clearly does invoke useMarkdownConvert (per the console warning, tied to this document's load), this strongly narrows the fault to that specific conversion path used only for read-only rendering — not the editor, and not the stored data, both of which are confirmed correct. A duplicate registration of the Tiptap link extension inside that converter's parsing instance is a plausible mechanism for a <th>-containing row (with nested styled <span> content) to be dropped or misparsed during that conversion, while it passes through untouched when loaded straight into the real editor.

A separate, unrelated console error was also observed during this investigation and is noted here in case it's useful, though it is not believed to be connected to the table issue: Failed to load resource: 400 (Bad Request) on an endpoint matching /api/v1/changel.../iso27001-a-8-34 — possibly related to changelog-writing issues similar in shape to #295.

**Impact**

This is data loss, not a cosmetic issue: the header row's labels — the column semantics a reader needs to interpret the table — are silently missing from the primary rendered document and from any exported/printed artefact, **despite being correctly present in the underlying stored file**. 
This affects any document containing a labelled table, which is common in ISO 27001/9001/14001 documentation (e.g. control mappings, register summaries, context/issues tables as shown in the attached example). Since the printed/PDF output is frequently what gets shown to an external auditor, this could result in an auditor seeing an unlabelled, seemingly malformed table with no indication anything is missing — and since the underlying data is intact, this should be a render-layer fix only, not a data recovery concern.

unidoc-alip:
Fixes #304

## What was broken

Tables created with the editor's "Insert table" button lost their header row — the column labels — everywhere a reader sees the document. The labels showed correctly while editing, then vanished in the read-only document view, in the review view, and in the printed / exported PDF. The table also lost its top border, so it ran straight into the text above it with nothing to suggest a row was missing.

## Why it matters

The exported PDF is often what goes to an external auditor. They would receive a table of unlabelled columns with no indication anything had been dropped — the values were there, but nothing said what they meant.

## The good news

Nothing was ever lost from storage. The labels have always been in the saved document; only the display step dropped them. This fix restores them everywhere at once, with no data recovery and no migration.

## What changed

The code that prepares a document for display looked for the header row by searching for a specific HTML wrapper element. Tables written in plain markdown produce that wrapper, so they always worked. Tables created in the editor are saved in a different but equally valid shape that has no such wrapper, so the header was never found and was skipped entirely. Header rows are now identified by the cells they contain rather than by the wrapper around them, which covers both shapes.

The same fix is applied in two places, because the review view carries its own copy of this display logic.

## Risk

Low, and contained to how a table is displayed — no API, database or stored-document changes.

The one thing worth knowing: inline comments attach to a row by matching the exact markup of that row, so if the markup shifted even slightly, existing comments would silently detach from their table. Tables written in plain markdown come out character-for-character identical after this change, which keeps every existing comment attached. This is pinned by a test that was deliberately written and run *before* the fix, so it proves the output is unchanged rather than simply describing whatever the new code happens to produce.

## Verification

- New regression test reproducing the editor-authored table shape, which fails without the fix and passes with it.
- New test pinning the exact markup of a plain-markdown table, to guard comment attachment.
- Full web test suite green (200 tests); production build clean.

## Not covered

A table using the first column as a row label still drops those label cells. That is a pre-existing, separate limitation and is unchanged by this PR.


unidoc-ahall (approved):
Independently verified.

`web/`: 200/200, `npm run build` clean.

**Mutation-tested the actual fix**: reverted `contentBlocks.js`'s header detection back to the old `thead`/`tbody` querySelector approach (keeping the rest of the diff, including the new tests). The new regression test fails exactly as described — `assert.deepEqual(blocks.map(b => b.tag), ['thead', 'tr'])` gets `['tr', 'tr']` instead, i.e. the editor-authored table's header row is treated as an ordinary body row with no distinct header block. The byte-stability test for the plain-markdown case still passed even under the revert, confirming that path was never broken — only the editor-authored shape was. Restored, `git status` clean, full suite green again (200/200).

**Verified the "everywhere at once" claim architecturally, not just by reading it**: traced how the PDF export actually works — `Documents.vue`'s `printDocument()` is just `window.print()` on the already-rendered `DocumentViewer.vue` DOM (CSS print-media classes toggle visibility), not a separate rendering pipeline. So there's no third code path that also needed this fix; fixing `DocumentViewer.vue` and `contentBlocks.js` (review view) genuinely covers all three surfaces the description lists.

**Checked the comment-detachment risk directly in `useDocumentComments.js`**: `blockHash()` hashes `block.raw || block.html`, and table blocks are added via `addBlock()` with no `raw` argument — so `html` really is the hash anchor for every table row, and `commentsForBlock()` does hard-reject on a hash mismatch (no index fallback once a comment carries a hash). That's exactly the failure mode the byte-stability test protects against, and the claim that plain-markdown tables render character-for-character identical after this change holds under the mutation test above.

**Confirmed the "not covered" row-label limitation is genuinely pre-existing and unaffected**: a row mixing `<th>` and `<td>` correctly stays classified as a body row (excluded from `headerRows` by the `!tr.querySelector('td')` check), but the body-row loop only ever collects `querySelectorAll('td')` — so a row-header `<th>` cell was silently dropped before this PR and still is now. Not a regression, not something this PR claims to fix.

Solid, precisely-scoped fix with a test that proves the failure mode rather than just describing it. Approving.

ArturLorek:
Confirmed #307 fixes the missing header content on v0.master (built from source, 21/09/2026) — header row and labels now render correctly, thank you. 
One remaining gap worth flagging: the original per-cell styling (custom text colour, column widths) authored in the editor is not preserved in the read view — headers now render as a fixed uppercase style regardless of what was set in Edit mode. 
Not blocking, but worth a follow-up if exact-fidelity styling in the printed/PDF output matters for future review — happy to open a separate issue if useful.

unidoc-ahall:
Thanks for excellent QA @ArturLorek , we will fix it. Everything you do for us is extremely useful, open as many issues as you like! We will fix it all.

